PR #48, opened 2015-11-13
Include availability for records without items

tampakis:
Changed the availability service api to include availability for each holding_id rather than location code. Advances #40 
Example:
updated_availability_api

PR #50, opened 2015-11-15
include holding id with holding record fields

### LINKED ISSUE #49: Include holding record id with the fields collapsed into the bib record
We want to include the holding record id in all fields that are pulled from the holding record into the bib record. This should allow for a more reliable way of grouping holding-record-specific information in the Solr index. To accomplish this, whatever fields we pull from the holding record will need to have the holding id included as a subfield. The subfield `$0` isn't used so we will that to include the holding id. For now I'm going to add this for the 852 and 856 fields. Based on the outcome of #46, we may include more.

tampakis:
Closes #49 
I'd like to have a full Voyager dump completed before I'm back in office with this new subfield $0 so we can index the holding id with the holding block info in solr (which will allow Orangelight to consume the new availability API in #48).

### DIFF
--- a/spec/requests/bib_gets_spec.rb
+++ b/spec/requests/bib_gets_spec.rb
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'marc'
 
 RSpec.describe "Bibliographic Gets", :type => :request do
   describe "GET /bibliographic/430472/items" do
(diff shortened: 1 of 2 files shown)


kevinreiss:
:+1:

PR #52, opened 2015-12-03
retrieve all holdings when single bib given

tampakis:
When you make an api call in the form of `/availability?id=4679943` all of the holdings are retrieved. Only the first item is currently retrieved. The response looks like this:
full_avail


kevinreiss:
:+1: Thanks for adding this one nikitas.

Issue #47, opened 2015-11-13
On Order status isn't available in items JSON


kevinreiss:
http://catalog.princeton.edu/cgi-bin/Pwebrecon.cgi?BBID=9222024 is an "On Order" in the "Pre-Order Process" 

If has an "f" holding record, http://bibdata.princeton.edu/bibliographic/9222024/holdings but a returns the empty set for http://bibdata.princeton.edu/bibliographic/9222024/items. 

Is the query https://github.com/pulibrary/marc_liberation/blob/development/voyager_helpers/lib/voyager_helpers/queries.rb#L57 supposed to handle this? It looks like it is firing https://github.com/pulibrary/marc_liberation/blob/6663b16b8136555f989d96abcdae55083132b36c/voyager_helpers/lib/voyager_helpers/liberator.rb#L156 and should add an order key to the item response if query is successful.

jpstroop:
It's ringing a bell with me that Janet said something about the [status values](https://github.com/pulibrary/marc_liberation/blob/development/voyager_helpers/lib/voyager_helpers/queries.rb#L59-L60) perhaps needing to be expanded in that query--right now I believe the order has to be approved.

Paging @mzelesky. Any thoughts about this? This might be another good one to rewrite.

mzelesky:
I'll look at it today and see if there are any other values that should be included.

jpstroop:
:+1:

mzelesky:
Ok, I looked at it today. I should have a final query for you tomorrow morning.

mzelesky:
Here is a revised function. If I didn't get the Ruby syntax correct, at least the query works. The 2 PO Statuses to consider are 1 (Approved/Sent) and 3 (Received Partial).


def approved_orders(bib_id)
        # PO Statuses to consider: 1, 3
        line_item_status_approved=8
        %Q(
        SELECT LINE_ITEM.BIB_ID, 
		PURCHASE_ORDER.PO_STATUS, 
		LINE_ITEM_COPY_STATUS.LINE_ITEM_STATUS, 
		LINE_ITEM_COPY_STATUS.STATUS_DATE 
		FROM ((PURCHASE_ORDER 
		INNER JOIN LINE_ITEM ON PURCHASE_ORDER.PO_ID = LINE_ITEM.PO_ID) 
[...]

PR #54, opened 2015-12-16
include received on-order pos, on-site access update

### LINKED ISSUE #40: Availability for bibs without item records and/or Call numbers
Currently the availability service only returns availability info for bibs that have item records with call numbers, even if they have holding records. Have the service return record status for these 4 situations:
 - [x] elf* holdings are 'availabile' through the 856 electronic access link in the bibliographic record
 - [x] on-order records are available for request, but aren't linked to a particular holding location.
 - [ ] A lot of records like [this one](http://bibdata.princeton.edu/bibliographic/1917173/holdings) are in [non-circulating locations](http://bibdata.princeton.edu/locations/holding_locations/spc). Consider an additional status code like 'available only at the library'
 - [x] Items that were cataloged straight to ReCAP or have item records w/o a call number for whatever reason should still be considered availabile. [E.g](http://bibdata.princeton.edu/bibliographic/9153352/items)

tampakis:
- Include status code when retrieving on-order items. Expand "approved records" to also include partially and completely received orders. Attempts to address #47. 
 - Update availability API to distinguish between "On-Order" and "Order Received" when item record is lacking.
 - always_requestable locations (like rare books) should always have "on-site access" availability regardless of the item status message (if any). Closes #40.

### DIFF
--- a/spec/controllers/availability_controller_spec.rb
+++ b/spec/controllers/availability_controller_spec.rb
@@ -46,16 +46,23 @@ RSpec.describe AvailabilityController, :type => :controller do
       expect(availability[bib_online][holding_id]['status']).to eq('Online')
     end
 
-    it 'on-order records have a status of requestable' do
-      bib_on_order = '9160439'
-      holding_id = '9040953'
+    it 'on-order records have a status of On-Order' do
+      bib_on_order = '9173362'
+      holding_id = '9051785'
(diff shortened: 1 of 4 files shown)


kevinreiss:
:+1: :+1: These will definitely make easier to sort out these various edge cases.

Issue #53, opened 2015-12-16
Log Bibs with Unknown Statuses


tampakis:
"Unknown" is the fallback status for bibs not online, in an open location, not on-order, or without item records. We should log/compile a list of all of these "unknown" bibs and figure out if we can actually assign them a status. Some examples:
https://pulsearch.princeton.edu/catalog/9167195
https://pulsearch.princeton.edu/catalog/9292387

tampakis:
@joycebcat what should we do with records that don't have approval for purchase? https://pulsearch.princeton.edu/catalog/9468676
Should they be in the catalog at all?
Voyager includes "In the Pre-Order process" for records with a line_item_status of 0.

joycebcat:
@tampakis "In the Pre-Order process" should be treated just like "on order" (once we decide how those should be treated).

tampakis:
@joycebcat - I ran a report trying to log all of the bibs that return "Unknown" as its status. The query I wrote ran too slowly and after 24 hours of processing only got through 10% of all the records. I'm attaching the results here so maybe we can find some common patterns. After processing about 750,000 bib Ids it found 14,000 unknowns. https://www.dropbox.com/s/z853k7n40fk5nbh/unknown_mfhds_1.txt?dl=0

mzelesky:
I think we can develop a couple of queries to find them in Oracle. I'll work on that. First, I'll find all bibs with holdings in open locations with no attached item records. Then, from there, I'll see which of those do not have order records attached to them. What is the list of non-open locations?

tampakis:
I match it against all of the "always_requestable" locations here: https://bibdata.princeton.edu/locations/holding_locations

joycebcat:
Thanks Nikitas and Mark.  I think from this initial list patterns will start to show up.  I can't remote into my new computer right now so I don't think I'll get to check this out until tomorrow.

mzelesky:
I did an Oracle query on this and I will post a full file tomorrow. There are a lot more than I expected.

mzelesky:
I have attached an Excel spreadsheet that shows the extent of 'unknown' status bibs. The SQL Query is included, but I will also provide it below. There were a total of 454,672 bibs and 459,492 MFHDs.
[unknownstatusbibs.xlsx](https://github.com/pulibrary/marc_liberation/files/109878/unknownstatusbibs.xlsx)


WITH Step1 AS (
[...]

PR #55, opened 2015-12-17
allow all orders to be returned in items response

### LINKED ISSUE #47: On Order status isn't available in items JSON

tampakis:
- It's useful to figure out on-order information for all bibs that don't have item records. I changed the "approved_orders" query to return the on-order info regardless of approval status. The order statuses appear when requesting a bib's item record (mainly for debugging). Closes #47.

 -  The availability API will prune out non-approved orders. Although we might want to consider including pending records, as they are included in the Voyager OPAC. See #53.

### DIFF
--- a/spec/controllers/availability_controller_spec.rb
+++ b/spec/controllers/availability_controller_spec.rb
@@ -52,7 +52,6 @@ RSpec.describe AvailabilityController, :type => :controller do
       holding_id = '4847980'
       get :index, ids: [bible], format: :json
       availability = JSON.parse(response.body)
-      puts availability
       expect(availability[bible][holding_id]['status']).to eq('Limited')
     end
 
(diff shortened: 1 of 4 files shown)


tampakis:
Based on @joycebcat's [comment](https://github.com/pulibrary/marc_liberation/issues/53#issuecomment-165565849), I just pushed another commit that also includes the pending orders.

PR #56, opened 2015-12-18
Added 2 queries for course reserve information

mzelesky:
active_courses generates a list of all active courses with their list ID.

course_bibs populates each list with its associated bibs. if item_id is sufficient, reserve_list and reserve_list_items would give all the information needed; this would eliminate the need for a GROUP BY.


tampakis:
Thanks @mzelesky!

PR #59, opened 2016-01-19
reference locations, on-order for online and limited

tampakis:
Records that are at a reference location should have on-site access.
Records that are limited access or online access should also check first to see if there are order records for the item before assuming a status of 'limited' or 'online.'


kevinreiss:
:+1:

PR #68, opened 2016-02-23
Moving Traject from orangeindex, adding Solr view of bib records

escowles:


tpendragon:
:astonished: Not it

escowles:
Well, I think we're going to get a ton of commits for the orangeindex code that's coming over -- we want to keep that history, so we'll need to take all those commits.

But I'll squash my commits at least and see if that messes up the orangeindex commits at all.

jpstroop:
Yeah. I think squash HEAD~4 and call it good. Maybe we should add a deprecation note to the orangeindex README for a while, just to keep it around for a bit, but I suspect anything that's broken by this will get fixed along the way.

escowles:
Closing in favor of #69

PR #69, opened 2016-02-23
Moving Traject from orangeindex, adding Solr view of bib records (second try)

escowles:


jpstroop:
Did the specs make it over?

https://github.com/pulibrary/orangeindex/tree/development/spec

escowles:
I knew it was too easy -- I did not copy the specs.  Let me retry with keeping the specs.

jpstroop:
_wishes for Harry Potter emoji_

PR #70, opened 2016-02-24
Moving Traject from orangeindex, adding Solr view of bib records

escowles:


jpstroop:
I'm :+1: but I think @tampakis should take a look as well. Let's put a deprecation note in the orangeindex README for now too, rather than just wiping it out. That way if anything was missed or broken...

tampakis:
Looks good to me

PR #74, opened 2016-02-24
Adding IIIF view of Solr record

### LINKED ISSUE #58: Make a service for retrieving metadata as IIIF
See [the spec](http://iiif.io/api/presentation/2.1/), but basically it's like this:

```json
[
  { "label": "Author", "value": "Anne Author" },
  { "label": "Published", "value": [
    {"@value": "Paris, circa 1400", "@language": "en"},
    {"@value": "Paris, environ 1400", "@language": "fr"}
  ]},
  {"label": "Source", "value": "<span>From: <a href=\"http://example.org/db/1.html\">Some Collection</a></span>"}
]
```

Traject has all the formatting we need for MARC fields, so it would be nice if the could be re-used somehow. @tampakis ?

escowles:
Closes #58 

@tampakis Here is a pretty simple implementation of generating IIIF-style JSON from the Solr doc.  This doesn't handle some of the complications we've talked about with multiple titles in different languages.  Want to point me to some sample records? Or punt on that for another ticket?

### DIFF
--- a/app/models/iiif_record.rb
+++ b/app/models/iiif_record.rb
@@ -1,0 +1,53 @@
+class IIIFRecord
+  def initialize(solr_doc)
+    @solr_doc = solr_doc
+  end
+
+  def to_json
+    { label: label, description: description, metadata: metadata }
+  end
+
+  def contributors
+    return [] unless @solr_doc['related_name_json_1display']
+    JSON.parse(@solr_doc['related_name_json_1display'].first).values.flatten.uniq
+  end
+
+  def description
+    (@solr_doc['summary_note_display'] || [""]).first
+  end
+
+  def label
+    (@solr_doc['title_citation_display'] || [""]).first
+  end
+
+  def metadata
+    metadata = []
+
+    metadata_map.each do |solr_key, metadata_key|
+      values = @solr_doc[solr_key.to_s] || []
+      values = values.first if values.size == 1
+      metadata << { label: metadata_key, value: values } unless values.empty?
+    end
+
+    contrib = contributors
+    metadata << { label: 'contributor', value: contrib } unless contrib.empty?
+
+    metadata
+  end
+
+  def metadata_map
+    {
+      author_display:        'creator',
+      pub_date_display:      'date',
+      description_display:   'description',
+      edition_display:       'edition',
+      format:                'format',
+      genre_facet:           'genre',
+      language_facet:        'language',
+      language_code_s:       'language_code',
+      notes_display:         'note',
+      pub_citation_display:  'publication',
+      subject_facet:         'subject'
+    }
+  end
+end
--- a/spec/models/iiif_record_spec.rb
+++ b/spec/models/iiif_record_spec.rb
@@ -1,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe IIIFRecord, :type => :model do
+  let(:solr_doc) {{ "foo" => ['bar'] }}
+  let(:solr_doc) {{
+    "title_citation_display"     => ['This is the Title'],
+    "summary_note_display"       => ['This is a note about it.'],
+    "pub_date_display"           => ['1970'],
+    "language_facet"             => ['English', 'Spanish'],
+    "language_code_s"            => ['eng'],
+    "author_display"             => ['Author, Alice'],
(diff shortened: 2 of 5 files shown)


app/models/iiif_record.rb
```diff
@@ -0,0 +1,51 @@
...
+  def description
+    (@solr_doc['summary_note_display'] || [""]).first
+  end
+
+  def label
+    (@solr_doc['title_citation_display'] || []).first
+  end
+
+  def metadata
+    map = {
```
tpendragon: Extract to instance method

spec/models/iiif_record_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    "author_roles_1display"  => ['{"secondary_authors":[],"translators":["Translator, Bob"],"editors":[],"compilers":[],"primary_author":"Author, Alice"}']
+  }}
+  subject { described_class.new solr_doc }
+  it 'produces iiif json' do
+    iiif_json = {
+      label: 'This is the Title',
+      description: 'This is a note about it.',
+      metadata: [
+        { label: 'date', value: '1970' },
+        { label: 'language', value: ['English', 'Spanish'] },
```
tpendragon: I'm not sure if this is quite right. @jpstroop Everything in the IIIF presentation spec uses resources for the values of an array in the metadata bucket. Do you need something like
```json
{
  label: 'language',
  value: [
    {
      '@value': 'English'
    },
    {
       '@value': 'Spanish'
    }
  ]
}
```

?

escowles: Some clarification would be good.  I see two permutations in the spec:

The first is a single simple value:

```
{ "label": "Author", "value": "Anne Author" }
```

The second is multiple complex values:

```
      { "label": "Published", "value": [
        {"@value": "Paris, circa 1400", "@language": "en"},
        {"@value": "Paris, environ 1400", "@language": "fr"}
      ]
    }
```

There's no example for how to handle multiple simple values (nor a single complex value, for that matter).

tpendragon: Single complex value would be a resource not in an array - there's similar things elsewhere in the spec.

I'm hoping what I posted is the right way to do it, because that's what Plum does now, and there's a ticket in UniversalViewer to handle not having a language tag.

jpstroop: We've never been that clear about this in the spec because there are so many permutations. There was a [thread on IIIF discuss recently](https://groups.google.com/d/msg/iiif-discuss/OAa1uhaIJ6s/H7aCdgDkAAAJ).

I don't see why you'd need all those extra nodes if you're not going provide any meta-metadata.

```
{
  label: 'language',
  value: [
    {
      '@value': 'English'
    },
    {
       '@value': 'Spanish'
    }
  ]
}
```

app/models/iiif_record.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def to_json
+    { label: label, description: description, metadata: metadata }
+  end
+
+  def authors
+    authors_json = (@solr_doc['author_roles_1display'] || []).first
+    return {} unless authors_json
+    JSON.parse(authors_json)
+  end
```
tampakis: I would probably use the `author_display` which is for the the primary author (Marc 100/110/111) and its accompanying localized script version. Not all records have a primary author. If we have a preference for localized scripts, it always appears after the romanized version. 

How interested are we in creator roles (and including more than one creator)? I don't think the `author_roles_1display` field is reliable. The roles included there are supposed to conform to various citation standards, but the `related_name_json_1display` actually has a more accurate classification of roles. 

If there is a more specific role than "author" encoded for the primary author, it is stored in the `marc_relator_display` field. Gotta love Marc.

escowles: We can probably just map the primary creator to dc:creator and the other names to dc:contributor.  So I'll switch to mapping `author_display` to creator and map the other names from `related_name_json_1display` to contributor.

tampakis: Sounds right.

escowles:
I've updated this PR to change how authors are handled, to just have creator and contributors.

So far, there isn't any attempt to figure out the language/script of the values.  I've got some sample data to look at, so I'll see what I can come up with tomorrow morning.  I'm fine including that in this PR or doing that in a separate issue.

tampakis:
I think it makes sense doing the language script stuff in a separate issue.

PR #101, opened 2016-02-29
voyager_helpers gem, location logic

tampakis:
Use extracted voyager_helpers gem. Removed local path copy.
Moved circulating location based availability logic in availability controller.


config/initializers/voyager_helpers.rb
```diff
@@ -1,5 +1,4 @@
 require 'yaml'
-require_relative '../../voyager_helpers/lib/voyager_helpers'
 include VoyagerHelpers::OracleConnection
 
 vh_config = YAML.load_file("#{File.dirname(__FILE__)}/../voyager_helpers.yml")
```
escowles: Standalone usage of voyager_helpers is going to need this too, right?  So maybe the creation of this initializer (or at least docs about doing that) should be in there?

jpstroop: There are instructions with the new gem. Maybe just ref. those?

https://github.com/pulibrary/voyager_helpers/blob/development/README.md#configuration

-Js

Sent via mobile. Please excuse typos, brevity, etc.

escowles: This initializer is quite different from those instructions, though — parsing a YAML file instead of just doing it in ruby, assigning to vh_config instead of using VoyagerHelpers.configure.  Should we align marc_liberation with the config instructions?

jpstroop: It does feel a little convoluted ([this](https://github.com/pulibrary/marc_liberation/blob/development/config/initializers/voyager_helpers.rb)), though I suppose it's a valid approach ([I say since I did it](https://github.com/pulibrary/marc_liberation/blame/development/config/initializers/voyager_helpers.rb) :frowning: ). I've never really known which was better:

 1. Put that kind of config data in YAML so that (in a perfect world) the sysadmin deploying the app doesn't have to touch Ruby
 2. Put it in Ruby. Sysadmin deals.
 3. The application assumes some environment variables.

I'm OK w/ any of the above, but we should be consistent.

I think keeping the instructions somewhat vague in voyager_helpers is fine, since not everyone using it will be in Rails land.

jpstroop:
And there's a typo in the YAML file: https://github.com/pulibrary/marc_liberation/blob/development/config/voyager_helpers.yml.tmpl#L2

config/initializers/voyager_helpers.rb
escowles: I think the best option is 2 + 3: a ruby file that uses environment variables.  The sysadmin doesn't need to edit the file at all, and developers can edit the file locally if that's easier than setting env variables.

escowles:
At least there's a matching typo in the initializer: https://github.com/pulibrary/marc_liberation/pull/101/files#diff-35400167c485eed19291009815f8a71fL8

config/initializers/voyager_helpers.rb
jpstroop: that seems to be where we're added with the app_configs dir in Plum. :+1:

jpstroop:
We need to make sure the env vars are set on the server before deploying!

Issue #30, opened 2015-08-26
Configure resque


tampakis:
- [ ] Make daily updates higher priority than full dumps
- [ ] Have resque start up on system reboot
- [ ] Automate resque requeueing of failures? Mainly caused by the voyager daily reboot

Currently resque is run/configured as follows:

```
PIDFILE=./resque.pid COUNT=5 BACKGROUND=yes QUEUE=* bundle exec rake resque:workers >> log/resque.log 2>> log/resque.log
```

kevinreiss:
https://github.com/pulibrary/plum/blob/master/config/deploy.rb#L61 use upstart scripts. Need workers to be refreshed upon shutdown/startup.

tampakis:
We don't use resque any more.

PR #107, opened 2016-03-03
[DO NOT MERGE] rake tasks to start/stop workers

tampakis:
Restart resque workers on cap deploy. Advances #30.

### DIFF
--- a/Capfile
+++ b/Capfile
@@ -25,3 +25,4 @@ require 'capistrano/rails'
 
 # Load custom tasks from `lib/capistrano/tasks' if you have any defined
 Dir.glob('lib/capistrano/tasks/*.rake').each { |r| import r }
+Dir.glob('lib/tasks/workers.rake').each { |r| import r }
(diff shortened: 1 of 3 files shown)


jpstroop:
This seems :+1: to me, but would like @escowles or @tpendragon to have a look.

tpendragon:
Only concern I'd have is this is a bit different than how the other projects do it, which rely on upstart scripts.

jpstroop:
Is this ever going to be merged, or should we close it?

kevinreiss:
@tampakis we need to work out how we want to manage a queue for this application.

jpstroop:
Isn't there a queue for generating the subject browse lists already?

jpstroop:
If starting/stopping workers is the issue we should do whatever the proc machines are doing and keep things consistent.

escowles:
Yes, we should probably do this with cap like we do in plum: https://github.com/pulibrary/plum/blob/master/config/deploy.rb#L61

tampakis:
Woah we can finally close this PR, we don't use resque anymore!

PR #108, opened 2016-03-14
temp location display text in availability

tampakis:
Item temporary location codes are converted to their full location display text with library and holding location label (if not blank) in the availability service. This will allow Blacklight to display on-reserve locations in the search results and record show view.


eliotjordan:
:sunglasses:

PR #124, opened 2016-06-02
update to locations engine v0.3.0

tampakis:
Enhances table view to allow for sorting and establishes "holding library" relationship for ReCap location codes that have associated on-campus library locations.


kevinreiss:
👍 Having this realized in the search indexes for OL will address a number of long-standing requests from branch libraries that we were never able to tackle.

Issue #63, opened 2016-02-09
Patron Group Review


kevinreiss:
Review Voyager Patron groups in the context of patrons who have an active barcode in more than one group. Also think about mapping the rather messy string labels to something more digestible for an API call.  

See https://docs.google.com/a/princeton.edu/spreadsheets/d/1L0o8XDh2wJ27El8Da0Fh5nv6uxieV15uE9g1_K9etng/edit?usp=sharing for group names and current usage levels.

The response should provide the highest privilege group the patron is attached to.

kevinreiss:
A stat code of "EL" indicates someone is a Library Staff member. Patrons can have multiple stat codes attached to them.

kevinreiss:
On Alma this is moot.

PR #130, opened 2016-06-13
Advance voyager_helpers to 0.2.0. Add route for patron stat codes lookup

kevinreiss:
Advances #63. "stat" codes actually allow us to see what specific borrowing groups a person belongs to. The "patron_group" value is way too broad. @mzelesky could you post an excel spreadsheet to our shared drive for the OL group that details how to interpret the various patron "stat" codes that are now returned by this route. Thanks.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,7 @@ group :development, :production do
   gem 'ruby-oci8', '~> 2.2.1'
 end
 
-gem 'voyager_helpers', git: 'git@github.com:pulibrary/voyager_helpers.git', :tag => 'v0.1.2'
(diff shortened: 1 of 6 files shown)


mzelesky:
There are several stat codes that I don't understand, but I'll post a spreadsheet of the ones I know. Perhaps other staff can add to it after I make the preliminary one.

kevinreiss:
Sounds good @mzelesky.

mzelesky:
I have added the spreadsheet to the Blacklight folder. It is called 'Patron Stat Codes'. I don't know what a lot of the codes mean, and some have hundreds of patrons attached. Joan would be the one to know, if anyone does.

PR #140, opened 2016-06-27
Updating JSON-LD predicate mapping.

escowles:
Adding call number and adjusting edition/physical description predicates for https://github.com/pulibrary/plum/issues/415

### DIFF
--- a/app/models/jsonld_record.rb
+++ b/app/models/jsonld_record.rb
@@ -4,7 +4,7 @@ class JSONLDRecord
   end
 
   def to_h
-    metadata = { title: title, description: description, language: language_codes }
+    metadata = { title: title, language: language_codes }
 
     metadata_map.each do |solr_key, metadata_key|
       values = @solr_doc[solr_key.to_s] || []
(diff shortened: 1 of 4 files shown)


app/models/jsonld_record.rb
```diff
@@ -92,7 +93,8 @@ def title_language
   def metadata_map
     {
       author_display:        'creator',
-      description_display:   'description',
+      call_number_browse_s:  'call_number',
```
tampakis: I'm not totally in the loop here, but I think we want to use the `call_number_display` field instead. The `call_number_browse_s` field is used for sorting the call number browse list in Orangelight and doesn't include the full call number (it excludes the subfield $c). Here's an example where the values are different: https://bibdata.princeton.edu/bibliographic/3785401/solr

escowles: @tampakis Do you know why there are sometimes multiple identical call numbers in `call_number_display`?  e.g., https://bibdata.princeton.edu/bibliographic/1234567/solr

tampakis: That record has holdings at multiple locations, each with the same call number. However, there's no utility to having duplicate values there (I think there was at one point and the call number needed to display twice). If you get rid of the [allow_duplicates option](https://github.com/pulibrary/marc_liberation/blob/development/marc_to_solr/lib/traject_config.rb#L967) for the field, it should work fine.

PR #144, opened 2016-06-29
make name-title string field, include all uniform title fields

tampakis:


tampakis:
Updated based on chat with @joycebcat.
Current approach for browse list:
100+240/245a (only use 245a if there is no 240)
700/10/11 (t must be present)
76x/77x/78x at (a and t must be present)
800/10/11

PR #147, opened 2016-07-26
Add query interface for course reserves

tpendragon:


escowles:
I've merged `voyager_helpers/add_course_reserve_methods` - do we want to make a release and update this to use that?

jpstroop:
:+1: if @tpendragon doesn't have anything else to add.

tpendragon:
@escowles Updated to released version.

PR #155, opened 2016-08-26
Advance locations gem to 0.4.0.

kevinreiss:
After this is merged Orangelight's base URL for stackmap routing needs to be pointed at ```/locations/map```. Also need to point the main catalog at this new base URL. This allows us to kill the old shim that has a dependency on Primo's API.


jpstroop:
@tampakis - can you take care of this later today?

PR #162, opened 2016-09-29
dedup series display field, remove non-filing indicators for within series search

tampakis:
Dedup series 490. Closes pulibrary/orangelight#622
Series starts with search takes filing indicator into account. Closes pulibrary/orangelight#730.


kevinreiss:
👍

PR #180, opened 2016-11-22
[WIP] Barcode lookup

tampakis:


app/controllers/concerns/formatting_concern.rb
```diff
@@ -36,6 +36,13 @@ def pass_records_through_xml_parser(records)
     end
   end
 
+  def valid_barcode(barcode)
+    valid_barcode = false
+    if barcode =~ /^(32101[0-9]{9}|PULTST[0-9]{5})$/
+      valid_barcode = true
+    end
+  end
```
tampakis: This method should not return `nil` when the barcode is invalid.  I think the method can be simplified to:
```
def valid_barcode(barcode='')
  barcode =~ /^(32101[0-9]{9}|PULTST[0-9]{5})$/
end
```

spec/controllers/barcode_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      let(:valid_barcode2) { 'PULTST12345' }
+      it 'returns true for both cases' do
+        expect(described_class.valid_barcode(valid_barcode1)).to be true
+        expect(described_class.valid_barcode(valid_barcode2)).to be true
+      end
+    end
+    context 'barcode is correct length but not valid' do
+      let(:invalid_barcode_proper_length) { '31101123456789' }
+      it 'returns nil' do
+        expect(described_class.valid_barcode(invalid_barcode_proper_length)).to be_nil
```
tampakis: once the valid_barcode method gets updated, this should return false

spec/controllers/barcode_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+    end
+    context 'barcode is not proper length' do
+      let(:invalid_barcode_improper_length) { '321011234567890' }
+      it 'returns nil' do
+        expect(described_class.valid_barcode(invalid_barcode_improper_length)).to be_nil
```
tampakis: Should return false.

spec/controllers/barcode_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+    end
```
tampakis: Add a test case for when `nil` is sent to the valid_barcode method. It should also return false.

spec/requests/barcode_gets_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+  end
+  describe "GET /barcode/321010702214" do
+    it "returns an error when the barcode is not a valid form" do
+      allow(VoyagerHelpers::Liberator).to receive(:get_records_from_barcode).and_return([])
+      get '/barcode/321010702214'
+      expect(response.status).to be(404)
+    end
+  end
+  describe "GET /barcode/32101070300319" do
```
tampakis: This test case is unnecessary because it's the same situation as the one above - when the `get_records_from_barcode` method returns and empty array the status should be 404.

PR #183, opened 2016-11-29
Using type instead of genre

### LINKED ISSUE #182: JSONLD uses undefined "genre" property
There is no "genre" property in DC or DC Terms — we should be using dc:type instead.

escowles:
Fixes #182


jpstroop:
Merging, will make a new ticket.

PR #187, opened 2016-12-01
Add barcode lookup route

### LINKED ISSUE #62: Lookup by Barcode
Looking up an item by barcode is a use-case for the ReCAP middleware project. Will support the ReCAP accession process for newly acquired materials.

mzelesky:
Closes #62.


mzelesky:
There is a bug with the trim_punctuation feature in the latest version of traject (2.3.2). For now, we have rolled our version of traject back to 2.3.1.

app/controllers/concerns/formatting_concern.rb
```diff
@@ -36,6 +36,10 @@ def pass_records_through_xml_parser(records)
     end
   end
 
+  def valid_barcode(barcode='')
+    barcode =~ /^(32101[0-9]{9}|PULTST[0-9]{5})$/
```
tampakis: Could the method be renamed `valid_barcode?`.
Also the method should either return true or false, not 0 or nil. Try this to get a boolean response:
```
(barcode =~ /^(32101[0-9]{9}|PULTST[0-9]{5})$/) == 0
```

mzelesky:
Ok, I have made that change.

tampakis:
Cool

PR #208, opened 2017-02-06
load traject config in initializer

tampakis:
The memoization for loading the traject config doesn't work in the controller, so I moved it to an initializer. Speeds up the load time for the /solr and /jsonld views significantly.


escowles:
👍  cuts the time for me in half (~1.5 sec instead of ~3.0 sec per request)

PR #209, opened 2017-02-07
process locations from db not from web service

tampakis:
Deploying #208 failed because Traject needs to download the [locations](https://github.com/pulibrary/marc_liberation/blob/master/marc_to_solr/lib/location_extract.rb#L7) from bibdata, which is impossible when the web server isn't already running.

This PR updates the initializer to update the location data by accessing the locations tables using Active Record. In the test environment there are no locations to load so Circle CI uses the locations template data instead.

I deployed this branch and it is running successfully in production right now.


kevinreiss:
👍

Issue #212, opened 2017-02-16
404 Error when Patron route receives an ID it can't find.


kevinreiss:
Currently when an ID that can't be found in Voyager is passed (https://bibdata.princeton.edu/patron/blah) you get a 500 error returned. Should we return a 404 error instead?

jpstroop:
Yes. I feel like that was the case at some point, so this may be a bug.

kevinreiss:
Yeah, looks like https://github.com/pulibrary/marc_liberation/blob/0c945be3215c6ecb9025a1edad268604f9a219e6/app/controllers/patron_controller.rb#L6 we need to test for nil? rather than empty?.

kevinreiss:
Fixed

PR #213, opened 2017-02-16
Respond with 404 when user doesn't exist

kevinreiss:
@tampakis We sometimes test using ```.nil?``` to see if a VoyagerHelpers::Liberator query has no data, other times we test for ```.empty?```.  I marked this as WIP for now pending your comment. Moving the patron tests over to ```.nil?``` resolved the immediate issue in #212.


jpstroop:
As this is a rails app, wouldn't it be safest to use `.blank?` ?

kevinreiss:
Sure. That seems the safest test condition to use in this context.

kevinreiss:
This should be good now.

PR #214, opened 2017-02-23
adds coverage to jsonld

tampakis:
Advances pulibrary/plum#1004.

### DIFF
--- a/marc_to_solr/lib/geo.rb
+++ b/marc_to_solr/lib/geo.rb
@@ -1,0 +1,30 @@
+# geo.rb
+# extract geo-related data from MARC
+
+def decimal_coordinate record
+  coverage = []
+  Traject::MarcExtractor.cached('034defg').collect_matching_lines(record) do |field, spec, extractor|
+    c = {}
+    field.subfields.each do |s_field|
+      c['w'] = s_field.value if s_field.code == 'd' and valid_coordinate_format?(s_field.value, record)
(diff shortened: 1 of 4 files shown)


public/context.json
```diff
@@ -23,6 +23,7 @@
     "source": {"@id": "dc:source"},
     "subject": {"@id": "dc:subject"},
     "type": {"@id": "dc:type"},
+    "coverage": {"@id": "dc:coverage"},
```
eliotjordan: `coverage` is in this list twice. See line 13.

tampakis: nice catch! updated

marc_to_solr/lib/geo.rb
```diff
@@ -0,0 +1,30 @@
...
+      break
+    end
+    coverage << "northlimit=#{c['n']}; eastlimit=#{c['e']}; southlimit=#{c['s']}; westlimit=#{c['w']}; units=degrees; projection=EPSG:4326"
+  end
+  logger.error "#{record['001']} - multiple 034s" if coverage.length > 1
+  coverage.first
+end
+
+def valid_coordinate_format? c, record
+  unless c =~ /^[WENS]/ || c =~ /[-+]?([0-9]*\.[0-9]+|[0-9]+)/
```
eliotjordan: Looks like this is returning coverages with degree minute seconds. Is that true?

PR #226, opened 2017-04-03
allow rake db migrate to run when locations table does not yet exist

tampakis:
@kevinreiss this should be familiar to you


kevinreiss:
Thanks @tampakis

PR #228, opened 2017-04-17
extend json-ld with geo-related fields

tampakis:
Adds alternative title, scale, and projection fields to json-ld service. Advances pulibrary/plum#1160 and pulibrary/plum#1161.

### DIFF
--- a/public/context.json
+++ b/public/context.json
@@ -2,6 +2,7 @@
   "@context": {
     "id" : "@id",
     "bf": "http://id.loc.gov/ontologies/bibframe/",
+    "bf1": "http://bibframe.org/vocab/",
     "dc": "http://purl.org/dc/elements/1.1/",
     "dcterms": "http://purl.org/dc/terms/",
     "edm": "http://www.europeana.eu/schemas/edm/",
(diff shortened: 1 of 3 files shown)


public/context.json
```diff
@@ -35,7 +36,10 @@
     "title": {"@id": "dcterms:title"},
     "abstract": {"@id": "dcterms:abstract"},
 
+
     "edition": {"@id": "bf:editionStatement"},
+    "projection": {"@id": "bf:cartographicProjection"},
+    "scale": {"@id": "bf:cartographicScale"},
```
jpstroop: Are the bibframe predicates really that unencumbered? I'm shocked.

escowles: Alas, projection and scale both look like object properties that point to classes instead of literals:

http://id.loc.gov/ontologies/bibframe.html#c_Projection
http://id.loc.gov/ontologies/bibframe.html#c_Scale

tampakis: So in the rdf-vocab gem the range is listed as [literal](https://github.com/ruby-rdf/rdf-vocab/blob/731d4d43ad5819a4a412893d062ca4426a10c92e/lib/rdf/vocab/bibframe.rb#L488). Could we still get away with using it?

escowles: Oh, I think the issue is that those predicates are defined in Bibframe 1 (http://bibframe.org/vocab/) — not Bibframe 2 (http://id.loc.gov/ontologies/bibframe/).

So I think you just need to define a `bf1` or whatever namespace for the Bibframe 1 predicates, and then that should be fine.

tampakis: Got it thanks!

jpstroop:
I'm :ok: with this if necessary, but are there really not better predicates?

PR #239, opened 2017-05-31
display status of inaccessible for lewis library locations

tampakis:
We'll have to add "Inaccessible" to the list of unavailable statuses in the orangelight availability js


kevinreiss:
Thanks @tampakis.

Issue #237, opened 2017-05-23
Need new barcode lookup for SCSB that includes suppressed records


mzelesky:
Some records in ReCAP have legitimate suppressed holdings (e.g., master copy of microforms, other dark storage), so suppressed bibs and holdings need to be returned for SCSB lookup. The current barcode lookup method is for "public" use, where one wouldn't want suppressed records to be returned, so another route needs to be created for SCSB lookup.

kevinreiss:
Add this under the route "/scsb". Need to create new query in Voyager Helpers so we can return suppressed records.

kevinreiss:
How is using ```/barcode/32101075573871/scsb```?

tampakis:
sounds good to me

Issue #171, opened 2016-10-24
General Barcode Lookup on Bibdata 


kevinreiss:
In addition to the work to provide a ReCAP friendly export by barcode https://github.com/pulibrary/marc_liberation/commits/barcode_lookup we should also provide a general lookup by barcode that produces a reasonable. The legacy getVoyRec in use in various utilities provides this lookup, http://libweb5.princeton.edu/getvoyrec/.

mzelesky:
The method I have created works for all barcodes. The one difference is that ReCAP materials have the 852 $h and $i merged and additional subfields are added to the 876 for ReCAP-specific information. Is that sufficient, or does a method need to be created that omits the extra ReCAP-specific fields even for ReCAP records?

kevinreiss:
I will try and identify if any applications actually utilize the barcode lookup present in GetVoyRec.

PR #247, opened 2017-07-07
Recap features

kevinreiss:
Addresses #237, finishes #171. @tampakis for your initial review next week. I think it would probably to get this merged and then work on bugfixes for the ReCAP stuff in small increments for the rest of the summer. This adds:

1. Some processing rules in our traject configuration to deal with data specific to the ReCAP project that will come with partner records. Item level data gets stuffed in the holdings display hash key.
2. A new dump job that produces a MARCXML file of updated recap records we'll need to ship to the new ReCAP system on a daily frequency.
3. A new route with a barcode lookup that deals with suppressed records and includes boilerplate data required by the new ReCAP system.


Gemfile
```diff
@@ -7,15 +7,15 @@ gem 'jquery-rails'
 gem 'turbolinks'
 gem 'jbuilder', '~> 2.0'
 gem 'sdoc', '~> 0.4.0', group: :doc
-gem 'mysql2'
+gem 'mysql2', '0.4.5'
```
tampakis: Is there a reason why we pin to a particular version?

app/controllers/barcode_controller.rb
```diff
@@ -9,6 +9,28 @@ def index
     end
   end
 
+  def scsb
+    unless valid_barcode?(params[:barcode])
```
tampakis: do an `if !valid_barcode?` if it is paired with an else

marc_to_solr/lib/traject_config.rb
```diff
@@ -188,8 +190,19 @@
     accumulator << record.end_date_from_008
 end
 
-to_field 'cataloged_tdt', extract_marc('959a') do |record, accumulator|
-  accumulator[0] = Time.parse(accumulator[0]).utc.strftime("%Y-%m-%dT%H:%M:%SZ") unless accumulator[0].nil?
+# to_field 'cataloged_tdt', extract_marc('959a') do |record, accumulator|
+#   accumulator[0] = Time.parse(accumulator[0]).utc.strftime("%Y-%m-%dT%H:%M:%SZ") unless accumulator[0].nil?
+# end
```
tampakis: Keep this field the same please. The logic for checking for scsb makes this unnecessarily complex.

marc_to_solr/lib/traject_config.rb
```diff
@@ -910,6 +923,16 @@
...
+  recap_notes = process_recap_notes(record)
+  unless recap_notes.empty?
+    recap_notes.each_with_index do |value, i|
+      accumulator[i] = value
+    end
+  end
+end
+
 each_record do |record, context|
   dissertation_note = context.output_hash['dissertation_notes_display']
```
tampakis: The cataloged_tdt value can be deleted here if the id matches a SCSB value.

spec/controllers/patron_controller_spec.rb
```diff
@@ -26,7 +26,7 @@
   end
 
   it "404 when patron info is not found" do
-    stub_patron('123456789')
+    #stub_patron('123456789')
```
tampakis: @kevinreiss any particular reason why this is commented out?

tampakis:
Thanks @kevinreiss

PR #253, opened 2017-08-01
process multiple item statuses, display highest priority one

tampakis:
Thanks for the work on this @mzelesky.


kevinreiss:
👍 Thanks to both of you.

PR #256, opened 2017-08-28
Remove inaccessible status

kevinreiss:
Helps address https://github.com/pulibrary/orangelight/issues/927. Construction project is over as of Today.

### DIFF
--- a/app/controllers/availability_controller.rb
+++ b/app/controllers/availability_controller.rb
@@ -114,19 +114,6 @@ class AvailabilityController < ApplicationController
     'On-Site'
   end
 
-  def inaccessible_status
-    'Inaccessible'
-  end
-
-  def inaccessible_locations
-    #%w(sci scith sciref scirefl scilaf scilal scimm)
-    []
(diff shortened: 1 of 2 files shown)


kevinreiss:
@tampakis When you are back let's check in on this. I'm not sure why unrelated specs started to fail when I tried to excise the code related to the "inaccessible" status.

app/controllers/availability_controller.rb
```diff
@@ -192,11 +180,7 @@ def update_item_loc(item)
     unless loc.nil?
       item[:label] = location_full_display(loc)
       unless order_status?(item[:status])
-        item[:status] = if inaccessible?(loc_code)
-          inaccessible_status
-        else
-          location_based_status(loc, item[:status])
-        end
+        location_based_status(loc, item[:status])
```
tampakis: by the way, it was error because `item[:status]` needs to get re set
```
item[:status] = location_based_status(loc, item[:status])
```

Issue #258, opened 2017-09-06
SCSB Records and the Unknown Format


kevinreiss:
Since we can't control the source data here we should keep the "Unknown" value from being added to the format facet.

kevinreiss:
@tampakis we've got this one covered. Closing.

PR #259, opened 2017-09-06
Keep the "Unknown" value out of format facet

kevinreiss:
For #258.


marc_to_solr/lib/traject_config.rb
```diff
@@ -936,7 +936,9 @@
 each_record do |record, context|
   dissertation_note = context.output_hash['dissertation_notes_display']
   if dissertation_note && dissertation_note.first.downcase.gsub(/[^a-z]/, '').include?("seniorprincetonuniversity")
-    context.output_hash['format'] << Traject::TranslationMap.new("format")['ST']
+    if context.output_hash['format'].is_a?(Array) && !context.output_hash['foramt'].nil?
```
tampakis: typo
`if !context.output_hash['format'].nil?`

PR #273, opened 2017-11-11
refactor index rake tasks to only commit once

tampakis:
@kevinreiss this change will have traject post a commit to solr only once per rake task call. Since the index is so large each commit takes about 2 minutes for solr to process. When running through hundreds of scsb update files this can cause the update process to take a very long time.

I cleaned up, refactored, and tested out each of the methods, but please take a close look when you get a chance.


kevinreiss:
Okay, I will check this out tomorrow morning.

kevinreiss:
Looks good @tampakis.

PR #302, opened 2017-12-20
Convert translation map files from template to regular file

### LINKED ISSUE #301: Convert marc_to_solr files from template to regular file
No need for these files to be templates.

https://github.com/pulibrary/marc_liberation/blob/master/marc_to_solr/translation_maps/holding_library.rb.tmpl

https://github.com/pulibrary/marc_liberation/blob/master/marc_to_solr/translation_maps/location_display.rb.tmpl

https://github.com/pulibrary/marc_liberation/blob/master/marc_to_solr/translation_maps/locations.rb.tmpl

eliotjordan:
Close #301


eliotjordan:
@tampakis Do these files change regularly? If so, should they be in Ansible?

tampakis:
These template files are only used for getting circle ci to pass without having to access bibdata.princeton.edu to access the locations info (which marc_liberation stores in it's database).

hackmastera:
@tampakis can you point to an example of how the files are used in the code? Is the same end not achievable via webmock?

tampakis:
Perhaps I made things more complicated than they need to be. Here is the PR where I introduced the template files: https://github.com/pulibrary/marc_liberation/pull/209. It was set up to avoid having this line called while running the tests: https://github.com/pulibrary/marc_liberation/blob/master/marc_to_solr/lib/location_extract.rb#L7

PR #304, opened 2017-12-20
ensures only unique values get indexed to solr

### LINKED ISSUE #303: Only index unique recap notes
Blacklight handles the display properly, but currently dozens of identical recap notes get indexed.

tampakis:
Closes #303. @kevinreiss or @escowles - can you think of an example where we'd ever want duplicate field values indexed to solr or exposed in the solr route on bibdata?


kevinreiss:
👍

PR #309, opened 2017-12-21
WIP - Transfer recap dump files to project server

kevinreiss:
Sets default period for recap submit collection dump to 1 day.


kevinreiss:
@tampakis Feels like we should probably catch a failure of the file transfer in a way this currently won't.

PR #315, opened 2018-01-02
Indexes OL electronic access ARK resources with resolved URLs

### LINKED ISSUE #314: Extend the handling for indexing electronic access URI's for cases where an Orangelight catalog record is recursively referenced
This addresses cases where ARK's resolve to Orangelight records imported from the same bibliographic record (wherein relative URL's are provided for the OL interface).  It also blocks the full resolution of https://github.com/pulibrary/orangelight/issues/1048

jrgriffiniii:
Resolves #314


marc_to_solr/lib/princeton_marc.rb
```diff
@@ -341,20 +343,63 @@ def oclc_normalize oclc, opts = {prefix: false}
...
+  redirected_uri = begin
+                     response = conn.get
+                     response.env.url
+                   rescue StandardError => err
+                     logger.error err
+                     uri
+                   end
+
+  # Determine whether or not this ARK resolves to an Orangelight resource
+  orangelight_host = 'pulsearch.princeton.edu' # @todo Resolve this properly (please @see https://github.com/pulibrary/marc_liberation/issues/313)
```
hackmastera: Maybe this piece should be done on the OL side?

marc_to_solr/lib/princeton_marc.rb
```diff
@@ -341,20 +343,63 @@ def oclc_normalize oclc, opts = {prefix: false}
   end
 end
 
+# Resolves ARK's throught executing an HTTP GET request
```
hackmastera: Typos! :)  should be "ARKs" and "through"

escowles:
My only concern with this is indexing speed: adding an HTTP request for each record may slow things down significantly.  Can we test this PR with a largish batch of records to see what the real-world impact is?

One alternative may be to fetch Figgy's entire bib/manifest mapping as a single request (back-of-envelope math says it would be ~2MB).  That map could be loaded into memory for much faster access.

jrgriffiniii:
> One alternative may be to fetch Figgy's entire bib/manifest mapping as a single request

I have attempted to implement this using a `CacheMap` Class (which is populated using Solr responses from Figgy).

tpendragon:
Since you have all the arks from Figgy in a cache, is it necessary to resolve the arks anymore? Can you just compare arks between the record/figgy and adjust the index accordingly?

marc_to_solr/lib/princeton_marc.rb
```diff
@@ -341,20 +343,208 @@ def oclc_normalize oclc, opts = {prefix: false}
...
+
+        @values[ark] = bib_id
+      end
+    end
+
+    # Query the service using the endpoint
+    # @param [Integer] the page parameter for the query
+    def query(page: 1)
+      begin
+        http_response = Faraday.get("#{@endpoint}.json&q=&rows=#{@rows}&page=#{page}")
```
escowles: Do we want to limit to records that have ARKs?  I think `f[identifier_tesim][]=ark` would do it.

jrgriffiniii: Thank you for identifying this, the query has been updated accordingly

jrgriffiniii:
- @tpendragon ARKs are no longer resolved, and are instead cached (along with the BibID's) from the Blacklight JSON endpoints exposed by the repositories
- @escowles I have updated this PR in order to feature a `CompositeCacheMap` which handles two separate Blacklight endpoints

marc_to_solr/lib/princeton_marc.rb
```diff
@@ -341,20 +343,213 @@ def oclc_normalize oclc, opts = {prefix: false}
...
+
+        @values[ark] = bib_id
+      end
+    end
+
+    # Query the service using the endpoint
+    # @param [Integer] the page parameter for the query
+    def query(page: 1)
+      begin
+        url = URI::HTTPS.build(host: @host, path: @path, query: 'q=&rows=1000000&page=1&f[identifier_tesim][]=ark')
```
escowles: The `page` parameter needs to be a variable instead of hard-coded as `1`, right?

PR #356, opened 2018-01-24
Use binwrite when saving gzipped marc files

### LINKED ISSUE #354: Error when downloading downloading marc files during indexing
`Encoding::UndefinedConversionError: "\x8B" from ASCII-8BIT to UTF-8` at 
https://github.com/pulibrary/marc_liberation/blob/master/marc_to_solr/lib/index_functions.rb#L39

eliotjordan:
Closes #354


hackmastera:
Good find!! Any idea why this resolves the issue?

eliotjordan:
@hackmastera I'm guessing it has something to do with using newer versions of ruby / jruby.

PR #371, opened 2018-02-26
Fixes the failing spec for #set_pub_created

jrgriffiniii:


tampakis:
Thanks @jrgriffiniii, when I ran the tests locally I totally had missed this. These specs are removed and replaced in #370

PR #374, opened 2018-03-08
add note index field

christinach:
Advances pulibrary/orangelight#1149


escowles:
Can you add a test to `marc_to_solr/spec/lib/config_spec.rb` showing that the notes from a sample record show up in the expected field?

christinach:
I will do, thanks.

christinach:
Test has been added @escowles.

PR #377, opened 2018-03-29
When more that one data in isbn qualifiers, adds : between q an…

### LINKED ISSUE #761: Load 'allow' items from full hathi report into HathiAccess
Note: does the full report have 'allow' information?? Answer: yes!

We could then use an oclc-based lookup to add links to, e.g., catalog show pages for NYPL items. This will be used in powering https://github.com/pulibrary/marc_liberation/issues/760

christinach:
fixes #761


christinach:
Thank you @jrgriffiniii

PR #380, opened 2018-04-04
Sort tag=264 on indicator2, add new field pub_created_vern_display for 880

christinach:
Sorts datafield with tag 264 on indicator2.     

https://github.com/pulibrary/orangelight/issues/1155


marc_to_solr/lib/traject_config.rb
```diff
@@ -186,6 +190,15 @@
       end
     end
   end
+  if record['264']
+    rec_264 = []
+    MarcExtractor.cached("264abcefg3").collect_matching_lines(record) do |field, _spec, _extractor|
+      rec_264 << field
```
tampakis: I think that the subfield 6 will be excluded if you replace the line with this:
```
  MarcExtractor.cached("264abcefg3").collect_matching_lines(record) do |field, spec, extractor|
    rec_264 << extractor.collect_subfields(field, spec).first
  end
```

christinach: Thanks @tampakis !

tampakis:
Looks good

PR #402, opened 2018-06-08
Ensuring that the ARK is extracted for JSON-LD regardless of the neighboring link text subfield value

### LINKED ISSUE #401: Identifier no longer showing in JSON-LD
See: https://bibdata.princeton.edu/bibliographic/10002695/jsonld

Our identifiers to reconcile report depends on that field existing. Is it gone for a particular reason? There's a references now..

jrgriffiniii:
Resolves #401


mzelesky:
Shouldn't the ark always be in the 856 subfield u?

jrgriffiniii:
@mzelesky understood, I can address that with another commit.  However, if that's not the case, how should that be handled (should it raise an error or fail silently, log an error, and not offer `identifier`)?

mzelesky:
If it's not in the subfield u, then the URL is not in the right place. That's the subfield where the address is supposed to be. If it's not in that subfield, I would say it shouldn't be considered as an electronic location candidate.

The subfield 3 was used appropriately in the original record to denote that the link only has an image of the table of contents.

PR #408, opened 2018-06-20
Adds language_iana_s field to support the IANA Language Subtag Registry

christinach:


christinach:
First step to fix pulibrary/requests#349

christinach:
@tampakis thank you for the _s !

christinach:
After checking the requests repo, to update the spec/cassettes with the new field, first we will have to index it and then run the test suite in requests to update the various spec/cassettes/ .yml files to include the new language_iana_s field.

PR #414, opened 2018-06-23
Language_iana_s accepts only valid ISO_639 elements

christinach:


christinach:
@tampakis regarding this record https://bibdata.princeton.edu/bibliographic/8543052
the test will fail with a nil value. In development the accumulator will return ["it"].

christinach:
@tampakis can you check if the value for the language_iana_s makes sense, considering the html lang label? Please check the record 
```
<controlfield tag="001">8543052</controlfield>
``` 
Thanks

christinach:
closing because @tampakis submitted some hours ago a PR for this issue.

PR #451, opened 2018-08-28
JSON-LD should not return finding aid ark.

tpendragon:


app/services/ark_resolver.rb
```diff
@@ -0,0 +1,26 @@
...
+    @location ||=
+      begin
+        return unless final_result && final_result.status == 302
+        final_result.headers["location"]
+      end
+  end
+
+  private
+
+    def initial_result
```
hackmastera: this method is called twice do you want to memoize?

app/services/ark_resolver.rb
```diff
@@ -0,0 +1,26 @@
...
+      return unless ark.present?
+      Faraday.head(ark)
+    end
+
+    def final_result
```
hackmastera: memoize in here also?

spec/models/jsonld_record_spec.rb
```diff
@@ -24,12 +24,14 @@
       'provenance_display'         => ['Provenance: Johann Anton André, of Offenbach, 1799; Philharmonische Gesellschaft, Laibach (Ljubljana); Robert Ammann.'],
       'source_acquisition_display'         => ['Obtained, Nov. 16, 1961, at Stargardt Sale of the collection of Dr. Robert Ammann.'],
       'references_display'         => ["Stillwell B460.", "Goff B-526."],
-      'electronic_access_1display' => ['{"http://arks.princeton.edu/ark:/88435/dr26z114k":["arks.princeton.edu"],"http://digital.lib.cuhk.edu.hk/crbp/servlet/list":["First page of main text"]}'],
+      'electronic_access_1display' => ["{\"http://arks.princeton.edu/ark:/88435/47429918s\":[\"arks.princeton.edu\",\"Finding aid\"],\"http://arks.princeton.edu/ark:/88435/7p88ch283\":[\"arks.princeton.edu\"]}"],
```
hackmastera: Maybe a comment here or in the spec since this doesn't have its own spec documenting the functionality

tpendragon:
@hackmastera Fixed! Thanks.

PR #457, opened 2018-09-10
Scsb event

tampakis:
- Splits "RECAP_RECORDS" dump type into "PRINCETON_RECAP" and "PARTNER_RECAP"
 - Creates partner recap update job to import updated marc records and deleted record ids from scsb server.
 - Jobs are set up to pull down all delete and update files since the last partner recap job was run.
 - Records are cleaned using marc_cleanup gem and fixes are logged in a log file attached to the partner recap dump object.
 - Files are copied to a separate directory within the marc liberation data mount.


kevinreiss:
Looks good.

PR #471, opened 2018-09-21
Genre subject browse

tampakis:
- Adds a subdivision along subfield $t for subjects. Closes pulibrary/orangelight#1159 
 - Includes genre field in the subject facet. Closes pulibrary/orangelight#1377
 - Handles punctuation and subdivisions in genre field using em dash (as is done with the subject field). Thanks for your spec @jrgriffiniii! Closes pulibrary/orangelight#1352


tampakis:
Based on Thursday's authorities, modifying the PR to only include gft terms in the genre field

PR #494, opened 2018-10-02
[WIP] Refactor diffs

### LINKED ISSUE #473: Refactor incremental updates
To process new/updated bibs based on the query described here: pulibrary/voyager_helpers#87

### LINKED ISSUE #477: For incremental update rake tasks, allow timestamp to be set
Most of the time when we run incremental updates, both from Voyager and from SCSB partner updates, we take all updated records from the last time the incremental update task was run. However, in times where there are unexpected errors, or the first time a process is run, it is helpful to be able to set a time/date and then process all updates since that time.
Example, updates from scsb since May: https://bibdata.princeton.edu/dumps/7970.json

tampakis:
Fixes #473; Fixes #477


mzelesky:
We are going to test in staging first. When we deploy to production, all events will have to be destroyed.

Issue #469, opened 2018-09-19
Cron jobs should use the app_configs for Solr URL


escowles:
Currently, the Solr URL is specified on the command line in the cron config for regular incremental indexing.  But that means that it has to be updated separately when the ansible config is changed for the webapp.  We should use the webapp config (`/home/deploy/app_configs/marc_libration`) instead of hard-coding this in the cron config.

escowles:
Either read from env variable, or add cron jobs to Ansible

hackmastera:
currently these are
```
irb(main):001:0> ENV['SOLR_URL']
=> "http://lib-solr.princeton.edu:8983/solr/catalog"
irb(main):002:0> ENV['SOLR_REINDEX_URL']
=> "http://lib-solr.princeton.edu:8983/solr/catalog-reindex"
```

but they should be
```
irb(main):001:0> ENV['SOLR_URL']
=> "http://lib-solr.princeton.edu:8983/solr/catalog-production"
irb(main):002:0> ENV['SOLR_REINDEX_URL']
=> "http://lib-solr.princeton.edu:8983/solr/catalog-rebuild"
```

hackmastera:
once crons are on whenever this should be removed from ansible:
https://github.com/pulibrary/princeton_ansible/blob/d7b3b44a1d712a4c14289d16e775e8f03d941224/roles/pulibrary.bibdata/tasks/main.yml#L18

hackmastera:
That ark cache seeder job doesn't appear to have been run on staging or production since march 19th and march 17th, respectively. I think because it's not setting `RAILS_ENV=production`. That will be the first job I move to whenever, then we can deploy and check on it before proceeding with more jobs.

PR #504, opened 2018-10-18
Configure whenever and add a first job

hackmastera:
refs #469


tampakis:
This looks great @hackmastera!

PR #506, opened 2018-10-22
Use whenever's built-in cap integration, migrate all crons

### LINKED ISSUE #469: Cron jobs should use the app_configs for Solr URL

hackmastera:
this can be deployed once we merge https://github.com/pulibrary/princeton_ansible/pull/358 and reprovision bibdata servers

fixes #469

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -5,7 +5,7 @@ set :application, 'marc_liberation'
 set :repo_url, "https://github.com/pulibrary/marc_liberation.git"
 
 # Default branch is :master
-ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call
+set :branch, ENV['BRANCH'] || 'master'
 
 # Default deploy_to directory is /var/www/my_app_name
 set :deploy_to, "/opt/#{fetch(:application)}"
@@ -41,6 +41,8 @@ set :linked_dirs, %w{
 
 # Default value for keep_releases is 5
 set :keep_releases, 5
+
+set :whenever_roles, ->{ [:cron, :cron_staging, :cron_production] }
 
 namespace :sidekiq do
   task :restart do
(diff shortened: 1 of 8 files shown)


config/schedule.rb
```diff
@@ -4,22 +4,80 @@
...
+  )
+end
+
+# 3 updates to catalog-production per day
+every 1.day, at: ["12:40am", "10:55am", "2:55pm"], roles: [:cron_production] do
+  liberate_latest_production(
+    "liberate:latest",
+    set_url: ENV["SOLR_URL"],
+    update_locations: "true",
+    output: "/tmp/daily_upates.log"
```
jrgriffiniii: `daily_upates.log` should be replaced with `daily_updates.log`

config/deploy.rb
```diff
@@ -5,7 +5,7 @@
 set :repo_url, "https://github.com/pulibrary/marc_liberation.git"
 
 # Default branch is :master
-ask :branch, proc { `git rev-parse --abbrev-ref HEAD`.chomp }.call
+set :branch, ENV['BRANCH'] || 'master'
```
escowles: I am 👍 on this change — and assume it works with pulbot because we do this in other projects.  There's nothing in the README about deployment now, but do you think it would be good to add something?

hackmastera: @escowles yeah, I changed this (copied from figgy) because I couldn't get it to actually ask me. not sure what is up with that. I just pushed a note to the README.

config/schedule.rb
hackmastera: Thank you!! Fixed.

hackmastera:
deployed to prod, double-checked resulting jobs, disabled other crons, watched one run successfully.

PR #512, opened 2018-10-29
Scsb index tasks

### LINKED ISSUE #244: Create job that consumes daily partner deletes for SCSB
Also https://uat-recap.htcinc.com:9093/swagger-ui.html#!/data-dump-rest-controller/exportDataDump

tampakis:
Closes #244.
 - Ensures that scsb partner updates are in chronological order
 - Makes rake tasks to either index the most recent recap partner job or all since a certain date
 - Schedules job using whenever

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -63,9 +63,9 @@ every 1.day, at: ["12:40am", "10:55am", "2:55pm"], roles: [:cron_production] do
 end
 
 # 3 updates to catalog-rebuild per day
-# These are 1 hour earlier than production to stagger the load on solr but ensure that
+# These are earlier than production to stagger the load on solr but ensure that
 #   no updates get lost when indices are swapped
-every 1.day, at: ["1:40am", "11:55am", "3:55pm"], roles: [:cron_production] do
+every 1.day, at: ["12:00am", "10:30am", "2:30pm"], roles: [:cron_production] do
   liberate_latest_production(
     "liberate:latest",
     set_url: ENV["SOLR_REINDEX_URL"],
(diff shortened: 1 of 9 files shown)


config/schedule.rb
```diff
@@ -65,7 +65,7 @@
 # 3 updates to catalog-rebuild per day
 # These are 1 hour earlier than production to stagger the load on solr but ensure that
 #   no updates get lost when indices are swapped
-every 1.day, at: ["11:40pm", "9:55am", "1:55pm"], roles: [:cron_production] do
+every 1.day, at: ["1:40am", "11:55am", "3:55pm"], roles: [:cron_production] do
```
hackmastera: why are you changing these? see comment.

lib/tasks/scsb.rake
```diff
@@ -9,4 +11,46 @@ namespace :scsb do
     %x[bundle exec loops stop 2>&1]
   end
 
+  desc "Index SCSB records with all changed records since SET_DATE, against SET_URL"
+  task updates: :environment do
```
hackmastera: these are pretty long for rake tasks. what do you think about putting the logic in code, so it can be tested, and then calling it from the task?

config/schedule.rb
tampakis: Thanks for the comment @hackmastera. I switched the timestamp because the indexing jobs need to happen after get_changes job. Those are scheduled at `["10:10am", "2:10pm", "10:35pm"]`.

lib/tasks/scsb.rake
tampakis: Makes sense. Easier to test.

config/schedule.rb
hackmastera: Ah, okay. I think it's important for rebuild to run before production, to protect against index entry loss during a swap. What if we use the times currently under `SOLR_URL` here on `SOLR_REINDEX_URL` and then make the ones on `SOLR_URL` be `["1:40am", "11:55am", "3:55pm"]`?

PR #519, opened 2018-12-19
Ensures that the latest voyager_helpers commit to the master branch is used

### LINKED ISSUE #515: Certain records trigger ArgumentError: comparison of Hash with Hash failed
## Backtrace

    line 86 of [PROJECT_ROOT]/app/controllers/availability_controller.rb: find_availability
    line 6 of [PROJECT_ROOT]/app/controllers/availability_controller.rb: index

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/54497/faults/39884146)

https://catalog.princeton.edu/catalog/10987449

Child of pulibrary/voyager_helpers#97

jrgriffiniii:
Resolves #515 after https://github.com/pulibrary/voyager_helpers/pull/98 was merged.  Further, as `voyager_helpers` only has one dependent, (please see https://github.com/pulibrary/voyager_helpers/network/dependents) this ensures that future improvements to `voyager_helpers` can be introduced without requiring that a new version of the Gem be released.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem 'stringex', github: "pulibrary/stringex", tag: 'vpton.2.5.2.2'
 gem 'traject', '2.3.1'
 gem 'turbolinks'
 gem 'uglifier', '>= 1.3.0'
-gem 'voyager_helpers', github: "pulibrary/voyager_helpers", tag: 'v0.7.1'
+gem 'voyager_helpers', github: "pulibrary/voyager_helpers", branch: 'master'
 gem 'whenever', "~> 0.10"
 gem 'yaml_db', '~> 0.6.0'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,8 +46,8 @@ GIT
 
 GIT
   remote: git://github.com/pulibrary/voyager_helpers.git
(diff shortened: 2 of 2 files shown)


escowles:
Do we want to just fold `voyager_helpers` into `marc_liberation`?  If it's only used in one app, it seems like it's not serving any real purpose to have it in a separate repo...

jrgriffiniii:
> Do we want to just fold voyager_helpers into marc_liberation? If it's only used in one app, it seems like it's not serving any real purpose to have it in a separate repo...

That's perfectly acceptable to me 👍. Any objections @tampakis, @kevinreiss, @mzelesky ?

kevinreiss:
@jrgriffiniii that gem is used in various scripts that have been written to facilitate bulk data work for various operations around Voyager.

escowles:
@kevinreiss Thanks for the info — it probably makes sense to keep that gem separate then, instead of having those scripts use marc_liberation instead.

mzelesky:
It was also split so that the logic of gathering data from the ILS can be separate from the processing of that data into Solr. So we'd have alma_helpers if we move to Alma, folio_helpers if we move to Folio, etc.

PR #523, opened 2018-12-26
Adds 001 authority index field

### LINKED ISSUE #520: Add 010 as a field for authority index
Probably as a string field (_s)

christinach:
Adds new dynamic field `auth_001_s` to index authority records

closes #520

### DIFF
--- a/marc_to_solr/lib/authority_traject_config.rb
+++ b/marc_to_solr/lib/authority_traject_config.rb
@@ -3,6 +3,7 @@
 require 'traject/macros/marc21_semantics'
 require 'traject/macros/marc_format_classifier'
 require 'bundler/setup'
+require 'pry-byebug'
 
 extend Traject::Macros::Marc21Semantics
 extend Traject::Macros::MarcFormats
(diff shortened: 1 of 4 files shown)


marc_to_solr/lib/authority_traject_config.rb
```diff
@@ -14,15 +14,18 @@
...
   provide "marc4j_reader.source_encoding", "UTF-8"
   provide "log.error_file", "./log/traject-error.log"
-  provide "allow_duplicate_values",  false
+  provide "allow_duplicate_values", false
   provide "solr_writer.commit_on_close", "true"
 end
 
 $LOAD_PATH.unshift(File.expand_path('../../', __FILE__)) # include marc_to_solr directory so local translation_maps can be loaded
 
-to_field 'id', extract_marc('001', first: true)
```
tampakis: can you keep the id field the same?

marc_to_solr/lib/authority_traject_config.rb
```diff
@@ -14,15 +14,18 @@
...
+to_field 'auth_001_s' do |record, accumulator|
+  value = record['001'].gsub(/\s+/, '') if recorc['001']
```
tampakis: record instead of recorc. This should pull from the 010 field instead of the 001 field

christinach: right

christinach: I went again through my notes and we said we will index the 001 with no spaces.

tampakis: oh ok

tampakis: so we do that already :)

christinach: I put back the id that pulls from the 001. We keep the new field because @mzelesky  said that it needs to be with no spaces.

PR #527, opened 2018-12-27
Adds auth_010_s field for authorities

### LINKED ISSUE #520: Add 010 as a field for authority index

christinach:
closes #520 

@tampakis I'm sorry I meant to rename the branch but because the PR was open it closed it.

### DIFF
--- a/marc_to_solr/lib/authority_traject_config.rb
+++ b/marc_to_solr/lib/authority_traject_config.rb
@@ -13,7 +13,7 @@ settings do
   provide "marc_source.type", "binary"
   provide "solr_writer.max_skipped", "50"
   provide "marc4j_reader.source_encoding", "UTF-8"
-  provide "log.error_file", "./log/traject-error.log"
+  provide "log.error_file", "./log/authority-traject-error.log"
   provide "allow_duplicate_values", false
   provide "solr_writer.commit_on_close", "true"
 end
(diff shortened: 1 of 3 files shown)


christinach:
I added one field that pulls from the 001 (auth_001_s_ and one that pulls from the 010a (auth_010_s). When I logged the authority data I didn't see any spaces in the 001 field so I removed the code to remove the spaces from the 001 field. I'm not sure which one we want to keep. In my notes I have the 001 but maybe it's the 010a. @tampakis @mzelesky ?

mzelesky:
The 001 will be used for indexing and record lookup, but the 010a is to be used for display. Does that make sense? And the reason behind that was because the 001 doesn't contain spaces. So it's correct that you don't need to add in code that will trim spaces from the 001.

marc_to_solr/lib/authority_traject_config.rb
```diff
@@ -14,15 +14,19 @@
...
+  provide "allow_duplicate_values", false
   provide "solr_writer.commit_on_close", "true"
 end
 
 $LOAD_PATH.unshift(File.expand_path('../../', __FILE__)) # include marc_to_solr directory so local translation_maps can be loaded
 
 to_field 'id', extract_marc('001', first: true)
 
-to_field 'marc_display', serialized_marc(:format => 'xml', :binary_escape => false, :allow_oversized => true)
+to_field 'auth_001_s', extract_marc('001', first: true)
```
tampakis: I think you can remove the 'auth_001_s` field since it's identical to the `id` field.

christinach:
@tampakis I removed the auth_001_s field and also added a log for the authorities to log out any record that might not have the range of fields '100..199' while having the 010 or 010a.

PR #528, opened 2019-01-24
update voyager helpers to include updated recap barcode id query

tampakis:


tampakis:
Thanks @mzelesky, thanks @christinach

PR #548, opened 2019-05-30
report rake task errors to honeybadger

tampakis:
Also see https://github.com/pulibrary/orangelight/pull/1655

### DIFF
--- a/config/honeybadger.yml
+++ b/config/honeybadger.yml
@@ -1,3 +1,3 @@
 ---
 exceptions:
-  rake_rescue: true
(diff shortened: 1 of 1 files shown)


config/honeybadger.yml
```diff
@@ -1,3 +1,3 @@
 ---
 exceptions:
-  rake_rescue: true
\ No newline at end of file
+  rescue_rake:
```
christinach: 'true'

tampakis: thanks!

PR #570, opened 2019-10-31
updates genre and siku subject fields

tampakis:
Updates the indexing strategy for non-LC subjects and genre headings.
 - Each subject/genre heading type is indexed into its own field. I _think_ we still wan to display LC and siku headings in the same subject field, but we have the flexibility of separating them with new `lc_subject_display` and `siku_subject_display` fields.
 - Added AAT genre vocab
 - Added all rbms genre terms into a single rare books genre term field.
 - Made sure that these non-lc genres and subjects are included in the subject topic and subject genre facets.
 - Removed the "form_genre_display" field because the authorities group decided to display more specific genre labels based on vocab type.

### DIFF
--- a/marc_to_solr/lib/princeton_marc.rb
+++ b/marc_to_solr/lib/princeton_marc.rb
@@ -307,6 +307,18 @@ def process_subject_topic_facet record
       subjects << subject.map { |s| Traject::Macros::Marc21.trim_punctuation(s) }
     end
   end
+  Traject::MarcExtractor.cached('650|*7|abcxz').collect_matching_lines(record) do |field, spec, extractor|
+    subject = extractor.collect_subfields(field, spec).first
+    should_include = false
+    unless subject.nil?
+      field.subfields.each do |s_field|
+        should_include = s_field.value == 'sk' if s_field.code == '2'
+        subject = subject.gsub(" #{s_field.value}", "#{SEPARATOR}#{s_field.value}") if (s_field.code == 'x' || s_field.code == 'z')
+      end
+      subject = subject.split(SEPARATOR)
+      subjects << subject.map { |s| Traject::Macros::Marc21.trim_punctuation(s) } if should_include
+    end
+  end
   subjects.flatten
 end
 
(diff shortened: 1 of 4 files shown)


marc_to_solr/lib/princeton_marc.rb
```diff
@@ -307,6 +307,18 @@ def process_subject_topic_facet record
...
+    unless subject.nil?
+      field.subfields.each do |s_field|
+        should_include = s_field.value == 'sk' if s_field.code == '2'
+        subject = subject.gsub(" #{s_field.value}", "#{SEPARATOR}#{s_field.value}") if (s_field.code == 'x' || s_field.code == 'z')
+      end
+      subject = subject.split(SEPARATOR)
+      subjects << subject.map { |s| Traject::Macros::Marc21.trim_punctuation(s) } if should_include
+    end
+  end
   subjects.flatten
```
tampakis: this is the subject_topic_facet field

marc_to_solr/spec/lib/config_spec.rb
```diff
@@ -486,4 +477,65 @@ def fixture_record(fixture_name)
         expect(subject_marc['subject_display']).to match_array(['LC Subject', 'Siku Subject'])
     end
   end
+  describe 'form_genre_display' do
```
tampakis: moved this spec from princeton_marc_spec.rb so it would pass locally.

marc_to_solr/spec/lib/princeton_marc_spec.rb
```diff
@@ -583,6 +532,10 @@
       expect(@subjects).not_to include("split genre")
     end
 
+    it 'excludes non-approved subfield $2 vocab types' do
+      expect(@subjects).not_to include("Bad subject")
+    end
+
```
escowles: Can you add some test coverage verifying that the Siku subjects show up?

tampakis: @escowles the siku subjects have been displaying in staging for a few months now: https://catalog-staging.princeton.edu/catalog/1204912

tampakis: Through this PR https://github.com/pulibrary/marc_liberation/pull/549

PR #601, opened 2020-03-16
Update to include where in ansible the ENV data can be found

carolyncole:


christinach:
@carolyncole I think it would be helpful to have it and since it is encrypted it's just fine.

PR #605, opened 2020-03-17
Update yaml_db to 0.7.0

### LINKED ISSUE #604: Update yaml_db gem to 0.7.0
yaml_db 0.6.0 doesn't work with ruby 2.6 and cannot load the backup.

christinach:
closes #604


christinach:
we need to update this gem or the back up / load task using this gem will not work with ruby 2.6 . which is now running on bibdata

Issue #595, opened 2020-03-13
Change the Subject -- implement correct alternative terms with traject technique


christinach:
## Background 

* [Technical project plan](https://docs.google.com/document/d/1pnLH4GVVAwfl-mu_Ksb8XPjoy0MaM8qlgxeGtnb6hJQ/edit)
* [marc_cleanup gem](https://github.com/pulibrary/marc_cleanup)
* [Subject headings mappings - work in progress](https://docs.google.com/spreadsheets/d/191sqH8tqpc0EFv2l1U9mh1-S1K4YFrtUQAaUlWhb4eg/edit?ts=60888289#gid=0)

{
"id":["65071"],
"subject_display":["Illegal aliens"],
"lc_subject_display":["Illegal aliens"],
"subject_facet":["Illegal aliens"],
"subject_topic_facet":["Illegal aliens"]
}

Proposed version:
{
"id":["65071"],
"subject_display":["Undocumented immigrants"],
"subject_unstem_search":["Illegal aliens","Undocumented immigrants"]
"lc_subject_display":["Undocumented immigrants"],
"subject_facet":["Illegal aliens","Undocumented immigrants"],
"subject_topic_facet":["Illegal aliens","Undocumented immigrants"]
}

**ACCEPTANCE CRITERIA**
- [x] At index time we will index synonyms, so we'll need a synonym file 
- [x] If you search for "Illegal aliens" it finds items that were cataloged as "Undocumented immigrants"
- [x] The subject term "illegal aliens" should *NOT* appear in the catalog anywhere (in facets or on show page) but it will still be indexed

mzelesky:
Updated:

{
"id":["65071"],
"subject_display":["Undocumented immigrants"],
"subject_unstem_search":["Illegal aliens","Undocumented immigrants"]
"lc_subject_display":["Undocumented immigrants"],
"subject_facet":["Undocumented immigrants"], # Subject browse list field
"subject_topic_facet":["Undocumented immigrants"] # Subject facet search results
}

tampakis:
{
"id": ["10989023"],
"subject_display": ["Undocumented immigrants—Africa—Drama","Undocumented immigrants—Europe—Drama"],
"lc_subject_display": ["Undocumented immigrants—Africa—Drama","Undocumented immigrants—Europe—Drama"],
"subject_facet": ["Undocumented immigrants—Africa—Drama","Undocumented immigrants—Europe—Drama"],
[...]

PR #606, opened 2020-03-17
adds unstem_search fields for subjects, removes combined subject display

tampakis:
The DSSG approved displaying the Siku and LC subject fields separately so the subject_display field won't need to be used.
Indexes an unstem_search field for these subject fields. Eventually the subject_unstem_search field will include original and modified subject headings (index only fields) as per the "change the subject" project. See #595 for the proposed indexing change.


tpendragon:
A corresponding Orangelight ticket will have to get added.

Issue #610, opened 2020-03-24
Allow access to patron route for developers


carolyncole:
curl https://bibdata-staging.princeton.edu/patron/cac9 available from the devs boxes

Potentially VPN access

carolyncole:
If we add VPN access this allows anyone on the university access to all the students and historical students. @kevinreiss we may not want to do this, since it allows anyone in the university access to student information that is not widely available.

hackartisan:
closing as wontfix

PR #614, opened 2020-03-25
allowing ip ranges

carolyncole:
This will allow us to add the vpn ip ranges to the white list

refs #610


hackartisan (commented):
@carolyncole Should we close this and #610 as wontfix? The concern expressed on the issue is pretty convincing.

PR #683, opened 2020-05-28
update voyager helpers

christinach:
get the circ status from the item.


christinach:
opened another pr with the release of voyager-helpers 0.7.4

Issue #705, opened 2020-07-09
Use "allow list" instead of "whitelist"


hackmastera:
1. Make/deploy code changes in marc_liberation, leaving fallback in ENV
1. Update the ENV published by princeton_ansible
1. Remove the old ENV from this codebase in a follow-up PR

PR #706, opened 2020-07-09
Use 'allow list' instead of 'whitelist'

hackmastera:
refs #705
Does not close that issue because the ENV still needs to be changed once
ansible scripts hav been updated

Note to reviewer: I'd like to test this in staging to make sure the ENV update will work, but I'm not familiar enough with the application to do so myself. Would need to pair or would be happy for someone who knows what's up to just try it out.


config/ip_allowlist.yml
```diff
@@ -5,4 +5,4 @@ test:
   addresses: "127.0.0.1"
 
 production:
-  addresses: <%= ENV["BIBDATA_IP_WHITELIST"] || "" %>
+  addresses: <%= ENV["BIBDATA_IP_ALLOWLIST"] || ENV["BIBDATA_IP_WHITELIST"] || "" %>
```
christinach, line 8: let's look at it with https://github.com/pulibrary/princeton_ansible/pull/1628/files so that we dont need the || "BIBDATA_IP_WHITELIST". Thanks.

hackmastera:
@christinach I rebased this and tried it in staging. The configuration comes through correctly. We should be good to merge this, then update ansible, then create a followup marc_liberation PR to simplify the ENV check by dropping the old key.

Issue #743, opened 2020-08-07
Create an ADR


christinach:
- [x] Indexing Decisions
- [x] Date Cataloged
- [x] Merged Holdings

christinach:
closed by #746

PR #746, opened 2020-08-10
Adds architecture decisions#743 to record indexing.

christinach:
part of #743


architecture_decisions/indexing_decisions.md
```diff
@@ -0,0 +1,18 @@
+# Record indexing decisions
+
+Date:
+
+## Status
+Accepted
+
+#### Context
```
hackmastera, line 8: Use `##`; it should be the same header level as "Status"

architecture_decisions/indexing_decisions.md
```diff
@@ -0,0 +1,18 @@
...
+**1. Date of cataloging**
+
+- The need is to identify newly available materials in a facet.
+
+**2. Merging holdings information for indexing**
+
+- To speed up indexing we need holdings information included in the dump.
+#### Decision
```
hackmastera, line 16: Use `##` here as well

christinach:
Thank you @hackmastera !

architecture_decisions/0001-document-architecture-decisions.md
```diff
@@ -0,0 +1,10 @@
+# Record architecture decisions
+Date: 2020-08-17
+## Status
+Accepted
+## Context
+We need to record the architectural decisions made on this project.
+## Decision
+We need to record the architectural decisions made on this project to orient new developers.
```
hackmastera, line 8: Decision should be like "We will..."

architecture_decisions/0003-index_merge-holdings-info.md
```diff
@@ -0,0 +1,11 @@
+# Merging holdings information for indexing
```
hackmastera, line 1: I think this file should be called "index-merged-holdings-info.md". The difference between "merge" and "merged" here is meaningful and this change will clarify. Also consistency in using dashes instead of underscores.

christinach:
@hackmastera can you check again?

Issue #757, opened 2020-08-27
Dump all Columbia ETAS Allow/Deny into a table.


tpendragon:
The table should have five fields: `id, oclc_number, bibid, status, origin`. ID is an auto-incrementing ID. Create a unique index on oclc_number. Status should be "ALLOW" or "DENY". Origin should be "CUL". Add an index to origin and status.

https://drive.google.com/file/d/1M00nUBSvnJfaVKwpMxVzcY2Z9zfaB1Kh/view?usp=sharing 

Name the table `hathi_access`

hackmastera:
Can we not require that the OCLC number be unique?

The data has multiple rows with the same oclc number. In the first case I see one row is "allow" and one is "deny". I'm not sure how this is possible, but it's there and I'd need to associate the status with the bibid in order to keep them straight. But then, at some point if we load more files we'll have more rows with the same oclc number. At that point we would need to make the bibid a more complex object as opposed to just being an id, since it would need to have the source as part of that object.

mzelesky:
My guess is that in serials, there are some volumes that are public domain and some that are in copyright. Are there any items identified as monographs that have this issue?

hackmastera:
@mzelesky The one I saw was a serial. Should we not load serials?

mzelesky:
I don't think so, but it's just another edge case. Perhaps we can surface the URL while still blocking other access, since some issues at least are available online. I'd want us to try accessing a couple records where 'allow' and 'deny' are both on the same record.

hackmastera:
The data was loaded in production.

PR #762, opened 2020-08-31
Add rake task, loader class, and models to load hathi access data into a table

hackmastera:
advances #757


lib/tasks/hathi.rake
```diff
@@ -11,7 +11,17 @@ conn = Faraday.new(url: bibdata_url) do |faraday|
...
-  
+  desc "load institution-specific hathi access file into HathiAccess objects"
+  task load_access: :environment do
+    input_location = ENV['HATHI_INPUT']
+    origin = ENV['HATHI_ORIGIN']
+    abort "usage: HATHI_INPUT=/path/to/file.tsv HATHI_ORIGIN=CUL rake hathi:load_access" unless input_location && origin
+    before_count = HathiAccess.all.count
+    Hathi::LoadAccessFile.new(input_location: input_location, origin: origin).load
+    after_count = HathiAccess.all.count
+    puts "#{after_count - before_count} records created"
```
tpendragon, line 22: This function won't be able to be run twice over the same file right? Since you call `create!` I'd expect it to make a bunch of records with duplicate information. Should we find_or_initialize or something similar?

hackmastera: Updated to use `find_or_create_by!` does that seem good? Thanks for catching this.

hackmastera:
I ran this on my dev machine and it took 1.5 hours and created 2,309,614 records.

lib/tasks/hathi.rake
tpendragon: It does, unless the ALLOW or DENY status is likely to change? I'm okay with it for now, but I wonder if in the end we'll end up with find_or_initialize_by(oclc_number:, origin:), and then update it and save.

hackmastera: Makes sense. I created https://github.com/pulibrary/marc_liberation/issues/763

hackmastera:
This took about twice as long (3 hours) with the find_and_create logic.

PR #771, opened 2020-09-03
Adds hathi access oclc lookup, updates bibid lookup

### LINKED ISSUE #760: Create a Hathi Status checkpoint /hathi/status/oclc/:oclc
@mzelesky proposed to add an oclc lookup so that: a user looking a NYPL item can view the item on Hathi trust if the oclc number has an allow status.

### LINKED ISSUE #767: hathi/access returns `id:null`
I think this is because the active record object is retrieved with only the other fields selected, but then directly rendered as json.

It probably shouldn't return `id` at all. It's a bit alarming to see the null value there.

christinach:
closes #760 
closes #767

1. `http://localhost:3000/hathi/access?bib_id=100000`
=> `[{"oclc_number":"6092569","bibid":"100000","status":"DENY","origin":"CUL"}]`
1. `http://localhost:3000/hathi/access?oclc=2243109`
=> `[{"oclc_number":"2243109","bibid":"100","status":"ALLOW","origin":"CUL"},{"oclc_number":"2243109","bibid":"100","status":"DENY","origin":"CUL"}]`

### DIFF
--- a/app/controllers/hathi_controller.rb
+++ b/app/controllers/hathi_controller.rb
@@ -1,5 +1,8 @@
 class HathiController < ApplicationController
 
+  # ToDo Remove hathi_access_bib_status when requests and Orangelight will be updated 
+  # with the new bib_id and oclc route hathi#hathi_access
+  # Also remove it from the routes.rb
   def hathi_access_bib_status
     if params[:bib_id]
       @record = HathiAccess.where(bibid: params[:bib_id])
--- a/app/controllers/hathi_controller.rb
+++ b/app/controllers/hathi_controller.rb
@@ -3,12 +3,12 @@ class HathiController < ApplicationController
   # ToDo Remove hathi_access_bib_status when requests and Orangelight will be updated 
   # with the new bib_id and oclc route hathi#hathi_access
   # Also remove it from the routes.rb
-  def hathi_access_bib_status
-    if params[:bib_id]
-      @record = HathiAccess.where(bibid: params[:bib_id])
-      render json: @record, except: [:id, :created_at, :updated_at]
-    end
-  end
+  # def hathi_access_bib_status
+#     if params[:bib_id]
+#       @record = HathiAccess.where(bibid: params[:bib_id])
+#       render json: @record, except: [:id, :created_at, :updated_at]
+#     end
+#   end
 
   def hathi_access
     return hathi_access_bib if params[:bib_id]
--- a/app/controllers/hathi_controller.rb
+++ b/app/controllers/hathi_controller.rb
@@ -1,11 +1,18 @@
 class HathiController < ApplicationController
 
-  def hathi_access
-    return hathi_access_bib_status if params[:bib_id]
-    return hathi_access_oclc_status if params[:oclc]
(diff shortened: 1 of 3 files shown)


app/controllers/hathi_controller.rb
```diff
@@ -1,9 +1,31 @@
...
+
   def hathi_access_bib_status
-    if params[:bib_id]
-      @hathi_record = HathiAccess.where(bibid: params[:bib_id]).select(:bibid,:status,:origin,:oclc_number)
-      render json: @hathi_record
+    @record = HathiAccess.where(bibid: params[:bib_id])
+    if @record.present?
+      render json: @record, except: [:id, :created_at, :updated_at]
+    else
+      status_404
```
christinach, line 13: I kept the bib_id lookup in case we decide that we want to use it in the near future.

config/routes.rb
```diff
@@ -34,7 +34,7 @@
   get '/courses', to: 'courses#index', defaults: { format: :json }
   get "/bib_ids", to: 'courses#bibs', defaults: { format: :json }
 
-  get '/hathi/access', to: 'hathi#hathi_access_bib_status', defaults: { format: :json }
+  get '/hathi/access', to: 'hathi#hathi_access', defaults: { format: :json }
```
carolyncole, line 37: Can we leave the old route?  Then this could be deployed before the code is updated.  Otherwise we have to update orangelight and requests before this can be deployed.

carolyncole (changes requested):
Looking good, but I would recommend leaving the old route available.

config/routes.rb
christinach: Yes I will add the old route as well but will also keep the new route for the oclc and the bib_id. I will add a comment to remove the old route after we update orangelight and requests with the new route.

christinach: @carolyncole I added it and removed it again. I don't think the route changes. It is calling a different method but the url is the same. I deployed and tested on bibdata-staging. catalog-staging is looking bibdata-staging. Yesterday we kept the bibdata_base change in the orangelight/requests.yml staging part. I left some commented code. If you think it is ok let me know to remove the code before merging.
The only difference is that the response from the `bib_id` lookup does not include the `id: null` and we would need to update the stubbed urls in requests and orangelight.

christinach: https://catalog-staging.princeton.edu/catalog/SCSB-3526726
https://catalog-staging.princeton.edu/catalog/SCSB-3923657
https://catalog-staging.princeton.edu/requests/SCSB-3526726?source=pulsearch
https://catalog-staging.princeton.edu/catalog/SCSB-2879206

https://bibdata-staging.princeton.edu/hathi/access?oclc=2243109
https://bibdata-staging.princeton.edu/hathi/access?bib_id=1000046

carolyncole: @christinach you are correct.  The route is not changing.  Sorry.

carolyncole (approved):

config/routes.rb
christinach: No problem @carolyncole thanks you :)

PR #778, opened 2020-09-11
Fixes to CI

### LINKED ISSUE #779: CI never retrieves cached bundle
The initial checksum of the gemlock is correct, but after bundler runs the gemlock does not contain lines for `ruby-oci8` which is excluded from install in ci. This means afterward the bundle cache is saved under a different checksum, so when it looks again on the next run it doesn't find one and does the whole bundle install again.

hackmastera:
Move the admin user configuration from env to yaml
Allow bibliographic_controller#update specs to run/pass in CI (this increases coverage a bit)
Ensure bundle cache is retrieved in ci (this decreases test runs from 12-15 min down to 2 min)

fixes #779

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,7 @@ jobs:
       - run: gem install bundler -v '~> 2.0'
       - run:
           name: Install dependencies
-          command: bundle install --path=vendor/bundle --without development production --jobs 4 --retry 3
+          command: bundle install --path=vendor/bundle --jobs 4 --retry 3
       - save_cache:
           key: marc_liberation-{{ checksum "Gemfile.lock" }}-2
           paths:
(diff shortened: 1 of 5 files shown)


spec/support/with_queue_adapter.rb
```diff
@@ -0,0 +1,18 @@
+module WithQueueAdapter
```
carolyncole, line 1: @hackmastera Is this change needed?  I would have expected the change in environment/test.rb to just do this.

carolyncole (commented):
@hackmastera Looks Great.  Just a question.

spec/support/with_queue_adapter.rb
hackmastera: It was a remnant of my roundabout path to this solution. Removed!

carolyncole (approved):

PR #792, opened 2020-09-16
Adding a Model for the CampusAccess records

### LINKED ISSUE #788: Create a Database table and Model to store who is allowed to be on campus
refs https://github.com/pulibrary/requests/issues/822

Table should have patron netid.  It is assumed if you are in the table you have the right to be on campus.  The table will be recreated each time we get a new list of names.

carolyncole:
fixes #788 
refs https://github.com/pulibrary/requests/issues/822

### DIFF
--- a/app/models/campus_access.rb
+++ b/app/models/campus_access.rb
@@ -1,0 +1,14 @@
+require 'csv'
+
+class CampusAccess < ActiveRecord::Base
+  def self.has_access?(uid)
+    where(uid: uid).count.positive?
+  end
+  def self.to_csv
+    ::CSV.generate(headers: false) do |csv|
+      all.each do |user|
+        csv << [user.uid]
+      end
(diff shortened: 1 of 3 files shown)


christinach (approved):

app/models/campus_access.rb
```diff
@@ -0,0 +1,14 @@
+require 'csv'
+
+class CampusAccess < ActiveRecord::Base
+  def self.has_access?(uid)
+    where(uid: uid).count.positive?
+  end
```
christinach, line 6: @carolyncole where will you use this?

carolyncole: @christinach the third step in https://github.com/pulibrary/requests/issues/822

christinach: thanks!

PR #793, opened 2020-09-17
Adding rake task to process daily access file

### LINKED ISSUE #789: Create a Rake task to read in the access list from a file and update the table
refs https://github.com/pulibrary/requests/issues/822

depends on #788

carolyncole:
Add method to class to load file
Adds rake task to nicely run the class method
fixes #789


app/models/campus_access.rb
```diff
@@ -1,14 +1,43 @@
...
+        worksheet = workbook[0]
+        users = []
+        header_rows
+        for row_number in (header_rows-1)..(worksheet.count-(trailer_rows+1))
+          row = worksheet[row_number]
+          course = row[0].value
+          access = row[11].value
+          users << row[2].value if (course ==  1534 || course == 1512) && access == "Y"
+        end
+        users.uniq
```
christinach, line 40: @carolyncole can you add a comment what is course 1534 and 1512?

christinach (changes requested):

app/models/campus_access.rb
carolyncole: @christinach good call.  I added a comment.

christinach (approved):

christinach:
thanks @carolyncole !

PR #796, opened 2020-09-17
Adding a route to get the list of patrons allowed on campus

### LINKED ISSUE #730: Bib format for journals needs to be updated
https://github.com/pulibrary/marc_liberation/blob/0557f74c2cec0c35f53146f975a707d8577fd54a/marc_to_solr/lib/format/bib_format.rb#L42

Bibs with the formats of 'ai' and 'as' are journals. Currently, 'ai' is listed as a book. However,  according to the definitions in the MARC format, (https://www.loc.gov/marc/bibliographic/bdleader.html), 'ai' is an integrating resource.

carolyncole:
Updated to_csv to put in emails and lowercase since that is what libcal needs
fixes #730

### DIFF
--- a/app/controllers/campus_access_controller.rb
+++ b/app/controllers/campus_access_controller.rb
@@ -1,0 +1,7 @@
+class CampusAccessController < ApplicationController
+  def index
+    respond_to do |format|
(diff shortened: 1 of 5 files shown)


kevinreiss (approved):
👍

Issue #266, opened 2017-09-29
Enhance availability api


escowles:
- [x] `status_label`
- [ ] `status_badge_color`

PR #801, opened 2020-09-23
Add status_label to availability response

hackmastera:
refs #266


christinach (approved):
I checked all the fixtures to make sure we see the right statuses. They look food. some data are customized for code purposes in case someone else checks them.

app/models/voyager_lookup.rb
```diff
@@ -97,6 +101,40 @@ def context_based_status(loc, status)
...
+      elsif status_starts?('pending order', item[:status])
+        "Pending order"
+      elsif status_starts?('order received', item[:status])
+        "Order received"
+      elsif status_starts?('on-order', item[:status])
+        "On-order"
+      else
+        item[:status]
+      end
+    end
```
christinach, line 136: This is great. You pulled the info from the availability.es6 in orangelight, correct?

hackmastera: Yes I just worked my way down the list in the orangelight javascript.

Issue #751, opened 2020-08-13
Implement alma functionality for bibliographic#bib_items


christinach:
GET /bibliographic/:bib_id/items(.:format) bibliographic#bib_items {:format=>:json}
Notes for providing `patron_group`:
/almaws/v1/bibs/{mms_id}/holdings/{holding_id}/items/{item_id}/loans gives you the patron id, use it against the patron API to get the patron group (probably the `user_group` field). Note: don’t expose the patron ID from the marc_lib endpoint.

Also create the Alma form

christinach:
https://knowledge.exlibrisgroup.com/Primo/Product_Documentation/Primo/Technical_Guide/160Mapping_to_the_Normalized_Record/030Alma_MARC_21#Availability_Statuses_from_Alma

christinach:
Items to compare
---1st ---
- https://bibdata.princeton.edu/bibliographic/731418/items
- mms_id: 99227515106421

--- 2nd ---
- https://bibdata.princeton.edu/bibliographic/31303/items
- mms_id: 99224836506421

hackartisan:
Initial implementation is done; lots of follow-up; issues exist

PR #828, opened 2020-09-30
Sketch AlmaAdapter's get_items_for_bib method

hackmastera:
Using Alma.rb

advances #751; submitting for review now to unblock other issues


README.md
```diff
@@ -27,6 +27,25 @@ Set env vars in `config/initializers/voyager_helpers.rb` and `config/initializer
...
+If you don't have an account ask our local administrator to create one for you.
+1. You will get an invitation email and be prompted to create an account
+1. Once the account is created, wait for a 2nd email to activate that account
+1. Once you've activated the account, go back to the first email and use the
+   bottom link to accept the invitation. You should now have access to our keys.
+
+Select the 'api-na' north america server
+Select the read-only API key
+Click the api endpoint you want to use
+Click 'try it out'
```
christinach, line 44: Can you add: `Select media type: application/json or application/xml`

app/adapters/alma_adapter/bib.rb
```diff
@@ -38,12 +38,20 @@ def get_bib_records(ids, conn=nil, opts={})
...
         }
         doc = res.body
       end
 
+      # @params id [string]. e.g id = "991227850000541"
+      # @return [Hash] of holdings / items data
+      def get_items_for_bib(id)
+        opts = { limit: 100, expand: "due_date_policy,due_date" }
+        Alma::BibItem.find(id, opts)
+        # todo: transform alma response to marc liberation format
```
christinach, line 52: I understand this is a to do, but do you need to add the `:api_key` we configure for bibs_read_only ?

christinach (changes requested):
Thanks @hackmastera , I added 2 comments.

app/adapters/alma_adapter/bib.rb
christinach: We have different api_keys based on the request we want to do.

christinach: Please ignore this for now. I asked @mzelesky if we can create one api_key in the sandbox. Otherwise we will update the config so that we include multiple api_keys.

hackmastera: The gem passes the api key in the header as opposed to in the url. I made the stub more explicit to show what it looks like. I also reworked the config setup a little bit. We can refactor the AlmaAdapter config in the same way (we could just get rid of that initializer and use `Rails.configuration.alma` instead) but I didn't want to introduce conflicts if you were working on any config stuff.

If we have multiple keys we may need to just use the one with the highest required level of permissions.

I'll test the rest of the integration out when I get the parsing done and can deploy it to staging.

christinach: Yes I agree; With the current keys we couldn't pick one with the highest permissions because it doesn't exist. We can discuss this tomorrow at the checkin with the rest of the group. I think for now we can use the one we have configured and is only for the bibliographic call and read only.

christinach (approved):

app/adapters/alma_adapter/bib.rb
christinach: @hackmastera the alma_keys need to be added on bibdata-staging in order to test it.

PR #853, opened 2020-10-12
Remove pagination

### LINKED ISSUE #854: Lates updates issue
Kevin Connell reported to @mzelesky that several voyager updates have not been updated in blacklight.

christinach:
closes #854
Remove pagination.

@mzelesky and I troubleshoot the issue and found out that:

- Adding pagination made the rake tasks only look at the first page of events.

- Sorting by descending made the tasks look at the oldest event on the first page instead of the newest.


hackmastera (approved):
The rake task hits the controller?

PR #881, opened 2020-10-31
Adds the ahoy_matey gem to track the api hits

### LINKED ISSUE #871: Estimate how many times we hit the alma api.
We want to see how many times the alma api is hit in marc_liberation for any given job. 
It would be good to see how many times the marc_liberation queries voyager based on the various endpoints and drive an estimate from that.

Once we get a week's worth of stats, use them in combination with the number of (anticipated) API calls per controller action to produce an estimate. Compare with the API limits.

christinach:
Adds the gem ahoy_matey to estimate/ track analytics for the bibdata api usage and for the alma api hits.

refs #871

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,6 @@
 source 'https://rubygems.org'
 
+gem 'ahoy_matey'
 gem 'alma'
 gem 'bixby', '~> 2.0'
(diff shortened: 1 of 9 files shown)


christinach:
we will not use this now

PR #945, opened 2020-12-16
Calculate cataloged date from publishing job #926

### LINKED ISSUE #926: Update catalog_date field in traject using the publishing job files
**Publishing profiles**
refs https://github.com/pulibrary/marc_liberation/issues/822#issuecomment-719602951

Bibliographic Enrichment -> Create date subfield 950b
Physical Items Enrichment -> Create date subfield 876d
Electronic Inventory Enrichment -> Activation date subfield 951w

1. Similar to the Single Record if there is a `876` field don't use the `951` fields. 
2. If there are no `876` or `951` then use the bib create date: Bibliographic Enrichment `950b`.

christinach:
closes #926


marc_to_solr/spec/lib/config_spec.rb
```diff
@@ -43,7 +43,43 @@ def fixture_record(fixture_name)
...
+  end
+  describe 'the cataloged_date from publishing job' do
+    describe "the date cataloged facets" do
+      context "When the record has 876d and 951w fields" do
+        it "will index the 876d field" do
+          expect(fixture_record('99211662100521')['951']['w']).to be_truthy
+          expect(fixture_record('99211662100521')['876']['d']).to be_truthy
+          expect(fixture_record('99211662100521')['950']['b']).to be_truthy
+          record = @indexer.map_record(fixture_record('99211662100521'))
+          expect(Time.parse(record['cataloged_tdt'].first).utc.strftime("%Y-%m-%dT%H:%M:%SZ")).to be_truthy
```
hackartisan, line 57: I'm not sure we're testing that the desired value is the one we indexed. Can we use the specific value from the fixture?

marc_to_solr/lib/traject_config.rb
```diff
@@ -285,11 +285,29 @@
...
+# catalog_date https://github.com/pulibrary/marc_liberation/issues/926
+# Bibliographic Enrichment -> Create date subfield 950b
+# Physical Items Enrichment -> Create date subfield 876d
+# Electronic Inventory Enrichment -> Activation date subfield 951w
+to_field 'cataloged_tdt' do |record, accumulator|
+  extractor_doc_id = MarcExtractor.cached("001")
+  doc_id = extractor_doc_id.extract(record).first
+  extractor_876d = MarcExtractor.cached("876d")
+  extractor_951w = MarcExtractor.cached("951w")
+  extractor_950b = MarcExtractor.cached("950b")
```
hackartisan, line 297: Maybe we should move these `MarcExtractor.cached` lines down into the conditional? It seems like we don't need need them all for each record, and it would be easier to see how they match up with the `extract` lines if they're together.

hackartisan (changes requested):
Yay, excited to see this! A couple of hopefully small comments.

marc_to_solr/spec/lib/config_spec.rb
christinach: Yes I will change this. I had some circleci failures because of the timezone this is whay I tested to be present but will look into it more.

marc_to_solr/lib/traject_config.rb
christinach: I will move them yes. I had them initially in the conditional but I moved them up when I tried to refactor it. However I decided to keep the conditional as it is because I felt it to be more descriptive, at least in my mind, and didn't put them back in the conditional.

hackartisan (approved):
👍🏻 🎉 thank you!

PR #949, opened 2020-12-21
Allowing exception file from EHS to be processed

carolyncole:
data file is put in ENV['CAMPUS_ACCESS_DIRECTORY']


christinach (commented):
Do we want to add a new menu in the navbar: "Add Exceptions" 
bibdata_exceptions

kevinreiss (approved):

PR #956, opened 2020-12-22
Updating exception_file parameter to be required

### LINKED ISSUE #953: Make the file as required so the form will not submit without a file
https://app.honeybadger.io/projects/54497/faults/72467860.

carolyncole:
fixes #953 

Also updated the UI to be just a bit nicer...

Old UI
Screen Shot 2020-12-22 at 2 33 53 PM

New UI
Screen Shot 2020-12-22 at 2 33 34 PM

### DIFF
--- a/app/views/campus_access_exceptions/new.html.erb
+++ b/app/views/campus_access_exceptions/new.html.erb
@@ -6,7 +6,10 @@
       <% end %>
     </ul>
 <% end %>
-<%= form_with url: campus_access_exceptions_path, method: :post do |form| %>
-  <%= form.file_field :exception_file, accept: "application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" %>
+<%= form_with url: campus_access_exceptions_path, method: :post, class: "form-horizontal" do |form| %>
+  <div class="form-group">
+    <%= form.label :exception_file, "Additional Exceptions:", class: "col-sm-2" %>
+    <%= form.file_field :exception_file, accept: "application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", required: true %>
+  </div>
   <%= form.submit "Add Exceptions" %>
 <% end %>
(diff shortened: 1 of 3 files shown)


app/views/campus_access_exceptions/new.html.erb
```diff
@@ -6,7 +6,10 @@
       <% end %>
     </ul>
 <% end %>
-<%= form_with url: campus_access_exceptions_path, method: :post do |form| %>
-  <%= form.file_field :exception_file, accept: "application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet" %>
+<%= form_with url: campus_access_exceptions_path, method: :post, class: "form-horizontal" do |form| %>
+  <div class="form-group">
+    <%= form.label :exception_file, "Additional Exceptions:", class: "col-sm-2" %>
+    <%= form.file_field :exception_file, accept: "application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.spreadsheetml.sheet", required: true %>
+  </div>
```
christinach, line 13: 👍🏻

spec/controllers/campus_access_exceptions_controller_spec.rb
```diff
@@ -24,6 +24,12 @@
         expect(assigns["campus_access_filename"]).to eq("#{ENV['CAMPUS_ACCESS_DIRECTORY']}/additional_campus_access.csv")
         expect(File.exist?(assigns["campus_access_filename"])).to be_truthy
       end
+
+      it "requires the file" do
+        post :create, params: { exception_file: nil }
+        expect(response).to redirect_to("http://test.host/campus_access_exceptions/new")
+        expect(flash[:alert]).to eq('param is missing or the value is empty: exception_file')
+      end
```
christinach, line 32: I can't see the flash alert on the ui.

christinach (changes requested):

spec/controllers/campus_access_exceptions_controller_spec.rb
carolyncole: @christinach you would need to send data directly to the controller to ever see this.  The form required keeps you from sending this data via the UI.
Screen Shot 2021-01-06 at 1 48 31 PM

christinach: yes I noticed the 'require=true'. but in the ui what action will you make to have the flash alert since the require: true does not allow you to submit no file?

christinach: maybe I'm confusing you and my self. I dont know how the notification as a flash alert will be triggered since I can't submit an empty file because of the client restriction. required:true. So I dont know if the flash alert is needed.

christinach (approved):

christinach (approved):

Issue #859, opened 2020-10-15
Create full dump events


hackartisan:
Once alma has created the files, it will fire a webhook. Use the AWS-lambda webhook listener developed as part of #645 to put an event on the AWS message queue, and subscribe to that queue from marc_liberation.

The local worker should then copy the files from the sftp server and create the dump and event objects.

amazon console is at princeton.edu/aws

hackmastera:
Proposal:

1. Move work from https://github.com/pulibrary/alma_experimentation/pull/8/files and  https://github.com/pulibrary/alma_experimentation/pull/33/files into `marc_liberation`.
1. Update the webhook handler. If the webhook action indicates a full dump job is complete, enqueue the event body to an AWS SQS queue.
1. Use the sidekiq cron gem https://github.com/ondrejbartas/sidekiq-cron to schedule a job to run every 15 minutes as follows: ```
Check the queue and if there's an event invoke the service that will create the Dump and Event objects and copy the files.
```

This process does not need to do any indexing; full indexing is initiated manually.

hackmastera:
SQS is free for the first million requests per month, so no issue there. https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/welcome.html#sqs-pricing

tpendragon:
Why use a cron job rather than an event listening daemon similar to Figgy's CloudFixity::Worker?

https://github.com/pulibrary/figgy/blob/40ba9ea10f26693ec887c43a4b8b1905c4b1a5bc/app/services/cloud_fixity.rb

hackmastera:
Thanks for mentioning this. It looks like that start from a rake task -- it would die if the box was rebooted? I wanted something that would be resilient to down time.

hackmastera:
Docs for the [QueuePoller](https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/SQS/QueuePoller.html)

tpendragon:
There's a systemd service file for it like we do for sidekiq workers, it spins back up if it dies.

hackmastera:
https://github.com/pulibrary/princeton_ansible/blob/main/roles/pulibrary.figgy_pubsub_worker/templates/pubsub.service

Okay great this way seems better then since it can be evented in a true pub / sub way.

hackmastera:
Here's what the webhook gives for the full publishing job:

[...]

PR #959, opened 2021-01-05
Poll SQS and create full dump events

hackartisan:
Advances #859; only need the daemon now.

Please squash merge.


app/jobs/alma_full_dump_transfer_job.rb
```diff
@@ -0,0 +1,76 @@
+require 'net/sftp'
+
+class AlmaFullDumpTransferJob < ActiveJob::Base
+  queue_as :default
+
+  REMOTE_BASE_PATH = "/home/alma".freeze
+
+  def perform(dump:, job_id:, type_constant: 'BIB_RECORDS')
```
tpendragon, line 8: Question: Why is `type_constant:` an argument? Is there a case where it might be something else for a full dump transfer job?

app/jobs/alma_full_dump_transfer_job.rb
```diff
@@ -0,0 +1,76 @@
...
+    dump_file_type = DumpFileType.find_by(constant: type_constant)
+    Net::SFTP.start(sftp_host, sftp_username, password: sftp_password) do |sftp|
```
tpendragon, line 10: suggestion: I wonder if extracting this to a little Downloader would make this easier to parse? Something like
```
Downloader.files_for(job_id: job_id).each do |file|
end
```

Not totally necessary, but might make this job easier to parse?

app/models/concerns/file_path_generator.rb
```diff
@@ -1,5 +0,0 @@
-module FilePathGenerator
-  def generate_fp
-    File.join(MARC_LIBERATION_CONFIG['data_dir'], "#{Time.now.to_i.to_s}")
```
tpendragon, line 3: :+1: Great!

tpendragon (changes requested):
Just a couple small questions. All in all, looks great! These aren't really blockers, just requesting changes so they're looked at, if you decide not to fix them just say so and I'll change to approve.

eliotjordan:
@tpendragon We implemented your suggestions. Thanks for the review.

tpendragon (approved):

PR #960, opened 2021-01-05
Fix Alma Full Reindex service.

tpendragon:
* The ALL_RECORDS is a DumpFileType, not a DumpType.
* Fixes DumpFile URL generation
* Handles binary files.
* Handles the Alma dumps being `.tar.gz`

Tested on staging with a real dump - `AlmaIndexer.new(solr_url: ENV["SOLR_URL"]).full_reindex!` works.


app/services/alma/indexer.rb
```diff
@@ -20,41 +21,44 @@ def default_url_options
     def downloaded_full_reindex_files
       @downloaded_full_reindex_files ||=
         begin
-          full_reindex_file_urls.map do |url|
+          full_reindex_file_urls.flat_map do |url|
             extension = "." + url.split("/").last.split(".", 2).last
-            temp_file = Tempfile.new(["full_reindex_file", extension])
+            temp_file = Tempfile.new(["full_reindex_file", extension], binmode: true)
             temp_file.puts Faraday.get(url).body
```
christinach, line 27: 👍🏻

christinach (approved):
@tpendragon do you want to update the rake task as well or will this be a separate issue? https://github.com/pulibrary/marc_liberation/blob/4e179bbac4f2fc66aa97a1694d0ca32ed4219f54/lib/tasks/orangeindex.rake#L162

tpendragon:
> do you want to update the rake task as well or will this be a separate issue?

I figured we'd do that separately.

Issue #933, opened 2020-12-08
Alma implementation for barcode#scsb


hackartisan:
This endpoint is used when accessioning a new item. HTC (the company that supports scsb) has to find all the metadata.

from https://developers.exlibrisgroup.com/console/?url=/wp-content/uploads/alma/openapi/bibs.json#/Catalog/get:
Note: It is possible to retrieve item information by barcode using: GET /almaws/v1/items?item_barcode={item_barcode}. Calling this shorthand URL will return an HTTP 302 redirect response leading to a URL with the structure documented here.

This endpoint needs to return the actual (XML) MARC record with holding record.

The output record needs the holding information as well as item information (which only happens for ReCAP records - see https://github.com/pulibrary/voyager_helpers/blob/master/lib/voyager_helpers/liberator.rb#L1108-L1127 for how we do it in Voyager)

Example item to pull enrichment info for:

Voyager bib ID 7991903
Item barcode 32101089814220

Enrichments:

- [x] Holding ID
- [x] Item ID
- [x] Barcode
- [x] Copy Number
- [x] RECAP Use Restriction
- [x] RECAP Group Designation
- [x] RECAP Customer Code
- [x] Item Status (Blocked by #937 ?)  - @mzelesky confirmed that although it's a required subfield the value is not actually used, so it can be anything.
- [x] enum/cron (Need to understand this - Alma has several more of these than Voyager did, example would be good.)
- [x] Holding 852 (Do we need to get the entire holding MARC?)

kevinreiss:
SCSB is not changing with respect to how this interface is used.

tpendragon:
I'm having a tough time getting `/items` to work, it's just timing out for me.

hackartisan:
Next step: contact exlibris either through basecamp or their support ticket system. start out with basecamp.

hackartisan:
the api call works over `https`

PR #969, opened 2021-01-13
SCSB Record Enrichment

tpendragon:
Work towards #933 

The only things missing now are:

1. Status (Blocked)
2. enum_chron (Blocked until I understand this)
3. Replication of this chunk of code I don't totally get: https://github.com/pulibrary/voyager_helpers/blob/e468d9ae29367d74ba7e09620238e801a7ce7bad/lib/voyager_helpers/liberator.rb#L1109-L1116

I think that last chunk deletes all existing 852s in the record, plugs in the holding record's 852, and prepends the holding record's ID as subfield 0 of the 852. I'm not sure if we want to do the same thing. If so, we'd at least have to create an 852 from scratch using the item's holding data.


app/adapters/alma_adapter/marc_response.rb
```diff
@@ -6,28 +6,52 @@ class MarcResponse
     # @param bibs [Alma::BibSet]
     def initialize(bibs:)
       @bibs = bibs
-      remove_suppressed!
```
hackartisan, line 9: I like the refactor but it does appear to change the value of `bibs` which is a public method. Is it not used anywhere else in the codebase?

hackartisan (changes requested):
This looks good to me but one question about the refactor.

app/adapters/alma_adapter/marc_response.rb
tpendragon: @hackartisan It is not, because `bibs` is an `Alma::BibSet` and everywhere we've worked on this so far we've wanted MARC records.

christinach (approved):
Looks good to me

hackartisan (approved):

PR #970, opened 2021-01-13
947 call number browse

### LINKED ISSUE #947: Index call_number_browse_s
For the call_number_display we will use 852hik.

according to @mzelesky we will **have to decide** if subfield `k` should be considered for browse still.

christinach:
closes #947 
Updates the call_number_browse_s and the call_number_display to force index subfield k at the end.


marc_to_solr/lib/traject_config.rb
```diff
@@ -1276,9 +1276,37 @@
 
 # Call number: +No call number available
 #    852 XX hik
-to_field 'call_number_display', extract_marc('852hik')
-# Alma:needs to decide if subfield k should be considered for browse list
-to_field 'call_number_browse_s', extract_marc('852hik')
```
tpendragon, line 1281: I'm pretty sure `extract_marc('852hik')` is shorthand for what you've written below, and to test I reverted this file and checked your new tests and they passed. Is it possible we just need the test changes? Unless I've missed something, which is super likely.

tpendragon (changes requested):

marc_to_solr/lib/traject_config.rb
christinach: I did check that extract_marc('852hik') will index the subfields they way they are in the marc file; in this case the subfield k comes last alphabetically, and I also thought not to force it; but @mzelesky said that it is better to force it to be safe in case a cataloger puts the k before the h and i and that Alma doesn't force an order of the fields. This is why I did it this way.

christinach: I went back in the fixture and changed the order of the h and k and my solution failed while the extract_marc('852hik') passed. I don't know why it failed but I will revert it back how it was before. :( . I will keep this fixture with the changed order.

christinach: @tpendragon disregard what I said above. When I change the order in the record of the subfield k and h my solution will pass. The extract_marc('852hik') will fail. I will push on commit with the record so that you can also check if you want.

mzelesky: From my cataloging days, it is the norm to have the subfield k before subfield h in the 852. So it is more likely than not that records will have the subfield k before subfield h.

christinach: The record from the dump has the k after the h https://raw.githubusercontent.com/pulibrary/marc_liberation/cecfa5fef933cfcbab64eaa684d5c11ed589da2d/marc_to_solr/spec/fixtures/alma/9957270023506421.mrx
which will pass using `extract_marc('852hik')`. 
If the norm is to have the record k before h like https://raw.githubusercontent.com/pulibrary/marc_liberation/6a57f0b7f9e474320306a1c092b450a684f15891/marc_to_solr/spec/fixtures/alma/9957270023506421.mrx then the index will fail if using `extract_marc('852hik')` but will pass if we force it like I do.

tpendragon: Okay. Can we condense the code to something like

```ruby
to_field 'call_number_display' do |record, accumulator|
  MarcExtractor.cached('852hik').collect_matching_lines(record) do |field, _spec, _extractor|
    accumulator << [field['h'], field['i'], [field['k']]].compact.join(" ")
  end
end
```

then?

christinach: ok

christinach: @tpendragon I updated it.

tpendragon (approved):

PR #975, opened 2021-01-14
Use single production API key and set alma region variable in .env file

### LINKED ISSUE #974: Use a single production api key and set alma region
- Use a single read-only api key authorized for all areas (Bibs, Config, Users, etc..)
- Use API from production instance
- Set the alma region with rake task

eliotjordan:
Closes #974 

See companion ansible PR: https://github.com/pulibrary/princeton_ansible/pull/2071

### DIFF
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -101,6 +101,6 @@ class AlmaAdapter
   private
 
     def apikey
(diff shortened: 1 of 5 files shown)


hackartisan (approved):
This looks good thanks for doing it.

I think this will mean everyone will need to regenerate their .env, right? I'm going to approve but not merge and if that's true maybe you can announce it to everyone when you're merging it.

PR #976, opened 2021-01-15
Merge holdings into SCSB records.

### LINKED ISSUE #933: Alma implementation for barcode#scsb

tpendragon:
Closes #933


hackartisan (approved):
🎉

PR #990, opened 2021-01-20
Use ingest scratch space for indexing.

tpendragon:


hackartisan (approved):

config/marc_liberation.yml
```diff
@@ -2,9 +2,11 @@
 # Start with a '/' for absolute paths. Otherwise we assume this is relative to
 # Rails.root
 data_dir: <%= ENV['BIBDATA_DATA_DIR'] || 'data' %>
+tmpdir: <%= Rails.env.production? ? "/mnt/isilon/ingest_scratch" : Dir.tmpdir %>
```
eliotjordan, line 5: Should this be an ENV variable setting instead of hard-coded into the YAML file?

hackartisan: Properly it should be but this is a temporary stop-gap and  we'll revisit with #991 so I think it's okay for now.

eliotjordan (approved):

config/marc_liberation.yml
christinach: I think for now we can open an issue for the var that @eliotjordan mentioned. and merge this

Issue #964, opened 2021-01-11
Populate locations table and generate translation map from the Alma API


hackartisan:
Use the Alma locations API to get an initial population of the locations table.

Run as a rake task in a development instance, and then spin up the app development. Spinning up the app should generate `locations.rb`. 

For CI:
* Copy `locations.rb` to `locations.rb.tmpl`
* Also copy `location_display.rb.tmpl`
* Look into `holding_library.rb.tmpl` and see how this needs to be set up as well.

Blocking #823

tpendragon:
`locations.rb` gets generated automatically at app spinup.

hackartisan:
Updating language on this issue. We don't actually want a seeder because we don't want it to run in CI.

PR #995, opened 2021-01-21
Populate location tables with data from Alma

eliotjordan:
Work towards #964

### DIFF
--- a/app/services/location_data_service.rb
+++ b/app/services/location_data_service.rb
@@ -1,0 +1,60 @@
+class LocationDataService
+  # Delete existing locations data repopulate tables with data from Alma
+  def self.delete_existing_and_repopulate
+    new.delete_existing_and_repopulate
+  end
+
+  def delete_existing_and_repopulate
+    Locations::HoldingLocation.delete_all
+    Locations::Library.delete_all
+    populate_libraries
+    populate_holding_locations
+  end
+
+  def populate_libraries
+    libraries.each do |library|
+      Locations::Library.create(
+        label: library.name,
+        code: library.code
+      )
+    end
+  end
+
+  def populate_holding_locations
+    libraries.each do |library|
+      library_record = Locations::Library.find_by(code: library.code)
+      holding_locations(library.code).each do |holding_location|
+        Locations::HoldingLocation.new do |location_record|
+          location_record.label = holding_location.name
+          location_record.code = "#{library.code}$#{holding_location.code}"
+          location_record.open = open_value(holding_location.type)
+          location_record.library = library_record
+          location_record.save
+        end
+      end
+    end
+  end
+
+  private
+
+    # Retrieves holding locations from Alma.
+    # @param library_code [String] e.g. "main"
+    # @return Alma::LocationSet
+    def holding_locations(library_code)
+      Alma::Location.all(library_code: library_code)
+    end
+
+    # Retrieves libraries from Alma.
+    # @return Alma::LibrarySet
+    def libraries
+      @libraries ||= Alma::Library.all
+    end
+
+    # Generates 'open' value from holding location type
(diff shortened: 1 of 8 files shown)


app/services/location_data_service.rb
```diff
@@ -0,0 +1,60 @@
+class LocationDataService
+  # Update location tables with data from Alma
+  def self.populate
```
hackartisan, line 3: Maybe call it `repopulate` to reflect the deletion that will happen? I'm a little anxious about rolling it all into one and then allowing the rake task to be invoked in production. But we'll have to run it in production at least once, right? Should we consider putting more protections on this?

I know this isn't ready for review yet and this isn't a reivew; I just was thinking about this since I saw it.

eliotjordan: Since this is not a common task (we'll run it once before going into production, and rarely or never at other times), maybe the answer here is not to do this in a rake task. We can open a console and run the service there when we need it.

eliotjordan:
@hackartisan I renamed the `populate` method to be more explicit and removed the rake task so it can only be run in the console.

spec/services/location_data_service_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    it "deletes exsiting data and populates library and location data from Alma" do
+      LocationDataService.delete_existing_and_repopulate
+      library_record = Locations::Library.find_by(code: 'main')
+      location_record1 = Locations::HoldingLocation.find_by(code: 'main$stacks')
+      location_record2 = Locations::HoldingLocation.find_by(code: 'arch$reserves')
+
+      expect(Locations::Library.count).to eq 2
+      expect(Locations::HoldingLocation.count).to eq 13
+      expect(library_record.label).to eq 'Main Library'
+      expect(location_record1.label).to eq 'Main Library Stacks'
```
hackartisan, line 66: I just looked over the notes from the meeting; we talked about doing "Main Library - Stacks" @mzelesky can you confirm?

hackartisan: I can't figure out where this comes from, though. Does it just come from Alma directly?

hackartisan: Okay. In the notes we said the location label should be `"#{library.name} - #{location.name}"` -- but looking at the fixture data it seems like the location names sometimes include the library name? But sometimes they dont, e.g. "Microform Room". So I think we do need to use both, and potentially change some of the location names in alma to not also include the library names.

hackartisan (approved):
This looks great and it progresses the issue, so I'm going to merge it and we can continue to work on display issues in future PRs.

hackartisan (changes requested):
Oh actually I'm going to block; I'm not sure the upstream branch is correct since that PR isn't yet merged.

spec/services/location_data_service_spec.rb
hackartisan: Thanks @mzelesky

eliotjordan: The fixtures in this PR were drawn from the Alma API examples and don't reflect PUL names. I didn't concatenate the library names because as Mark suggested, they need to be cleaned up. E.g. the firestone stacks location would be rendered as. "Firestone Library - Firestone Library",

hackartisan: I think showing the data in the form it will actually display could motivator for those tasked with cleaning it, though.

christinach (approved):

eliotjordan:
@christinach @hackartisan The alma_rb libraries and locations PR was merged so I updated the Gemfile with the correct upstream. This is ready for re-review.

christinach (approved):

hackartisan (approved):

PR #997, opened 2021-01-21
Add support for bound-withs in SCSB barcode lookup.

### LINKED ISSUE #996: Return bound-withs from barcode#scsb with desired format
In voyager we get several records returned with the same 852 and 876. This is the desired output.
example:
https://bibdata.princeton.edu/barcode/32101066958685/scsb

In alma we just get the host bibliographic record. 

Our controller needs to use the 774 fields to get the 3 other records, add in the 852 and 876 data, and return all 3 records.

tpendragon:
Closes #996


christinach (approved):
thanks!

Issue #993, opened 2021-01-21
Fatal traject error in make_time


hackartisan:
The full index errored before completing. Here is some output:

```
2021-01-20T23:35:46+00:00 ERROR 001 9958367833506421 - Missing holdings
2021-01-20T23:35:46+00:00 ERROR 001 9938582623506421 - Invalid Location Code: annex
2021-01-20T23:35:46+00:00 ERROR 001 9958367833506421 - Invalid Location Code: firestone
2021-01-20T23:35:46+00:00 ERROR 001 9924621233506421 - Missing holdings
2021-01-20T23:35:46+00:00 ERROR Unexpected error on record id `9938589053506421/9938589053506421` at file position 7349
    while executing (to_field cataloged_tdt at marc_to_solr/lib/traject_config.rb:292)
    Exception: ArgumentError: no time information in "W45 2002"
    /usr/lib/ruby/2.6.0/time.rb:194:in `make_time'

2021-01-20T23:35:46+00:00 ERROR 001 9924621233506421 - Invalid Location Code: rare
2021-01-20T23:35:46+00:00 FATAL Traject::CommandLine: Unexpected exception, terminating execution: #<ArgumentError: no time information in "W45 2002">
2021-01-20T23:35:46+00:00 ERROR 001 9938581813506421 - Missing holdings
/usr/lib/ruby/2.6.0/time.rb:194:in `make_time': no time information in "W45 2002" (ArgumentError)
        from /usr/lib/ruby/2.6.0/time.rb:375:in `parse'
        from marc_to_solr/lib/traject_config.rb:303:in `block (2 levels) in load_config_file'
        from /opt/marc_liberation/shared/bundle/ruby/2.6.0/gems/traject-2.3.1/lib/traject/indexer/step.rb:149:in `execute'
        from /opt/marc_liberation/shared/bundle/ruby/2.6.0/gems/traject-2.3.1/lib/traject/indexer.rb:346:in `block (2 levels) in map_to_context!'
        from /opt/marc_liberation/shared/bundle/ruby/2.6.0/gems/traject-2.3.1/lib/traject/indexer.rb:400:in `log_mapping_errors'
        from /opt/marc_liberation/shared/bundle/ruby/2.6.0/gems/traject-2.3.1/lib/traject/indexer.rb:345:in `block in map_to_context!'
[...]

PR #1002, opened 2021-01-25
Rescue and log a Time.parse error

hackartisan:
refs #993

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -300,7 +300,11 @@ to_field 'cataloged_tdt' do |record, accumulator|
                      else
                        MarcExtractor.cached("950b").extract(record).first
                      end
-    accumulator[0] = Time.parse(cataloged_date).utc.strftime("%Y-%m-%dT%H:%M:%SZ") unless cataloged_date.nil?
+    begin
+      accumulator[0] = Time.parse(cataloged_date).utc.strftime("%Y-%m-%dT%H:%M:%SZ") unless cataloged_date.nil?
+    rescue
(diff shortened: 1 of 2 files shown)


christinach (approved):

christinach (dismissed):

marc_to_solr/lib/traject_config.rb
```diff
@@ -300,7 +300,11 @@
                      else
                        MarcExtractor.cached("950b").extract(record).first
                      end
-    accumulator[0] = Time.parse(cataloged_date).utc.strftime("%Y-%m-%dT%H:%M:%SZ") unless cataloged_date.nil?
+    begin
+      accumulator[0] = Time.parse(cataloged_date).utc.strftime("%Y-%m-%dT%H:%M:%SZ") unless cataloged_date.nil?
+    rescue
+      logger.error "#{record['001']} - error parsing cataloged date #{cataloged_date}"
+    end
```
christinach, line 307: @hackartisan do we want to add 'Failed to index ....and the error'. So that we know that it failed to index.

christinach: The rest of the errors we log are to inform us about cataloging info.

christinach (changes requested):
we use the traject log to log cataloging info. If we rescue a record we should indicate in the log that it Failed to index.

marc_to_solr/lib/traject_config.rb
hackartisan: I don't think it failed to index the record. It just failed to get this value. Is that what you mean?

christinach: oh ok, I thought it failed to index the record.

christinach (approved):

Issue #824, opened 2020-09-29
Process suppressed and deleted bibs


christinach:
For the indexing, we will check the leader. We turned on a flag in the publishing job to mark suppressed records as 'deleted' in the leader. The sixth character of the leader would be a 'd'. 

Because of this setting, suppressed and deleted bibs will look the same.

We should skip the record if it's not in solr, and delete it if it is.

Note: Supressed records also have 950 $a value "true", **but we won't reference it here** (configuration is: 950 $a Suppress from publishing Bibliographic Enrichment)

Example suppressed record: 9938589053506421

hackartisan:
How to tell whether a record is suppressed is unclear, as is what to do with a suppressed record. blocked pending specification.

mzelesky:
We can turn on a flag in the publishing job to mark suppressed records as 'deleted' in the leader, which is less prone to cataloger error than the 950 field.

The sixth character of the leader would be a 'd'.

christinach:
@mzelesky yes we can do that. I think it's a better option.

hackartisan:
It looks like traject added a `delete` function in 3.1.0: https://github.com/traject/traject/pull/214/commits/c731d09c6cb15fbe572e4a294a7a127b525b7277

we appear to be running 2.3.1. Here's the [changelog](https://github.com/traject/traject/blob/master/CHANGES.md).

PR #1003, opened 2021-01-26
Skip suppressed records during indexing

hackartisan:
refs #824


hackartisan:
We're going to put a flag in the leader instead of checking 950a

PR #1004, opened 2021-01-26
Traject contained_in_s

### LINKED ISSUE #922: Update contained_in_s field in traject for Alma
https://github.com/pulibrary/marc_liberation/blob/a691bf10e8f475172a586921d003f8f280329589/marc_to_solr/lib/traject_config.rb#L536-L541

christinach:
closes #922


christinach:
The fixture that is used in this pr has two 773s with the same host id

hackartisan (approved):

PR #1006, opened 2021-01-26
Update locations and library templates with Alma data

### LINKED ISSUE #998: Update locations, location_display, and holding_locations templates with Alma data
Part of #964

eliotjordan:
Closes #998


marc_to_solr/spec/lib/config_spec.rb
```diff
@@ -368,9 +368,9 @@ def fixture_alma_record(fixture_name)
   end
   describe 'excluding locations from library facet' do
     it 'when location is online' do
-      expect(@online['location_code_s']).to include 'elf1'
-      expect(@online['location_display']).to include 'Online - *ONLINE*'
-      expect(@online['location']).to be_nil
+      expect(@online['location_code_s']).to include 'online$elf1'
+      expect(@online['location_display']).to include 'Electronic Access - elf1 Internet Resources'
+      expect(@online['location']).to eq ['Electronic Access']
     end
```
christinach, line 374: As far as I understand we shouldn't have elfs at all. Is this an attempt to pass the voyager specs?

marc_to_solr/spec/fixtures/sample13.mrx
```diff
@@ -152,7 +152,7 @@
       <marc:subfield code="a">Collection Archives (Institut national de la communication audiovisuelle (France))</marc:subfield>
     </marc:datafield>
     <marc:datafield tag="852" ind1="8" ind2=" ">
-      <marc:subfield code="b">rcpqk</marc:subfield>
+      <marc:subfield code="b">recap$pq</marc:subfield>
       <marc:subfield code="h">LS- 11479</marc:subfield>
```
christinach, line 156: Did we decide to go with the `$` or the `-` between location and code? oh ok I see a follow up issue on that https://github.com/pulibrary/marc_liberation/issues/1012. @eliotjordan is it better to update it now with the `-` and regenerate the tables? @hackartisan what do you think?

hackartisan: The codes will have the `$`. The issue is to add the `-` into the labels. I don't think it's important to do it before merging this based on discussions I've had with @eliotjordan. I do think it will be good to do it on the soon side.

marc_to_solr/spec/lib/config_spec.rb
christinach: ok, I understand that this is only for the voyager specs. We will update it and remove the elfs when we clean the config_spec from the voyager fixtures.

christinach (approved):

PR #1020, opened 2021-01-29
Various fixes and refactors

eliotjordan:
- Fixes issue where running Lando generates an incorrect holding_library.rb file.
- Allows all specs to be run with test runners like vim-test.
- Moves LocationProcessorService and IndexFunctions out of marc_to_solr and into the main app where they are used exclusively.
- Refactors LocationProcessorService from a module of class methods into a real class called LocationMapsGeneratorService.
- Extracts indexer build method into a new IndexerService class.
- Adds tests for IndexerService and LocationMapsGeneratorService
- Moves all specs from marc_to_solr into the main app specs.
- Instructs rubocop to exclude the generated translation map files.


config/initializers/traject_initializer.rb
```diff
@@ -1,4 +1,7 @@
-require File.join(Rails.root, 'marc_to_solr', 'lib', 'location_processor_service')
+if Rails.env.test? || Rails.env.development?
+  LocationMapsGeneratorService.generate_from_templates
```
hackartisan, line 2: Does this mean we should remove the readme instructions to copy the files?

hackartisan (changes requested):
This is awesome. I think we just need to remove the copy instructions from the readme.

config/initializers/traject_initializer.rb
eliotjordan: I didn't see anything in the README about that, but if there is a mention somewhere of copying the files, we should remove it.

hackartisan: @eliotjordan it was added in https://github.com/pulibrary/marc_liberation/pull/1013. maybe you need to rebase?

eliotjordan: I see it now. I was searching for different terms. Pushed a commit to remove that section.

hackartisan (approved):

christinach (approved):

PR #1021, opened 2021-01-29
schedule job to delete old events/files

### LINKED ISSUE #1016: Schedule a job via whenever to delete dump files
Full dump events and all associated files should be deleted if they're older than 6 months.

Incremental dump events and all associated files should be deleted if they're older than 2 months.

hackartisan:
closes #1016

### DIFF
--- a/app/jobs/delete_events_job.rb
+++ b/app/jobs/delete_events_job.rb
@@ -1,0 +1,12 @@
+class DeleteEventsJob < ActiveJob::Base
+  def perform(dump_type:, older_than:)
+    start = Event.arel_table[:start]
+    ids = Event
+          .joins(dump: :dump_type)
+          .where('dump_types.constant': dump_type)
+          .where(start.lt(older_than))
+          .map(&:id)
+
+    Event.destroy(ids)
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/tasks/dev.rake
```diff
@@ -1,3 +1,30 @@
+namespace :marc_liberation do
+  namespace :server do
```
eliotjordan, line 2: Definitely seems like it belongs in `dev.rake`. I created a follow-up ticket: https://github.com/pulibrary/marc_liberation/issues/1025

app/jobs/delete_events_job.rb
```diff
@@ -0,0 +1,12 @@
+class DeleteEventsJob < ActiveJob::Base
+  def perform(dump_type:, older_than:)
+    start = Event.arel_table[:start]
+    ids = Event
+          .joins(dump: :dump_type)
+          .where('dump_types.constant': dump_type)
+          .where(start.lt(older_than))
```
eliotjordan, line 7: Took me a minute to grok that older date times are less than (<) newer date times. Seconds since the Unix Epoch rather that seconds since the current time. TIL.

eliotjordan (approved):

christinach (approved):

PR #1024, opened 2021-01-29
Adds rake task to index a single MARC XML file

hectorcorrea:
First pass at adding a rake task to allow to index a single MARC XML file via command like this

```
SET_URL=http://solr/core/path bundle exec rake liberate:index_file[/path/to/marc-file.xml]
```


lib/tasks/orangeindex.rake
```diff
@@ -167,6 +167,16 @@ namespace :liberate do
     solr.commit
   end
 
+  desc "Index a single MARC XML file against SET_URL"
+  task :index_file, [:file_name] => :environment do |_t, args|
+    solr_url = ENV['SET_URL'] || default_solr_url
+    file_name = args[:file_name]
```
hackartisan, line 173: pul codebases tend to use environment variables instead of arguments with rake tasks. I think we should use `ENV['FILE']` or similar.

lib/tasks/orangeindex.rake
```diff
@@ -167,6 +167,16 @@ namespace :liberate do
...
+    abort "MARC XML file name must be indicated" if file_name == nil
```
hackartisan, line 174: `unless file_name`?

hackartisan (changes requested):
This looks great I just left a couple of style comments. Good idea to add this task.

lib/tasks/orangeindex.rake
hectorcorrea: Thanks for the pointer. Done.

lib/tasks/orangeindex.rake
```diff
@@ -168,10 +168,10 @@ namespace :liberate do
   end
 
   desc "Index a single MARC XML file against SET_URL"
-  task :index_file, [:file_name] => :environment do |_t, args|
+  task index_file: :environment do |_t, args|
```
hackartisan, line 171: I think we don't need the block argument `|_t, args|` anymore now

hackartisan (changes requested):

lib/tasks/orangeindex.rake
hectorcorrea: Fixed.

hackartisan (approved):

PR #1037, opened 2021-02-02
Update the readme file with a new section on how to export a set of records from the exlibris production box

christinach:
closes https://github.com/pulibrary/marc_liberation/issues/1036


hackartisan (approved):
Thank you for adding this! I was also thinking we should get it down.

PR #1051, opened 2021-02-04
Added inline comments about dump types to yml file

### LINKED ISSUE #1038: Document dumptypes and dumpfile types
For each type, where does that data come from? How is it used in indexing? which cron job does it correspond to? How do we expect it to change once we're on alma?

Informs #927 and #924

hackartisan:
refs #1038

@christinach I put in these comments in the context of alma; several of these won't be used any more and there will be less complexity in the interactions of these dumps. I think this might be sufficient to close #1038; what do you think?

(note it helps to view the file with the split diff view)

### DIFF
--- a/config/marc_liberation.yml
+++ b/config/marc_liberation.yml
@@ -7,31 +7,31 @@ recap_dump_records_per_file: 5000
 test_records_per_file: 25
 
 dump_types:
-- label: 'Changed Records'
-  constant: 'CHANGED_RECORDS'
-- label: 'All Bib IDs'
-  constant: 'BIB_IDS'
-- label: 'All Bib IDs with Holdings'
-  constant: 'MERGED_IDS'
-- label: 'All Records'
+- label: 'All Records' # used for full dumps
   constant: 'ALL_RECORDS'
-- label: 'Updated Princeton ReCAP Records'
+- label: 'Changed Records' # used for incremental dumps
+  constant: 'CHANGED_RECORDS'
+- label: 'Updated Princeton ReCAP Records' # used for Princeton records we send to SCSB
   constant: 'PRINCETON_RECAP'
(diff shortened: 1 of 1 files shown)


eliotjordan (approved):

config/marc_liberation.yml
```diff
@@ -7,31 +7,31 @@ recap_dump_records_per_file: 5000
...
-- label: 'All Bib IDs with Holdings'
-  constant: 'MERGED_IDS'
-- label: 'All Records'
+- label: 'All Records' # used for full dumps
   constant: 'ALL_RECORDS'
-- label: 'Updated Princeton ReCAP Records'
+- label: 'Changed Records' # used for incremental dumps
+  constant: 'CHANGED_RECORDS'
+- label: 'Updated Princeton ReCAP Records' # used for records we send to SCSB
   constant: 'PRINCETON_RECAP'
```
christinach, line 15: Can we be more specific by saying: used for Princeton records we send to SCSB

config/marc_liberation.yml
```diff
@@ -7,31 +7,31 @@ recap_dump_records_per_file: 5000
...
-- label: 'Updated Partner ReCAP Records'
+- label: 'Updated Partner ReCAP Records' # used for records we get from SCSB partners
   constant: 'PARTNER_RECAP'
```
christinach, line 17: Can we also update to: used for records we get from SCSB partners (NYPL and Columbia)

christinach (changes requested):
@hackartisan thank you for adding the comments. Yes we can close https://github.com/pulibrary/marc_liberation/issues/1038 after merging this PR.

hackartisan:
@christinach I made those changes. thanks!

christinach (approved):

PR #1052, opened 2021-02-04
Add holding notes to bib items document 

### LINKED ISSUE #856: Add notes in the get_items_for_bib method
child of #751 

This method is in `app/adapters/alma_adapter.rb` and used by the bibliographic controller.

Example call to bibliographic#bib_items:
https://bibdata.princeton.edu/bibliographic/426420/items

eliotjordan:
Closes #856


app/adapters/alma_adapter.rb
```diff
@@ -103,4 +107,59 @@ def perm_location(bib_item)
...
+    # @param [Hash<String, Array>] hash with holding id as key and notes as values
+    def holding_notes_from_bib_record(mms_id)
+      record = get_bib_record(mms_id)
+      return {} unless record&.try(:fields)
+      notes_by_holding = {}
+      ava_subfields = record.fields.select { |f| f.tag == "AVA" }.select { |s| s.codes.include? "t" }
+      ava_subfields.each do |subfield|
+        holding_id = subfield["8"]
+        note = subfield["t"]
+        if notes_by_holding.key?(holding_id)
```
hackartisan, line 140: Could this conditional be collapsed by using fetch? `notes_by_holding[holding_id] = notes_by_holding.fetch(holding_id, []) << note`

app/adapters/alma_adapter.rb
```diff
@@ -103,4 +107,59 @@ def perm_location(bib_item)
...
+    def bib_item_holding_notes(mms_id, all_holding_ids)
+      notes_from_bib_record = holding_notes_from_bib_record(mms_id)
+
+      # Get list of holding ids not included in bib record AVA fields
+      missing_holding_ids = all_holding_ids - notes_from_bib_record.keys
+
+      # Bib record AVA fields do not include the holding ID for temporary holding locations.
+      # We have to fetch notes for these holdings individually using their 866
+      # value.
+      notes_from_holdings = holding_notes_from_holding_records(mms_id, missing_holding_ids)
```
hackartisan, line 123: So if missing_holding_ids is [] there aren't any extra api calls. I remember in the call we talked about checking whether there were temporary locations. Did you discover that wasn't necessary because the code already has to get the holding ids first? That's what it looks like, just wanting to confirm that we're sure we're not making more api calls than we have to!

spec/adapters/alma_adapter_spec.rb
```diff
@@ -140,13 +140,86 @@
     end
   end
 
-  describe ".get_items_for_bib" do
-    it "returns a Hash" do
-      expect(adapter.get_items_for_bib(unsuppressed)).to be_a Hash
+  describe "#get_items_for_bib" do
+    context "when a record has holdings with notes" do
```
hackartisan, line 144: Does the fixture you used have the temporary locations scenario we looked at? Could you add that to the context here?

hackartisan (changes requested):
This looks good. I am wondering if we want to expand the tests a little. Assuming the fixture used here Is the case that involves the temporary location scenario, should we add a test for the case that does have notes but doesn't have a temporary location? It would be nice to label (even just with a comment) one of the pre-existing tests as not having any notes, so we know that case has test coverage.

app/adapters/alma_adapter.rb
eliotjordan: That's correct. The code here tries to get all of the notes it can from the AVA fields on the bib-record. Temporary holding locations will have an AVA entry, but no note and no subfield 8, so those are skipped. The full list of holding ids is already known (from the items API call), the holdings with AVA fields are subtracted from this list, and individual calls to each holding to get the note from their 866 are done only on the remaining ids. It should be the minimum number of API calls needed to get all of the notes.

spec/adapters/alma_adapter_spec.rb
eliotjordan: It does. I'll add it.

app/adapters/alma_adapter.rb
eliotjordan: That's a nice refactor. I'll give it a try.

eliotjordan:
@hackartisan Update the PR with your suggestions and added a new spec for a record with only a permanent holding location note.

hackartisan (approved):
Awesome. Thank you!

Issue #1019, opened 2021-01-29
Solr in Lando


hectorcorrea:
The Solr admin screen does not show the list of fields in the Solr dev and Solr test in Lando. See screenshot below. 

~~Is this by design or an oversight?~~ [UPDATE: this drop down list is in fact enabled, it just took a long time to load the very first time. However, there is indeed an error in the solrconfig.xml that causes Solr to log errors on initialization, these errors are documented below on this ticket]

![solr_lando](https://user-images.githubusercontent.com/568286/106296499-c0051980-621f-11eb-9280-51a21788333b.png)

tpendragon:
This isn't on purpose. I've seen this happen when there's a broken field or something.

hectorcorrea:
A broken field in Solr, like a bad solrconfig.xml or schema.xml ?

hectorcorrea:
It seems that I can indeed click on the fields on the Schema and browse through the fields. I thought I couldn't (?)

But I also noticed that the Solr log is showing up and error on some of the fields defined in `solrconfig.xml`. The error says:

```
2/1/2021, 2:35:17 PM | ERROR true | x:marc-liberation-core-development | RequestHandlerBase | 
org.apache.solr.common.SolrException:  org.apache.solr.search.SyntaxError: 
Cannot parse  'cataloged_tdt:[NOW/DAY-7DAYS NOW/DAY+1DAY]': Encountered "  <RANGE_GOOP> "NOW/DAY+1DAY "" at line 1,&#8203; column 29.

org.apache.solr.common.SolrException: org.apache.solr.search.SyntaxError: 
Cannot parse 'cataloged_tdt:[NOW/DAY-7DAYS NOW/DAY+1DAY]': Encountered " <RANGE_GOOP> "NOW/DAY+1DAY "" at line 1, column 29. 

Was expecting:     "TO" ...      	at 

org.apache.solr.handler.component.QueryComponent.prepare(QueryComponent.java:218) 	at 
...
java.base/java.lang.Thread.run(Unknown Source) Caused by: org.apache.solr.search.SyntaxError: 
Cannot parse 'cataloged_tdt:[NOW/DAY-7DAYS NOW/DAY+1DAY]': Encountered " <RANGE_GOOP> "NOW/DAY+1DAY "" at line 1, column 29.
```

These fields are defined as follow:

```
[...]

PR #1056, opened 2021-02-08
Lando solrconfig_xml 

hectorcorrea:
Allows the Rake task to update Solr configs to indicate a path to use (e.g. `catalog-production` vs `catalog-staging`) to determine what Solr configuration to download from GitHub. If no parameter is given it uses `catalog-production` as the code used to do before.

Sample of usage: 

```
bundle exec rake server:solr:configs:update[,catalog-staging]
```

The reason for this PR is to be able to pull down config files from environments other than production to the local machine and test them, for example to address issue #1019

We should merge PR https://github.com/pulibrary/pul_solr/pull/253 before we can test this one.

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -37,7 +37,7 @@ namespace :server do
     end
     namespace :configs do
       desc "Updates solr config files from github"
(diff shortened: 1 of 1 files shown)


hackartisan (approved):
👍🏻

PR #1063, opened 2021-02-09
Strip non-numeric tags from SCSB barcode response.

### LINKED ISSUE #1062: Remove non-numeric fields from barcode#scsb response
All non-numeric fields like AVA / AVE, etc need to go away because their parser is strict. 

High priority

Implementation:
* We aren't actually using anything from these fields so if it's easiest to just request the record without them that should be fine.

tpendragon:
Closes #1062

## Implementation Note

I considered just not asking for E_AVAIL/P_AVAIL, etc, so these fields wouldn't be in the requested resource in the first place, but this seemed so easy to parse and write that seemed silly, when otherwise we'd have to manage multiple layers of fixtures and deciding whether to ask for those fields or not.

### DIFF
--- a/spec/controllers/barcode_controller_spec.rb
+++ b/spec/controllers/barcode_controller_spec.rb
@@ -36,6 +36,9 @@ RSpec.describe BarcodeController, type: :controller do
         expect(record["856"].as_json).to eq voyager_comparison["856"].as_json
         # Ensure 959 is correct (empty)
         expect(record["959"]).to be_nil
+        # Ensure there are no non-numeric fields
+        # ReCAP's parser can't handle them.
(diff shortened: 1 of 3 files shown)


spec/controllers/barcode_controller_spec.rb
```diff
@@ -36,6 +36,9 @@
         expect(record["856"].as_json).to eq voyager_comparison["856"].as_json
         # Ensure 959 is correct (empty)
         expect(record["959"]).to be_nil
+        # Ensure there are no non-numeric fields
+        # ReCAP's parser can't handle them.
+        expect(record.fields.map { |x| format("%03d", x.tag.to_i) }).to contain_exactly(*record.fields.map(&:tag))
```
hackartisan, line 41: String.to_i returns 0 when the string doesn't start with a number.

hackartisan (approved):
looks good. Thank you!

PR #1066, opened 2021-02-10
Use the Library/Location combo location for bib_items.

### LINKED ISSUE #1061: bibliograhic#bib_items endpoint should use library$holding location code
The [firestone_locator uses the location key](https://github.com/pulibrary/firestone_locator/blob/00ad7b643305ea055bee34cd717446aa950bb399/includes/bibdata_info.php#L60-L78) to drive some of its logic.

Since holding locations alone aren't unique, the endpoint should use the full library$holding locations as keys. see, e.g. https://bibdata-alma-staging.princeton.edu/bibliographic/9934883173506421/items with location "stacks" -- this is not a unique holding location.

Note also that code in the firestone locator looks for location code `f` and will need to be updated for migration.

related to pulibrary/orangelight#2348

tpendragon:
Closes #1061


hackartisan (approved):
👍🏻

PR #1068, opened 2021-02-10
Refactor Alma BibItems response.

tpendragon:
Moves around the responsibilities to independent objects and avoids
doing procedural function calls.

### DIFF
--- a/app/adapters/alma_adapter/bib_item_set.rb
+++ b/app/adapters/alma_adapter/bib_item_set.rb
@@ -1,0 +1,77 @@
+class AlmaAdapter
+  class BibItemSet
+    include Enumerable
+
+    attr_reader :items, :adapter
+    delegate :get_bib_record, to: :adapter
+    def initialize(items:, adapter:)
+      @items = items
+      @adapter = adapter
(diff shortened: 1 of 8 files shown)


app/adapters/alma_adapter/marc_record.rb
```diff
@@ -0,0 +1,69 @@
+class AlmaAdapter
+  class MarcRecord < SimpleDelegator
```
hackartisan, line 2: Add a class level comment just describing what this class is / does? Seems like a wrapper class.

app/adapters/alma_adapter/bib_item_set.rb
```diff
@@ -0,0 +1,73 @@
+class AlmaAdapter
+  class BibItemSet
```
hackartisan, line 2: Add a little class-level documentation?

hackartisan (changes requested):
It looks very good to me; I just added some requests for code docs

hackartisan (approved):

eliotjordan (approved):

PR #1070, opened 2021-02-10
Create incremental dump events from the sqs poller

### LINKED ISSUE #927: Create an incremental dump event
Requires #1000

Like #859 but for the incremental index

Work performed for #859 should get the alma event through the aws listener and onto the SQS poll, where it will be picked up by the poller daemon (see [diagram](https://lib-confluence.princeton.edu/pages/viewpage.action?spaceKey=ALMA&title=Systems+Documentation)). Detect that it's the incremental event and create an appropriate dump, kicking off a transfer job to pull the files.

hackartisan:
closes #927


app/services/aws_sqs_poller.rb
```diff
@@ -7,30 +7,38 @@ def self.poll
...
+      AlmaDumpTransferJob.perform_later(
         dump: dump,
         job_id: message_body["job_instance"]["id"]
       )
     end
   end
 end
 
-class AlmaFullDumpFactory
+def jobs_map
```
tpendragon, line 26: Is this out of the class definition on purpose?

app/services/aws_sqs_poller.rb
```diff
@@ -7,30 +7,38 @@ def self.poll
...
+class AlmaDumpFactory
   attr_reader :message
   def initialize(message)
     @message = message
   end
 
-  def self.full_bib_dump(message)
-    new(message).full_bib_dump
+  def self.bib_dump(message, constant)
+    new(message).bib_dump(constant)
```
tpendragon, line 37: Question: Does one message create multiple kinds of bib dump constants? Should constant be in the initializer rather than an argument to `bib_dump`? Or even better be able to be derived from the message given, so we can remove passing that argument around?

tpendragon (changes requested):

app/services/aws_sqs_poller.rb
hackartisan: oops thank you for catching this

hackartisan:
@tpendragon refactored. Take another look?

tpendragon (approved):

PR #1076, opened 2021-02-11
Revert holding notes work

### LINKED ISSUE #1064: revert holdings_note work
Adding this field greatly increased the number of Alma API calls generated by hitting the bibliographic#bib_items endpoint. Since in #1055 we determined it won't be used by any existing clients, we should revert that work.

field was added in https://github.com/pulibrary/marc_liberation/pull/1052

eliotjordan:
Closes #1064


spec/adapters/alma_adapter_spec.rb
```diff
@@ -233,18 +155,6 @@
             }
           )
           .to_return(status: 200, body: "{}", headers: { "content-Type" => "application/json" })
-
```
hackartisan, line 236: Can the stub above also be removed? It was added in https://github.com/pulibrary/marc_liberation/pull/1052

spec/adapters/alma_adapter_spec.rb
```diff
@@ -306,38 +216,13 @@
...
```
hackartisan, line 309: We should remove the "no holding notes" part from the spec title here. And can the first stub also be removed? It was added in https://github.com/pulibrary/marc_liberation/pull/1052

spec/adapters/alma_adapter_spec.rb
```diff
@@ -383,42 +268,6 @@
...
-        stub_request(:get, "https://api-na.hosted.exlibrisgroup.com/almaws/v1/bibs/#{unsuppressed_loc_with_two_holdings}/holdings/2284621870006421")
```
hackartisan, line 387: Can the first stub also be removed?

app/adapters/alma_adapter/bib_item_set.rb
```diff
@@ -13,51 +13,6 @@ def initialize(items:, adapter:)
       @adapter = adapter
     end
 
-    def each(&block)
```
hackartisan, line 16: A few of these seem like they could be nice to have around, but I guess we can always put them in again if we need them and if they're not tested they're probably a liability. Just curious your thoughts here for `each`, `holding_ids`, `mms_id`.

hackartisan (changes requested):
This looks good to me; I'm mostly wondering if there are a few more places we can delete lines of code; see inline comments.

app/adapters/alma_adapter/bib_item_set.rb
tpendragon: Keeping `each` seems nice, otherwise it's not really a composite object. The others seem fine to remove until we need them.

eliotjordan: OK. There was no test for it, so I removed it.

spec/adapters/alma_adapter_spec.rb
eliotjordan: Good catch. Removed.

spec/adapters/alma_adapter_spec.rb
eliotjordan: Removed.

app/adapters/alma_adapter/bib_item_set.rb
eliotjordan: Added back and tested.

eliotjordan:
@hackartisan @tpendragon Thanks for the comments. This is ready for re-review.

hackartisan (dismissed):

hackartisan (approved):

PR #1077, opened 2021-02-12
Return 404 when the SCSB endpoint hits suppressed records

### LINKED ISSUE #1046: SCSB endpoint should return 404 when it his suppressed records
SCSB will ask for suppressed records because sometimes these are sent to Recap. We can't get them from the Alma API, though. In the code, we get the item and the holdings but then the request for the bib record returns an error (e.g. https://api-na.hosted.exlibrisgroup.com/almaws/v1/bibs/9958708973506421?apikey=)

We can return a 404 for this; distinguishing between this and the other 404s wouldn’t be helpful.

honeybadger issue is 
https://app.honeybadger.io/projects/54497/faults/74759178

### Use case background

When Recap staff receive an item and scan it into the the SCSB system (aka “accession” it), their system hits the bibdata scsb endpoint to pull in the data. If there’s any problem at this point the item will have status “incomplete”. If the problem was a 500 or other error from bibdata, it gets “Dummy Title” and “Dummy Author” as the data values.

Recap staff regularly pull up and investigate the report of incomplete records, so there is already workflow around handling errors.

The records should clear from incomplete status once #1049 is implemented because Recap staff will re-save the records that don't look right to get them to re-export to SCSB.

eliotjordan:
Closes #1046 

Fixes a few skipped rubocops.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -6,11 +6,6 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 1
-Lint/AmbiguousOperator:
-  Exclude:
-    - 'app/models/dump.rb'
-
 # Offense count: 6
 Lint/AmbiguousRegexpLiteral:
   Exclude:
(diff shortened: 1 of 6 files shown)


hackartisan (approved):
looks great!

PR #1088, opened 2021-02-16
Index incremental dumps

### LINKED ISSUE #809: Incremental index from Alma
depends on #927

Incremental indexing will use the periodic file dumps produced by Alma. Files are created for records added deleted. This code also needs to delete suppressed records from the index. https://github.com/pulibrary/marc_liberation/issues/824 

Implementation:

- Update file frequency should be set at "the every 6 hours" (4am, 10am, 4pm, 10pm) option. This most closely matches our current 3x/day schedule.
- Build on work done in #927 such that once the dump files are created, they are indexed.

Considerations:

- There is an hourly update option available, but we're not sure if our updates ever take longer than an hour. We decided to use the "every 6 hours" option to avoid investigating questions about conflicting incremental updates for now.
- If conflicting incremental updates are a concern, only one incremental index should be allowed to run at a time to avoid a scenario where:
  - a long-running index job starts
  - a more recent, but smaller, dump is created and starts to index
  - the long-running / older index job overwrites an update made by the more recent job with older data.

Outstanding questions:

* How are deleted records discerned from the Alma file dumps?
  * Deleted records are included in the same file as new and updated records; most likely deletion will be marked by position 5 of the leader of the MARC record being 'd' (https://www.loc.gov/marc/bibliographic/bdleader.html)
  * Confirm the above! I (@hackartisan) thought I saw filenames suggesting that added and deleted records were grouped in separate files.
* Consider doing a bunch of API updates as part of end-to-end testing.

hackartisan:
closes #809

Please squash merge!


christinach:
@hectorcorrea @hackartisan thanks!
can you please check circleci?

hectorcorrea:
I've pushed a fix. Hopefully this works.

spec/services/alma/indexer_spec.rb
```diff
@@ -56,4 +56,23 @@
       expect(response['response']['numFound']).to eq 1
     end
   end
+
+  describe "incremental_index!" do
+    it "indexes a dump's files" do
+      # url = Rails.application.config.solr['url']
```
hectorcorrea, line 62: @hackartisan I reverted back to the old solr_url syntax (rather than using the value from Rails's config). If this works I can add a separate issue to figure out why the Rails config version does not work in Circle CI.

hackartisan: 👍🏻 thanks!

christinach (approved):

PR #1089, opened 2021-02-17
Delete from Solr records marked as deleted in MARC files

### LINKED ISSUE #824: Process suppressed and deleted bibs

hectorcorrea:
First pass at deleting records from Solr that have been marked as deleted in MARC files.

The delete code is inspired in the [code that Jonathan Rochkind posted](https://github.com/traject/traject/issues/166) but adapted for our needs and version of Traject.

Fixes #824


marc_to_solr/lib/solr_deleter.rb
```diff
@@ -0,0 +1,44 @@
...
+      url = @solr_url + "/update?commit=true&wt=json"
+      payload = "<delete>" + batch.map { |id| "<id>#{id}</id>" }.join("") + "</delete>"
+      content_type = "text/xml" # request payload is XML (even if the response is in JSON via the wt=json param)
+      http_post(url, payload, content_type)
+    end
+
+    def http_post(url, payload, content_type)
+      uri = URI.parse(url)
+      http = Net::HTTP.new(uri.host, uri.port)
+      if url.start_with?("https://")
```
hackartisan, line 35: I'm not sure this "`https` or not" code path is exercised by the test. Why can't we assume it will always be one or the other? Is this because of differences in test / production setups? Elsewhere in marc_liberation we use Faraday; does it simplify this interaction? Does it make sense to try Faraday here?

marc_to_solr/lib/solr_deleter.rb
```diff
@@ -0,0 +1,44 @@
...
+# and their implementation does not support batches.
+class SolrDeleter
+  def initialize(solr_url)
+    @solr_url = solr_url
+  end
+
+  def delete(ids, batch_size = 100)
+    ids.each_slice(batch_size) do |batch|
+      response = delete_batch(batch)
+      raise StandardError, "Error deleting Solr documents. Status: #{response.code}. Body: #{response.body}" if response.code != "200"
```
hackartisan, line 18: I think we'll want the list of ids to get logged, as well.

I also have some questions here which I think might make sense for a follow-up ticket.

1. In other places where we do batch operations, we retry one-at-a-time on failures like this. Perhaps we should log the error like this, then recurse with a smaller batch size (maybe go straight to batch_size 1)? 
1. What reaction is required if a record or a batch of records fails to get deleted? I.e. is some human intervention required? It seems likely that it would be, so this might be a good place for a honeybadger notification so that we are alerted to look into it.

hackartisan (changes requested):
This looks great; I left a few question inline.

marc_to_solr/lib/solr_deleter.rb
hectorcorrea: I've updates the code to use Faraday (and this logic is now gone)

marc_to_solr/lib/solr_deleter.rb
hectorcorrea: I've added logging using Traject's native logger (if available). I am not sure if/how this would connect to HoneyBadger but at least deletes would be logged as any other Traject operation.

hectorcorrea: If we want to try failed deletes with smaller batches I would agree that we should create a separate ticket.

hackartisan (commented):
This looks good to me, just one small rubocop issue.

marc_to_solr/lib/traject_config.rb
```diff
@@ -35,10 +35,12 @@
 
 id_extractor = Traject::MarcExtractor.new('001', first: true)
 deleted_ids = Concurrent::Set.new
-each_record do |record, _context|
+each_record do |record, context|
   # Collect records that need to be deleted
+  # and skip processing logic for them.
   if record.leader[5] == 'd'
     id = id_extractor.extract(record).first
+    context.skip!("#{id} marked as deleted")
```
hackartisan, line 43: Good catch!!

hackartisan (approved):

christinach (approved):

PR #1090, opened 2021-02-17
Add discharge endpoint

### LINKED ISSUE #1072: Discharge an item
For SCSB, we need the ability to discharge an item from the ReCAP circulation desk.

SCSB is aware of the MMS Id, Holding Id, and Item Id, so it should not require an item lookup by barcode before doing it.

proposed route: `post /bibliographic/:mms_id/holdings/:holding_id/items/:item_id/discharge`

The endpoint should require a valid, hard-coded auth token.

The alma endpoint to use is as follows. Note that the values provided for these parameters won't vary, just the ids sent in the api path.
```
url = 'https://api-na.hosted.exlibrisgroup.com'
response = alma_conn(url).post do |req|
  req.url 'almaws/v1/bibs/9968643943506421/holdings/22258767470006421/items/23258767460006421'
  req.headers['Content-Type'] = 'application/json'
  req.headers['Accept'] = 'application/json'
  req.params['apikey'] = [key]
  req.params['op'] = 'scan'
  req.params['external_id'] = 'false'
  req.params['library'] = 'recap'
  req.params['circ_desk'] = 'DEFAULT_CIRC_DESK'
  req.params['done'] = 'true'
end
```

hackartisan:
closes #1072

companion pr: https://github.com/pulibrary/princeton_ansible/pull/2127

### DIFF
--- a/spec/controllers/bibliographic_controller_spec.rb
+++ b/spec/controllers/bibliographic_controller_spec.rb
@@ -230,4 +230,35 @@ RSpec.describe BibliographicController, type: :controller do
       end
     end
   end
+
+  describe '#item_discharge' do
+    before do
+      stub_request(:post, /.*\.exlibrisgroup\.com\/almaws\/v1\/bibs\/.*\/holdings\/.*\/items\/.*/)
+        .with(headers: { 'Authorization' => 'apikey TEST_WR_KEY' })
+        .to_return(status: 200,
+                   headers: { "Content-Type" => "application/json" },
+                   body: Rails.root.join('spec', 'fixtures', 'files', 'alma', 'scan_23258767460006421.json'))
+    end
+
+    it "scans and returns the item, using the correct alma api key and re-setting the key after" do
+      post :item_discharge, params: { mms_id: "9968643943506421", holding_id: "22258767470006421", item_pid: "23258767460006421", auth_token: "hard_coded_secret" }, format: :json
+      expect(response).to have_http_status(:success)
+      expect(response.body).to start_with "{\"bib_data\":{\"mms_id\":\"9968643943506421\""
+      expect(Alma.configuration.apikey).to eq "TESTME"
+    end
+
+    context "when no auth token is provided" do
+      it "returns 401 unauthorized" do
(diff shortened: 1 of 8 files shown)


hackartisan:
This is ready for review. It's deployed and can be tested with a curl command like:

```
curl -d "auth_token=[htc_auth_token]" -X POST "https://bibdata-alma-staging.princeton.edu/bibliographic/9968643943506421/holdings/22258767470006421/items/23258767460006421
```

The auth token is in the shared lastpass drive at location `Shared-ITIMS-Passwords/bibdata/htc_auth_token`

spec/controllers/bibliographic_controller_spec.rb
```diff
@@ -230,4 +230,35 @@
...
+  describe '#item_discharge' do
+    before do
+      stub_request(:post, /.*\.exlibrisgroup\.com\/almaws\/v1\/bibs\/.*\/holdings\/.*\/items\/.*/)
+        .with(headers: { 'Authorization' => 'apikey TEST_WR_KEY' })
+        .to_return(status: 200,
+                   headers: { "Content-Type" => "application/json" },
+                   body: Rails.root.join('spec', 'fixtures', 'files', 'alma', 'scan_23258767460006421.json'))
+    end
+
+    it "scans and returns the item, using the correct alma api key and re-setting the key after" do
```
christinach, line 243: Can we add in parenthesis after `scans and returns`, (discharges).

christinach (changes requested):
@hackartisan it looks good. one simple comment.

spec/controllers/bibliographic_controller_spec.rb
hackartisan: Yes but I was using "returns" as in "the endpoint returns the item data", not as in "the system returns / checks in the book" 😁 I just pushed an edit; let me know if it's more clear.

christinach (approved):

christinach (approved):

christinach:
We tested it also with @mzelesky and it puts the `item in place` in alma.

hackartisan:
@christinach @mzelesky thanks for testing it from the alma side.

PR #1097, opened 2021-02-18
holding_library is used in advanced search

christinach:
In the alma checkin @mzelesky helped us to understand that the holding_library in traject is used for the advanced facet. We added a comment. We validated this by doing a keyword search using bib=995415, https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=995415. In the Library facet, `Firestone Library` has only 1.

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -1189,6 +1189,8 @@ each_record do |record, context|
     context.output_hash['location_code_s'] = location_codes
     context.output_hash['location'] = Traject::TranslationMap.new("location_display").translate_array(location_codes)
     mapped_codes = Traject::TranslationMap.new("locations")
+    # The holding_library is used with some locations to add an additional owning library,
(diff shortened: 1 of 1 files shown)


marc_to_solr/lib/traject_config.rb
```diff
@@ -1189,6 +1189,7 @@
     context.output_hash['location_code_s'] = location_codes
     context.output_hash['location'] = Traject::TranslationMap.new("location_display").translate_array(location_codes)
     mapped_codes = Traject::TranslationMap.new("locations")
+    # It adds another owning library for the advanced search.
```
hackartisan, line 1192: I think it would be helpful to flesh this out a little bit more. Maybe "the holding_library is used with some locations to add an additional owning library, which is included in advanced search but not facets."

hackartisan (changes requested):

hackartisan (approved):

PR #1098, opened 2021-02-18
Documents why we need this class and how to use it.

hectorcorrea:


app/services/indexer_service.rb
```diff
@@ -1,3 +1,11 @@
+# Allows us to process a MARC record through Traject
+# as a class without having to invoke Traject from
```
christinach, line 2: Can you remove `us` ?

christinach (changes requested):

app/services/indexer_service.rb
hectorcorrea: Removed `us` ;)

christinach (approved):

PR #1110, opened 2021-02-23
Fix and document discharge endpoint

hackartisan:
* Add `/discharge` to api route
* Return non-200 status when we get an error from Alma
* Document the endpoint

### DIFF
--- a/docs/api_endpoints.md
+++ b/docs/api_endpoints.md
@@ -1,0 +1,171 @@
+# API Endpoints provided by marc_liberation / Bibdata
+
+This documentation does not include all API endpoints. See [config/routes.rb](../config/routes.rb) for more endpoints.
+
+## POST '/bibliographic/:mms_id/holdings/:holding_id/items/:item_pid/discharge' (item discharge)
+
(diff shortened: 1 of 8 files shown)


christinach (approved):

docs/api_endpoints.md
```diff
@@ -0,0 +1,172 @@
+# API Endpoints provided by marc_liberation / Bibdata
+
+This documentation does not include all API endpoints. See [config/routes.rb](../config/routes.rb) for more endpoints.
+
+## POST '/bibliographic/:mms_id/holdings/:holding_id/items/:item_pid/discharge' (item
+discharge)
+
```
christinach, line 7: @hackartisan looks great. Can you please adjust the (item discharge ) comment so that it has the same format?

christinach (changes requested):

docs/api_endpoints.md
hackartisan: done, thanks

christinach (approved):

PR #1112, opened 2021-02-23
Implements method to get catalog_date

### LINKED ISSUE #925: reindex single record: catalog_date field
**Single Record**
Child of https://github.com/pulibrary/marc_liberation/issues/748 -- see implementation notes there!

1. Get a single alma record with `p_avail` and `e_avail` parameters: https://developers.exlibrisgroup.com/console/?url=/wp-content/uploads/alma/openapi/bibs.json#/Catalog/get%2Falmaws%2Fv1%2Fbibs%2F%7Bmms_id%7D

2. If it has any `AVA` fields that correspond to `p_avail` (physical availability), get the list of items to get the `<creation_date>` . Get the earliest `creation_date` from the <items> list.

3. If there are no `AVA` fields that correspond to `p_avail`, look for `AVE` fields that correspond to `e_avail`. Each `AVE` field has a portfolio id included. For each `AVE` field in the API response, you must retrieve the portfolio information. Use the oldest `<portfolio><activation_date>` from the attached portfolios. See https://developers.exlibrisgroup.com/alma/apis/docs/xsd/rest_portfolio.xsd/

4. If there is no `p_avail` or `e_avail`, use `<bib><created_date>`.

hectorcorrea:
Closes #925

### DIFF
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -48,9 +48,44 @@ class AlmaAdapter
     bib_item_set.holding_summary
   end
 
+  # @param id [String]. e.g id = "991227850000541"
+  # @return [String] date record was created e.g. "2019-10-18Z"
+  def get_catalog_date(id)
+    record = get_bib_record(id)
+    get_catalog_date_from_record(record)
+  end
+
   private
 
     def apikey
       Rails.configuration.alma[:read_only_apikey]
+    end
+
+    # @param record [AlmaAdapter::MarcRecord]
+    # @return [String] date record was created e.g. "2019-10-18Z"
+    def get_catalog_date_from_record(record)
+      return nil if record.nil?
+      ava = record.select { |f| f.tag == "AVA" }
+      if ava.count > 0
+        # Get the creation date from the physical items
+        dates = []
+        get_items_for_bib(record.bib.id).each do |_location, holdings|
+          items = holdings.map { |h| h["items"] }.compact.flatten
+          dates += items.map { |i| i["creation_date"] }.compact.flatten
+        end
+        return dates.sort.first if dates.count > 0
+      end
+
+      # Note: For now we are not searching for the activation date in portfolios
+      # indicated in the AVE fields.
+      #
+      # If needed we could use the alma/:mms_id/portfolios/:portfolio_id endpoint
+      # to get this information one by one. ExLibris has indicated that this information
+      # will also be available on the alma/:mms_id/portfolios endpoint around May/2021
+      # (see https://3.basecamp.com/3765443/buckets/20717908/messages/3475344037)
+      # which will result in one call per bib rather than one call per portfolio.
+
+      # Default to the Bib record created date
+      record.bib["created_date"]
     end
 end
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -48,13 +48,6 @@ class AlmaAdapter
     bib_item_set.holding_summary
   end
 
-  # @param id [String]. e.g id = "991227850000541"
-  # @return [String] date record was created e.g. "2019-10-18Z"
-  def get_catalog_date(id)
-    record = get_bib_record(id)
-    get_catalog_date_from_record(record)
-  end
-
   # @param record [AlmaAdapter::MarcRecord]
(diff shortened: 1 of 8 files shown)


app/adapters/alma_adapter.rb
```diff
@@ -48,9 +48,44 @@ def get_items_for_bib(id)
...
+    # @return [String] date record was created e.g. "2019-10-18Z"
+    def get_catalog_date_from_record(record)
+      return nil if record.nil?
+      ava = record.select { |f| f.tag == "AVA" }
+      if ava.count > 0
+        # Get the creation date from the physical items
+        dates = []
+        get_items_for_bib(record.bib.id).each do |_location, holdings|
+          items = holdings.map { |h| h["items"] }.compact.flatten
+          dates += items.map { |i| i["creation_date"] }.compact.flatten
```
hackartisan, line 74: So, in https://github.com/pulibrary/marc_liberation/pull/1104 I removed a bunch of keys from getting returned by `get_items_for_bib`, and now creation date will never be in there. The test still passes because the bib record and the item record have the same creation date, so it falls back to the bib and it's right. Maybe we can get together and figure out how to adjust the code. I was worried this exact thing would happen when I made that change -- at that time nothing else was calling get_items_for_bib, but I wondered "what if someone comes along and wants to use it?"...

app/adapters/alma_adapter.rb
```diff
@@ -48,9 +48,44 @@ def get_items_for_bib(id)
     bib_item_set.holding_summary
   end
 
+  # @param id [String]. e.g id = "991227850000541"
+  # @return [String] date record was created e.g. "2019-10-18Z"
+  def get_catalog_date(id)
```
hackartisan, line 53: I think we want this logic in the `AlmaAdapter::MarcRecord` so that it adds the catalog date to a field that will be indexed by the logic added in https://github.com/pulibrary/marc_liberation/pull/945/files.

It should be a method called `enrich_with_catalog_date` or something like that, that we can then call from the bibliographic#update action once we've fetched the fields.

It's probably fine to just find it and put it in the `876d`; I don't think we need to worry about only using that for a value found in the bib (putting it in the `951w` if it came from the physical item and the `950b` if it came from the electronic item).

hackartisan (changes requested):
The logic seems right according to the issue; I think we just need to move it and put the data the desired marc field.

app/adapters/alma_adapter.rb
hectorcorrea: I reintroduced the creation_date as a field that we return (since we already had it) and this is not an issue anymore.

app/adapters/alma_adapter.rb
hectorcorrea: Working on some ideas for this.

hectorcorrea: Done. Take a look and let me know what you think.

app/adapters/alma_adapter.rb
```diff
@@ -48,6 +48,42 @@ def get_items_for_bib(id)
...
```
hackartisan, line 53: Do we need the `get_catalog_date` method or can we just remove this?

hackartisan (changes requested):
This looks good thank you for talking through the use case with me. I think I'll make a follow up issue to refactor the holding_summary field filtering up closer to the controller action that depends on it and bulk up the controller tests for the bibliographic#bib_items.

One other question in line.

app/adapters/alma_adapter.rb
hectorcorrea: We don't need _need_ the `get_catalog_date(id)` method but it keeps the test simple (i.e. allow us to use similar fixtures to existing tests that also emulate Alma API responses). So I would be in favor for keeping it.

hectorcorrea:
@hackartisan Thanks for all the feedback and working around the nuances. 

I added a test for the actual record enrichment in `marc_record.rb` -- This should be the last review (famous last words)...

app/adapters/alma_adapter.rb
hectorcorrea: Good point. Done.

hackartisan (approved):
👍🏻 🎉 Yay thank you Hector! Looks great.

PR #1125, opened 2021-03-02
Refactor get_items_for_bib to return the BibItemSet and add a filter option to BibItemSet#holdings_summary

### LINKED ISSUE #1119: Refactor the holding_summary field filtering so it happens in or closer to bibliographic#bib_items action
Refactor the holding_summary field filtering up closer to the controller action that depends on it.

This will make it more clear that the field filtering is in support of the endpoint requirements.

Bulk up the controller tests for the bibliographic#bib_items.

hackartisan:
This moves the logic of which keys are returned directly to the controller, and allows simplified logic in get_catalog_date_for_record

closes #1119

### DIFF
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -43,8 +43,8 @@ class AlmaAdapter
   # @return [AlmaAdapter::BibItemSet]
   def get_items_for_bib(id)
     opts = { limit: 100, expand: "due_date_policy,due_date", order_by: "library", direction: "asc" }
-    bib_item_set = Alma::BibItem.find(id, opts).map { |item| AlmaAdapter::AlmaItem.new(item) }
-    AlmaAdapter::BibItemSet.new(items: bib_item_set, adapter: self)
+    bib_items = Alma::BibItem.find(id, opts).map { |item| AlmaAdapter::AlmaItem.new(item) }
+    AlmaAdapter::BibItemSet.new(items: bib_items, adapter: self)
   end
 
   # @param record [AlmaAdapter::MarcRecord]
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -40,12 +40,11 @@ class AlmaAdapter
   end
 
   # @param id [String]. e.g id = "991227850000541"
-  # @return [Hash] of locations/ holdings/ items data
+  # @return [AlmaAdapter::BibItemSet]
   def get_items_for_bib(id)
     opts = { limit: 100, expand: "due_date_policy,due_date", order_by: "library", direction: "asc" }
     bib_item_set = Alma::BibItem.find(id, opts).map { |item| AlmaAdapter::AlmaItem.new(item) }
-    bib_item_set = AlmaAdapter::BibItemSet.new(items: bib_item_set, adapter: self)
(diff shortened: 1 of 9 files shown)


app/controllers/bibliographic_controller.rb
```diff
@@ -92,15 +92,19 @@ def bib_holdings
...
   def bib_items
-    records = adapter.get_items_for_bib(bib_id_param)
-    if records.nil? || records.empty?
-      render plain: "Record #{params[:bib_id]} not found or suppressed", status: 404
-    else
-      respond_to do |wants|
-        wants.json  { render json: MultiJson.dump(add_locator_call_no(records)) }
-        wants.xml { render xml: '<todo but="You probably want JSON anyway" />' }
-      end
+    item_keys = ["id", "pid", "perm_location", "temp_location"]
```
hectorcorrea, line 95: I love this, filtering the keys at the controller level makes a lot of sense.

app/adapters/alma_adapter/bib_item_set.rb
```diff
@@ -19,27 +19,31 @@ def each(&block)
...
     end
 
-    def holding_items_filter(items)
-      items.map(&:as_json).map do |h|
-        h.keep_if do |k, _v|
-          ["id", "pid", "perm_location", "temp_location", "creation_date"].include? k
+    private
+
+      def holding_items_filter(items, keys)
+        return items unless keys
```
hectorcorrea, line 41: This might be clearer if the `keys` parameter was named `filter`

app/adapters/alma_adapter.rb
```diff
@@ -56,11 +55,8 @@ def get_catalog_date_from_record(record)
     ava = record.select { |f| f.tag == "AVA" }
     if ava.count > 0
       # Get the creation date from the physical items
-      dates = []
-      get_items_for_bib(record.bib.id).each do |_location, holdings|
-        items = holdings.map { |h| h["items"] }.compact.flatten
-        dates += items.map { |i| i["creation_date"] }.compact.flatten
-      end
+      item_set = get_items_for_bib(record.bib.id)
+      dates = item_set.items.map { |i| i["item_data"]["creation_date"] }
```
hectorcorrea, line 59: I am not sure the `creation_date` will always be there, maybe re-adding the `.compact` would be a good idea to prevent `nil` values into the `dates` array.

app/controllers/bibliographic_controller.rb
```diff
@@ -92,15 +92,19 @@ def bib_holdings
...
+    records = adapter.get_items_for_bib(bib_id_param).holding_summary(item_key_filter: item_keys)
```
hectorcorrea, line 96: nit-picking: Maybe rename to `record` (singular) since we only handle one record ?

hectorcorrea (commented):
Made a few comments. Sorry about the spam :)

app/controllers/bibliographic_controller.rb
hackartisan: what about `holding_summary`?

hackartisan:
Thank you @hectorcorrea! I made the changes you suggested if you'd like to give it another look.

hectorcorrea (approved):

PR #1129, opened 2021-03-08
Update circleci and inchci badges in readme

christinach:


christinach:
The inchci server is currently down, this is why it looks broken.

hackartisan (dismissed):

christinach:
The coveralls I kept it with the marc_liberation repo name. It is still reading from that.

christinach:
I added a new coveralls token for the bibdata repo in circleci.

PR #1135, opened 2021-03-12
Remove extra library label from locations

### LINKED ISSUE #1133: Refactor the LocationMapsGeneratorService so that we don't take an extra label in the locations template

christinach:
closes #1133

Update generate method and remove extra library label from locations.rb file.
Generate new template files for holding library and locations.
Remove rake db:seed task, we don't seed anything.


christinach:
Removes the duplicate label from the holding libraries when the library name and the location name are the same. 

Screen Shot 2021-03-17 at 8 39 15 PM

christinach:
I noticed that the code 'warf' is not included in Alma::Location. Based on the previous template it refers to 'Western Americana: Reference Collection (WARF)'. @kevinreiss @mzelesky does this code need to be added in Alma? Or has it been replaced with another location code?

christinach:
Orangelight search results with no duplicate label for the library.

Screen Shot 2021-03-17 at 10 25 51 PM

app/services/location_maps_generator_service.rb
```diff
@@ -20,22 +20,39 @@ def generate
...
   end
 
+  # Select the holding_libraries based on the holding location code from:
+  # https://github.com/pulibrary/bibdata/blob/main/marc_to_solr/translation_maps/holding_library.rb.tmpl
+  def holding_libraries
+    holding_codes = ["recap$gp", "recap$pb", "recap$pf", "recap$pg", "recap$ph", "recap$pj",
+                     "recap$pk", "recap$pl", "recap$pm", "recap$pn", "recap$pq", "recap$ps", "recap$pt", "recap$pw",
+                     "recap$pz", "recap$qk", "recap$ql", "recap$qv", "recap$xc", "recap$xg", "recap$xm", "recap$xn",
+                     "recap$xp", "recap$xr", "recap$xx", "recap$warf"]
+    locations.select { |l| holding_codes.include?(l.code) }
```
christinach, line 53: I couldn't find in Alma location code warf which refers to the Western Americana: Reference Collection (WARF)'. @mzelesky . Is this a location that we need to add in Alma ?

marc_to_solr/translation_maps/holding_library.rb.tmpl
```diff
@@ -1,26 +1,25 @@
...
+ "recap$qk"=>"ReCAP - rcpqk RECAP Music Lib.",
+ "recap$ql"=>"ReCAP - rcpql RECAP EA Microforms",
+ "recap$qv"=>"ReCAP - rcpqv RECAP Video Coll.",
+ "recap$xc"=>"ReCAP - rcpxc RECAP Cotsen Lib.",
+ "recap$xg"=>"ReCAP - rcpxg RECAP Graphic Arts",
+ "recap$xm"=>"ReCAP - rcpxm RECAP Manuscripts",
+ "recap$xn"=>"ReCAP - rcpxn RECAP Numismatics",
+ "recap$xp"=>"ReCAP - rcpxp RECAP Historic Maps",
+ "recap$xr"=>"ReCAP - rcpxr RECAP Rare Books",
+ "recap$xx"=>"ReCAP - rcpxx RECAP Special"}
```
christinach, line 25: These are the holding libraries that I get from Alma. Will we update the names in Alma? Should I proceed and add a mapping based on the old file? @mzelesky

app/services/location_maps_generator_service.rb
mzelesky: WARF is a location under the 'rare' library, not recap. At least, I see it in the Alma UI.

christinach: Thank you @mzelesky . So it will be rare$warf.

app/services/location_data_service.rb
```diff
@@ -25,7 +25,7 @@ def populate_holding_locations
       library_record = Locations::Library.find_by(code: library.code)
       holding_locations(library.code).each do |holding_location|
         Locations::HoldingLocation.new do |location_record|
-          location_record.label = "#{library_record.label} - #{holding_location.name}"
+          location_record.label = (library_record.label != holding_location.name ? "#{library_record.label} - #{holding_location.name}" : library_record.label)
```
hackartisan, line 28: Can we switch the order of the ternary around so that it compares on equality like `library_record.label == holding_location.name`

app/services/location_maps_generator_service.rb
```diff
@@ -20,22 +20,39 @@ def generate
...
 
     lib_display = {}
-    longer_display = {}
+    locations_display = {}
     holding_library = {}
     locations.each do |holding|
       holding_code = holding['code']
       lib_label = holding.library['label']
-      holding_label = holding['label'] == '' ? lib_label : lib_label + ' - ' + holding['label']
+      holding_label = holding['label'] == '' ? lib_label : holding['label']
```
hackartisan, line 28: Is there a test case where `holding['label']` is empty? Curious why we account for this.

app/services/location_maps_generator_service.rb
```diff
@@ -20,22 +20,39 @@ def generate
...
     end
 
     write_file('location_display.rb', lib_display)
-    write_file('locations.rb', longer_display)
+    write_file('locations.rb', locations_display)
     write_file('holding_library.rb', holding_library)
   end
 
+  # Select the holding_libraries based on the holding location code from:
+  # https://github.com/pulibrary/bibdata/blob/main/marc_to_solr/translation_maps/holding_library.rb.tmpl
```
hackartisan, line 47: Are these the locations that aren't in alma? Can you just add to the comment that they're not in alma because they are maintained by ReCAP's inventory management system? (I think that's why)

hackartisan (changes requested):

app/services/location_maps_generator_service.rb
christinach: I will look but I don't think I saw one. If there is no empty holding['label'] I'll remove the check.

app/services/location_maps_generator_service.rb
christinach: These are locations in Alma. The template file which is currently in production includes only these locations codes. 
I think it is related to a comment we added a while ago in traject about the holding library. `The holding_library is used with some locations to add an additional owning library, which is included in advanced search but not facets.` This is the file that traject is reading from (holding_library.tmpl). https://github.com/pulibrary/bibdata/blob/6a0a6b6d5d4f991e4e959b0a774933e7ea688b42/marc_to_solr/lib/traject_config.rb#L1193

marc_to_solr/translation_maps/holding_library.rb.tmpl
christinach: Well I dont think it's necessary to add a mapping for the names right now. When they will updated in Alma we will pull them again. @mzelesky can you confirm?

christinach:
@hackartisan can you check again?

hackartisan (changes requested):
👍🏻 It looks good thanks. I just want to have a little time in checkin to talk about the holding library stuff with @kevinreiss . blocking merge until then.

app/services/location_maps_generator_service.rb
hackartisan: I see. Okay I guess my question then is why do we prefer to list them out this way? Were they not pulling out of alma with the check for `holding.holding_library`?

christinach: No they were not. The way the code was set up in the alma branch would generate an empty holding_library.rb file.
Also holding.holding_library was nil in the alma branch. My guess is that the holding_library.tmpl file was a copy and paste from the production file with an update on the location_codes based on the alma criteria. At least this is what I came across when I tried to generate the files.

christinach: @mzelesky said that this file holding_library.tmpl shows which library handles items for these locations. Currently this is not maintained in Alma. @mzelesky is investigating if this can be implemented in Alma. But for now I will update the code to include the hash as it is in production with the new location codes.

christinach:
In the checkin we said to remove the tmpl files and keep only the .rb files. After looking into this I rejected it as an option. I'm quoting my note from slack `I kept solid only the holding_library.tmpl file so that we don’t need to generate it every time.
I rejected the idea of using the .rb files that we generate through the #generate method because they are being generated by retrieving data from the tables. With that said every time you start lando and you generate the .rb files using the tables from the database the tables would be empty and the generated .rb files would be empty.  Unless you also call the alma api every time you start lando so that you populate the tables.  We don’t want to call the alma api every time we run the app in development or test. So it is better to keep the tmpl files and use them through the #generate_from_templates method so that we will create the .rb files locally in case they don’t exist`.
Also this Pr addresses the removal of the double label in the orangelight display and updates the Location::Holdings table in bibdata. It also improves the code by removing unused lines of code and the generation of a nil file. 

If we want to improve the way we generate the locations this should be part of a new issue.

hackartisan (approved):

PR #1159, opened 2021-03-24
Update all holding fields in traject to account for subfield 8 and subfield 0

### LINKED ISSUE #1158: 'Location has' is not indexed properly
Update holding fields in princeton_marc.

christinach:
closes #1158


christinach:
Screen Shot 2021-03-24 at 6 10 08 PM

christinach:
Princeton record. Compare to https://catalog.princeton.edu/catalog/857469
Screen Shot 2021-03-24 at 6 36 33 PM

marc_to_solr/lib/princeton_marc.rb
```diff
@@ -638,13 +638,16 @@ def process_holdings record # rubocop:disable Metrics/AbcSize, Metrics/Cyclomati
...
     is_scsb = scsb_doc?(record['001'].value)
     field.subfields.each do |s_field|
-      if s_field.code == '8'
-        holding_id = s_field.value if is_alma || is_scsb
+      # Index holdings from Princeton records
+      if s_field.code == '8' && is_alma
+        holding_id = s_field.value
+      # Index holdings from SCSB
       elsif s_field.code == '0'
         holding_id = s_field.value
```
mzelesky, line 646: Should this have the `is_scsb` check?

christinach: No it shouldn't. The scsb check was added to make sure that we index the scsb because we have the check for the alma valid princeton record.

mzelesky (changes requested):
Either the holding fields need to be filtered before passing the fields into this method, or fields that fail the `is_alma` and the `is_scsb` checks should be skipped within the method.

christinach:
@mzelesky I updated. Thank you for reviewing and for the valuable feedback.

christinach:
Compare to https://catalog.princeton.edu/catalog/SCSB-8157262
Screen Shot 2021-03-25 at 11 42 02 AM

mzelesky (approved):
This PR looks good. It makes it very clear what the different parsing rules are for Alma records vs. partner SCSB records, and it skips fields that should be skipped.

PR #1161, opened 2021-03-25
Always copy translation map templates

### LINKED ISSUE #1060: Updated template files do not load in development
They are only copied if the files didn't exist. This means they aren't updated when they've changed

https://github.com/pulibrary/marc_liberation/blob/590c0712fe4b04114c69ce9998c7e7238ebfb858/app/services/location_maps_generator_service.rb#L50

Should they just be copied every time? Do we have any reason to keep / edit local copies of these?

eliotjordan:
Closes #1060

### DIFF
--- a/spec/services/location_maps_generator_service_spec.rb
+++ b/spec/services/location_maps_generator_service_spec.rb
@@ -5,10 +5,12 @@ RSpec.describe LocationMapsGeneratorService do
   let(:base_path) { './tmp/' }
   let(:locations_path) { File.join(base_path, 'locations.rb') }
   let(:location_display_path) { File.join(base_path, 'location_display.rb') }
+  let(:holding_library_path) { File.join(base_path, 'holding_library.rb') }
 
   after do
-    FileUtils.rm(locations_path)
-    FileUtils.rm(location_display_path)
+    FileUtils.rm(locations_path) if File.exist?(locations_path)
+    FileUtils.rm(location_display_path) if File.exist?(location_display_path)
+    FileUtils.rm(holding_library_path) if File.exist?(holding_library_path)
   end
 
   describe '.generate' do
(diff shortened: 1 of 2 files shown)


spec/services/location_maps_generator_service_spec.rb
```diff
@@ -46,5 +46,16 @@
...
+    context 'with existing template files' do
+      before do
+        allow(FileUtils).to receive(:cp)
+      end
+
+      it 'overwrites the templates' do
+        described_class.generate_from_templates
+        expect(FileUtils).to have_received(:cp).exactly(3).times
+      end
+    end
```
christinach, line 59: we want to overwrite the 2 templates. The one is custom and solid and we dont create the 3rd .rb for this one.

christinach (changes requested):

spec/services/location_maps_generator_service_spec.rb
christinach: I'm talking about this template. Should not be overwritten. https://github.com/pulibrary/bibdata/blob/1060-templates/marc_to_solr/translation_maps/holding_library.rb.tmpl

eliotjordan: @christinach The templates are not overridden. Only the `.rb` translation maps. I restructured the test so it works better and is clearer.

eliotjordan: If you think that we should check in `holding_library.rb` to the repo instead of copying from a template, I would open a new issue.

christinach: > @christinach The templates are not overridden. Only the `.rb` translation maps. I restructured the test so it works better and is clearer.

Thanks @eliotjordan !

christinach: > If you think that we should check in `holding_library.rb` to the repo instead of copying from a template, I would open a new issue.

We should copy from the template.

christinach (approved):

PR #1174, opened 2021-03-30
Use LocationMapsGeneratorService#generate to extract the locations when UPDATE_LOCATIONS = true

### LINKED ISSUE #1173: update_location should not look bibdata production when we are indexing in Alma
In production we run the full reindex and the scsb index with the flag UPDATE_LOCATIONS true. This is because we use the UI in bibdata to update a location, when a location changes in voyager. So we want to make sure we catch the changes. 

In Alma we should continue doing something similar but in my opinion we should use the #generate method and rebuild the .rb, and not pull any potential updates through Faraday and the UI. (This is what we are currently doing in production)

christinach:
closes #1173

Use LocationMapsGeneratorService.generate to extract the locations when UPDATE_LOCATIONS=true

### DIFF
--- a/marc_to_solr/lib/location_extract.rb
+++ b/marc_to_solr/lib/location_extract.rb
@@ -1,34 +1,3 @@
-require 'json'
-require 'pp'
-require 'faraday'
-
 def update_locations
-  begin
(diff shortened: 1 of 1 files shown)


hackartisan (approved):
This looks like the same logic, just pulling it via active record instead of from the app's API. 👍🏻

PR #1191, opened 2021-03-31
Availability enhancements 

### LINKED ISSUE #1186: Availability status_label values need to be processed
- The values for status_label should be capitalized
- "check_holdings" should be translated to "Some items not available"

https://github.com/pulibrary/bibdata/blob/alma/app/adapters/alma_adapter/availability_status.rb#L34

### LINKED ISSUE #1188: Remove "more_items" key from bibliographic/[:id]/availablility and bibliographic/availability
This key is no longer needed by orangelight, since it was used to let us know to fetch all the availability data for each item, which we can no longer afford to do

hectorcorrea:
Improves the logic for the status_label value and removes a property no longer needed in the response.

Fixes #1186
Fixes #1188

### DIFF
--- a/app/adapters/alma_adapter/availability_status.rb
+++ b/app/adapters/alma_adapter/availability_status.rb
@@ -26,13 +26,13 @@ class AlmaAdapter
     # TODO: Confirm that we need all of these values with the Alma implementation.
     # TODO: Find out if we can get copy_number from Solr (i.e. without hitting ExLibris' API)
     def holding_status(holding)
+      status_label = Status.new(bib: bib, holding: holding, holding_item_data: nil).to_s
       status = if holding["holding_id"]
                  {
                    on_reserve: "N",
                    location: holding["library_code"] + "$" + holding["location_code"],
                    label: holding["location"],
-                   status_label: holding["availability"],
-                   more_items: (holding["total_items"] || "").to_i > 1,
+                   status_label: status_label,
                    holding_type: "physical",
                    id: holding["holding_id"]
                  }
@@ -42,8 +42,7 @@ class AlmaAdapter
                    on_reserve: "N",
                    location: "N/A",
                    label: "N/A",
-                   status_label: holding["activation_status"],
-                   more_items: nil,
+                   status_label: status_label,
                    holding_type: "portfolio",
                    id: holding["portfolio_pid"]
                  }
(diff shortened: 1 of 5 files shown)


app/adapters/alma_adapter/status.rb
```diff
@@ -14,6 +14,10 @@ def initialize(bib:, holding_item_data:, holding:)
   end
 
   def to_s
-    return "On-Site" if holding["availability"] == "available"
+    return "Some items not available" if holding["availability"] == "check_holdings"
+    return holding["availability"].titlecase if holding["availability"]
+
+    # For electronic holdings
+    return holding["activation_status"].titlecase if holding["activation_status"]
```
hackartisan, line 21: I'm not sure we are aware of this use case from the orangelight side; we might need more info.

hackartisan (approved):

app/adapters/alma_adapter/status.rb
hectorcorrea: This is the reason I have been hesitant to close the two tickets about availability (single bib and multi-bib). 

There are new cases that Alma brings that we didn't have to deal with in Voyager. Hence the note in the code: https://github.com/pulibrary/bibdata/blob/alma/app/adapters/alma_adapter/availability_status.rb#L28-L64.

hackartisan: Okay let's talk about it at an alma checkin.

hectorcorrea:
@hackartisan I changed the code to return "Available" rather than "On-Site" and I broke some tests. 

Is there a reason to have both (sometimes Available and some other times On-Site) for the status text? If not, I can update the test to reflect the new status value. Otherwise we'll need to figure out when to return one or the other.

app/adapters/alma_adapter/availability_status.rb
```diff
@@ -73,7 +71,7 @@ def holding_summary(holding)
       holding_item_data = item_data[holding["holding_id"]]
       status = Status.new(bib: bib, holding_item_data: holding_item_data, holding: holding)
       {
-        more_items: holding["total_items"].to_i > 1,
+        # more_items: holding["total_items"].to_i > 1,
```
eliotjordan, line 74: issue: Is this still needed?

eliotjordan (changes requested):
Looks good. Just one comment/change.

app/adapters/alma_adapter/availability_status.rb
hectorcorrea: Good catch. I've removed it.

eliotjordan (approved):

PR #1192, opened 2021-03-31
Move the UPDATE_LOCATIONS in an ENV in the full reindex and the scsb …

### LINKED ISSUE #1187: Update location_extract to recognize LocationMapsGeneratorService
Error loading configuration file marc_to_solr/lib/traject_config.rb:33 NameError:uninitialized constant LocationMapsGeneratorService

christinach:
closes #1187

Move the logic to generate the locations every time we run the full reindex and the scsb in an ENV.

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -23,7 +23,7 @@ every 1.day, at: '3:00am' do
   rake "liberate:arks:clear_and_seed_cache"
 end
 
-# Daily racap / partner jobs
+# Daily recap / partner jobs
 every 1.day, at: "3:00pm", roles: [:cron_production] do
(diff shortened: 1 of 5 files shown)


hackartisan (approved):
This seems good to me. And we'll likely redesign the location data pipeline so it makes sense to just do what woks right now.

Moving it out of traject means we have to place it everywhere we could possibly want to index records. This PR accounts for:
* Full record load
* manual marc XML file load
* 2 scsb indexing tasks

There are 2 other ways to index that I can think of
* through the bibliographic#update controller action, which is not yet implemented for alma.
* the incremental load, which is run by the IncrementalIndexJob -- do we want to add this there? We can probably put this off until we discuss the location data pipeline.

PR #1213, opened 2021-04-07
Add LocationToFileService to create holding and delivery json files

### LINKED ISSUE #1103: Populate location table flags
Follow-on to https://github.com/pulibrary/marc_liberation/issues/964

https://github.com/pulibrary/marc_liberation/blob/alma/app/services/location_data_service.rb uses the Alma API to pull the new libraries and locations into the locations database tables. However no flags have been set.

If these flags can be set via the Alma API, we need specifications for doing so. If they can't, they must be set directly, either by mapping existing locations to alma locations and reading the values from current production, or by hand in the bibdata-alma-staging UI.

Question: Will there be new locations added to Alma? Previously we were told no, those were locked, but recently there was talk of new Marquand locations. If there are new locations that is relevant here because the code that syncs the locations from Alma deletes / repopulates the table, so it would wipe out any flags that had been updated if it were re-run.

christinach:
closes #1103 

add alma voyager mapping file,
add holding_locations and delivery_locations files
Update LocationDataService to populate holding locations from alma and alma voyager mapping file
Update LocatonDataService to populate delivery locations using the delivery json file


christinach:
The delivery locations has a many to many relationship with the holding locations. I will update the code and use as a unique identifier the gfa_pickup attribute from the delivery locations. I will also update the holding locations json file with the delivery locations so that I can use it to populate the tables and assign delivery locations to the holding locations.

christinach:
The files with the mapped values are checked in config/locations/holding_locations.json and delivery_locations.json. 
If we want to recreate them we would need to run the rake task `bundle exec rspec locations:delivery_holding_locations`. 
To populate the tables locally or on bibdata we would need to run from the rails console `LocationDataService.delete_existing_and_repopulate`

christinach:
changing to a draft

christinach:
ok one problem that I see is that the gfa_pickup that I thought it was unique for the delivery locations it is not. So I need to change some things in the mapping. I came across that when I compared code: "gss"

christinach:
https://bibdata.princeton.edu/locations/holding_locations/gss.json, There are many delivery locations in 'Firestone' with gfa_pickup 'PG'

christinach:
@kevinreiss said that these are old data and we should not worry about it. They are all speciall collections. So I will deploy the locations on alma-staging.

christinach:
I updated the delivery and holding locations again based on the changes in #1261.

hackartisan (approved):

Issue #1203, opened 2021-04-02
Remove elf locations from access.rb


christinach:
Remove `elf` locations from bibdata; elfs are not valid locations and we will not have them anymore in Alma.
Some records they still have them because they were added after the P2E (Print to Electronic) file was used to import electronic records into Alma.

hackartisan:
@christinach is it true that this ticket can be closed when https://github.com/pulibrary/bibdata/blob/alma/marc_to_solr/translation_maps/access.rb does not contain any keys with "elf"?

christinach:
I think so. Because by removing them from the https://github.com/pulibrary/bibdata/blob/alma/marc_to_solr/translation_maps/access.rb file they will not appear under the online facet. But I would need to test it as well.

tpendragon:
@christinach @hackartisan Should we also remove https://github.com/pulibrary/bibdata/blob/alma/marc_to_solr/translation_maps/locations.rb.tmpl#L37-L40 ?

christinach:
@tpendragon yes. 
One way to do this is to wait for the cleanup in alma. Alma should not have elf locations at all. @mzelesky is there  a time for this cleanup?
The other way is not to count it when we pull the locations from Alma.

hackartisan:
The code that generates the database tables needs to filter these out.

tpendragon:
@hackartisan Should we close this since there's the other ticket?

hackartisan:
yes i meant to do that!

PR #1215, opened 2021-04-07
Remove ELFs from access.rb and tmpl files.

tpendragon:
I need to get the fixture out of the full dump that I need, but my alma user doesn't have permission to run publishing profiles. I'm blocked until I get the MARC record for 99121576653506421.

Work towards #1203


christinach (approved):
This is ok but when we run to generate the tables form alma the elf will be again in the tmpl files.

tpendragon:
> This is ok but when we run to generate the tables form alma the elf will be again in the tmpl files.

Yes, but the test will fail.

christinach:
the elfs need to be removed from alma or  to exclude them in this method https://github.com/pulibrary/bibdata/blob/a973974e41920cbf4507d5f854e400519defd5b5/app/services/location_data_service.rb#L23 when we populate the tables. 
Your pr will work as long as we dont update the tables from alma which I will do when I update the flags for my pr. 
I'm ok if you want to go ahead and merge this and I will exclude the elfs in https://github.com/pulibrary/bibdata/pull/1213

spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -49,6 +49,15 @@ def fixture_alma_record(fixture_name)
       access_links = record["electronic_access_1display"]
       expect(JSON.parse(access_links.first)).to eq("http://dx.doi.org/10.1007/BFb0088073" => ["dx.doi.org"])
     end
+    it "does not index elf locations for alma" do
+      # I need a record where this would break without this change - probably 99121576653506421
```
eliotjordan, line 53: Don't need this comment anymore.

eliotjordan (changes requested):
Small change requested, otherwise 👍

christinach:
> > This is ok but when we run to generate the tables form alma the elf will be again in the tmpl files.
> 
> Yes, but the test will fail.

Yes you are right the test will fail.

tpendragon:
To be clear, after deploying this we'll have to delete the DB entries for the elf locations from bibdata-alma.

christinach:
I'm sorry @tpendragon this needs a rebase

christinach:
It's ok to merge and have the test in place.

christinach (approved):

PR #1218, opened 2021-04-07
Download/Transfer Alma ReCAP Dump

### LINKED ISSUE #1049: Migrate recap dump to alma
We generate a dump of our recap records to send to the SCSB system each day.

This will be implemented as a scheduled publishing job.

Once the publishing job is created / configured, the name of the publishing job can be used in the sqs_poller to set up a dump creation process similar to the one for full and incremental dumps. Once the dump file is created, the files need to be added to the s3 bucket.

Job is called Incremental Recap Records; confirm this and view other details by looking in Alma.

This will replace the cron job:
https://github.com/pulibrary/marc_liberation/blob/6111d339441b6b4d80e63502121aea085fabb416/config/schedule.rb#L26-L29

publishing job notes:
* suppressed records / holdings will be included.
* Only recap records will be included in the dump
* Full dumps can be run as needed, ad hoc, and we don't need to poll forthem. Will never be used for submit collection.

tpendragon:
Closes #1049

### DIFF
--- a/config/alma.yml
+++ b/config/alma.yml
@@ -14,6 +14,9 @@ default: &default
     "Publishing Platform Job Incremental Publishing via FTP":
       dump_type: "CHANGED_RECORDS"
       dump_file_type: "UPDATED_RECORDS"
+    "Publishing Platform Job Incremental ReCAP Records":
+      dump_type: "PRINCETON_RECAP"
+      dump_file_type: "RECAP_RECORDS"
 
 development:
   <<: *default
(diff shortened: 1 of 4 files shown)


hackartisan (commented):
Great this should get the files into the bibdata db.

We also need to put the files in the s3 location. Should we split that out?

tpendragon:
@hackartisan Yes please! Seems like it'd need some different info

hackartisan (changes requested):
@tpendragon I'm also now wondering, with the deletion of that cron job is there other code that can be deleted?

tpendragon:
> @tpendragon I'm also now wondering, with the deletion of that cron job is there other code that can be deleted?

I..don't know. Coverage isn't great, I don't think I'd want to do that kind of cleanup until during or after #1220.

hackartisan:
Fair enough!

PR #1222, opened 2021-04-08
 Add access_facet values

### LINKED ISSUE #1202: Include records with electronic portfolios in online facet
per @mzelesky 

A record should be in the online facet if either of the two below conditions are true:

- If there is an electronic portfolio include it in the online facet.
- If the second indicator of 856 is 1 or 0 or blank include in the online facet. https://www.loc.gov/marc/bibliographic/bd856.html

Do not consider holding locations these are only used for physical items. (The elfs are not valid any more. The access.rb file should not include elfs. https://github.com/pulibrary/bibdata/issues/1203)

Implementation: Use a flag `electronic_record`

eliotjordan:
Closes #1202


tpendragon (dismissed):
This is a great refactor and a great update to the tests which should make it a lot easier to change in the future too. Good work!

I might not take this approval as permission to merge, I don't know enough about the underlying logic to know it's correct.

hackartisan:
Perhaps @mzelesky or @kevinreiss could take a look at this one

marc_to_solr/lib/access_facet_builder.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    def hathi_locations
+      return [] unless context.output_hash['hathi_identifier_s'].present?
+      hathi_line = find_hathi_by_oclc(context.output_hash['oclc_s'].first)
+      parse_locations_from_hathi_line(hathi_line)
+    end
+
+    def hathi_temp
+      return 'Temporary Digital Access' if hathi_locations.include?('hathi_temp')
+    end
```
christinach, line 54: We don't have the 'Temporary Digital Access' facet. It should not appear in the catalog. catalog staging has it because of an old index. @kevinreiss @mzelesky

christinach: I'm not sure if it matters and if there is a reason to keep it. I looked in bibdata-worker2 under `shared/hathi/input and output` and we dont have the hathi file with the 'deny', 'allow' access anymore. Now I start to remember we used the 'Temporary Digital access' facet to index and separate the deny 'hathi' records but we never used it because the hathi access changed. So I think my question is more for @mzelesky and @kevinreiss can we remove this 'Temporary Digital Access' facet from the code?

kevinreiss: Yes, we can remove this. We never finished implementing this feature and even if we are asked to extend beyond the Voyager migration I do not think we would ever try to use this again.

eliotjordan:
@christinach @kevinreiss temp access hathi. Ready for re-review.

marc_to_solr/lib/access_facet_builder.rb
```diff
@@ -0,0 +1,59 @@
+# Class for building access_facet values
+class AccessFacetBuilder
+  # Build access facet
+  # @param record [MARC::Record]
```
christinach, line 4: I like the AccessFacetBuilder addition!

christinach (approved):
It looks good @eliotjordan, it follows https://github.com/pulibrary/bibdata/issues/1202#issue-849268462. Thank you !

Issue #858, opened 2020-10-14
Review error log 


christinach:
Review the error log we generate.

christinach:
- [ ] https://github.com/pulibrary/marc_liberation/blob/0a5d2e750bc07b4284ce3a9de167b83f70efc0c7/marc_to_solr/lib/traject_config.rb#L497 look for errors specifc to missing $a instead of $a $n and $p.

- [ ] What locations (852b) are attached to the bib. It would be useful to know if they are physical items. These are the ones the the catalogers are able to update. (e-resources / scsb they are beyond our control. ) https://github.com/pulibrary/marc_liberation/blob/0a5d2e750bc07b4284ce3a9de167b83f70efc0c7/marc_to_solr/lib/traject_config.rb#L1101

christinach:
see also https://github.com/pulibrary/bibdata/issues/1230

kevinreiss:
I think we've moved past this issue. @christinach @mzelesky let me know if this is still a thing. If it is a suggest we open a new ticket that reflects current desires to improve the error log.

PR #1228, opened 2021-04-09
Remove "Missing holdings" log entry

### LINKED ISSUE #1226: Downgrade "holding not found" log messages to debug
these are not really a concern and there are TONS of them.

hackartisan:
@christinach what do you think about removing this altogether? Do we ever need to look for these entries?

closes #1226

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -1131,11 +1131,7 @@ to_field 'subject_era_facet', marc_era_facet
 
 to_field 'holdings_1display' do |record, accumulator|
   all_holdings = process_holdings(record)
-  if all_holdings == {}
-    logger.error "#{record['001']} - Missing holdings"
-  else
-    accumulator[0] = all_holdings.to_json.to_s
-  end
+  accumulator[0] = all_holdings.to_json.to_s unless all_holdings.empty?
 end
 
 ## for recap notes


marc_to_solr/lib/traject_config.rb
```diff
@@ -1131,11 +1131,7 @@
 
 to_field 'holdings_1display' do |record, accumulator|
   all_holdings = process_holdings(record)
-  if all_holdings == {}
-    logger.error "#{record['001']} - Missing holdings"
-  else
-    accumulator[0] = all_holdings.to_json.to_s
-  end
+  accumulator[0] = all_holdings.to_json.to_s unless all_holdings.empty?
 end
```
christinach, line 1136: I think it is useful to have this info. In the current production catalog there are not a lot but there are some. I reviewed the error log with @mzelesky a while ago and we didnt add a note to remove this from the log. https://github.com/pulibrary/bibdata/issues/858 There is a google document with an older log. I would also ask @mzelesky again.

christinach (changes requested):

marc_to_solr/lib/traject_config.rb
hackartisan: If it's helpful I will keep it but downgrade it to info or debug. but it would be nice to hear a scenario where we've used it. With the alma indexes there are really a ton of these entries.

christinach: I have used the log in the past to review the index and this is why it was added. When we first reviewed the log there was not an alma index to compare the results. I understand that there are a lot of records in alma with no holdings and makes it difficult to review the log and this came up now with the alma qa index. I think we can remove it for now unless there is a different opinion from @mzelesky .

mzelesky: Yes, I think it's ok to remove for now. Eventually, it would be useful to know which bib records have no physical holdings or electronic portfolios, but we can take that as it comes.

christinach (approved):

marc_to_solr/lib/traject_config.rb
christinach: @mzelesky per you request i opened https://github.com/pulibrary/bibdata/issues/1230. Thank you.

Issue #1094, opened 2021-02-18
Add error handling around Alma's concurrent API threshold limit


hackartisan:
When the limit is hit, a 429 status is returned with error code "PER_SECOND_THRESHOLD" (https://developers.exlibrisgroup.com/alma/apis/#threshold)

We should have tests ensuring our endpoints handle this response and return appropriate failures. A 429 of our own?

We probably shouldn't retry automatically; that would likely compound the limit issue.

There are 22 api calls from bibdata to alma listed in https://docs.google.com/spreadsheets/d/1LmwK6s2IK6L98-DY3JIHHohUQcqYYa7qupZIJHv7Pu8/edit#gid=0; some of these likely use the same methods in bibdata.

Some data collection here -- maybe just logs? at some point we're going to want to know how often we're hitting them.

endpoints:
- [x] bibliographic endpoints addressed in https://github.com/pulibrary/bibdata/pull/1225/files
- [x] availability endpoints (see https://github.com/pulibrary/bibdata/pull/1257, https://github.com/pulibrary/bibdata/pull/1253, and https://github.com/pulibrary/bibdata/pull/1243)
- [x] barcode#scsb (see https://github.com/pulibrary/bibdata/pull/1264, and https://github.com/pulibrary/bibdata/pull/1266)
- [x] patron#patron_info (in progress, see https://github.com/pulibrary/bibdata/pull/1281)

The `bibliographic#item_discharge` endpoint will be removed (see https://github.com/pulibrary/bibdata/pull/1283)

hectorcorrea:
@hackartisan Does the [patron#patron_info](https://github.com/pulibrary/bibdata/blob/alma/app/controllers/patron_controller.rb) call Alma?

hackartisan:
@hectorcorrea It's a bit hidden in this memoized method: https://github.com/pulibrary/bibdata/blob/alma/app/controllers/patron_controller.rb#L45

hectorcorrea:
@hackartisan duh! If only there was a way to search for a word inside files, right? :)

Thank you!

PR #1253, opened 2021-04-15
Handles HTTP 429 in get_availability_holding

hectorcorrea:
This is part of issue #1094 and it adds an extra check to handle HTTP 409 errors in `get_availability_holding`

Notice that I had to use a slightly different configuration when calling `Alma::BibItem.find()` because the Alma gem by default [swallows the real exception for this call](https://github.com/tulibraries/alma_rb/blob/main/lib/alma/bib_item_set.rb#L36-L39) and prevents us from handling the HTTP 429 error. By forcing `config.enable_loggable = true` before making the call `Alma::BibItem.find()` we get the real exception and can live happily ever after.


hackartisan (approved):
👍🏻

Issue #1256, opened 2021-04-15
Bibliographic Controller is getting too big


hectorcorrea:
Rubocop is complaining that the Bibliographic Controller has gotten too big.

```
app/controllers/bibliographic_controller.rb:1:1: C: Metrics/ClassLength: Class has too many lines. [251/242]
class BibliographicController < ApplicationController ...
```

 I've disabled the warning for now. This issue is our reminder to address it.

hectorcorrea:
Not applicable after PR #1257

PR #1257, opened 2021-04-15
Added controller tests (request tests) for Availability endpoints

hectorcorrea:
New tests are to make sure we handle Alma's PER_SECOND_THRESHOLD errors. 
This is part of #1094. 

Notice that I disabled a Rubocop warning and created a separate issue (#1256) to address it. It is possible that the code to add this same validation to other controllers will make me move some of the new code that pushed the class over the limit to a shared location (e.g. ApplicationController) and will address the issue.

### DIFF
--- a/app/controllers/bibliographic_controller.rb
+++ b/app/controllers/bibliographic_controller.rb
@@ -320,4 +320,4 @@ class BibliographicController < ApplicationController # rubocop:disable Metrics/
         head :bad_request
       end
     end
-end
+end # rubocop:enable Metrics/ClassLength
--- a/app/controllers/bibliographic_controller.rb
+++ b/app/controllers/bibliographic_controller.rb
@@ -106,9 +106,8 @@ class BibliographicController < ApplicationController # rubocop:disable Metrics/
         render json: solr_doc
       end
     end
-  rescue Alma::PerSecondThresholdError => alma_client_error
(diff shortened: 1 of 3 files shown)


hectorcorrea:
I don't quite understand why coveralls is complaining that coverage decreased. If I read the output correctly the stuff that is not being tested is not new. Thoughts?

hackartisan:
re: coveralls, it's probably just a question of the math from the division when the number of lines changed. It shouldn't block merge.

app/controllers/bibliographic_controller.rb
```diff
@@ -1,4 +1,4 @@
-class BibliographicController < ApplicationController
+class BibliographicController < ApplicationController # rubocop:disable Metrics/ClassLength
```
hackartisan, line 1: I think we have to enable this again at the end of the class

hackartisan: Or add it as an exclusion in .rubocop.yml

hackartisan (changes requested):
This looks good to me, just one question about the rubocop exclusion.

app/controllers/bibliographic_controller.rb
hectorcorrea: Thank you! I did not know it will stay on by default. I've added the enable clause at the end.

hackartisan (approved):

PR #1264, opened 2021-04-16
PER_SECOND_THRESHOLD validation for barcode endpoint

hectorcorrea:
First pass at adding validation for PER_SECOND_THRESHOLD to the barcode endpoint.

I added some helper methods to the ApplicationController that, if the approach looks
good to the team, we can refactor the code in the BibliographicController to use them
as well and make that controller a bit skinnier.

Notice that I also moved some of the code in the BarcodeController to the AlmaAdapter
so that the logic for "Alma stuff" is consolidated in a single place. This also masks
the issue that Alma handles error reporting different from case to case: sometimes it
raises an exception while other times (like in this case) it overloads the return object
with the error information.

Part of #1094


hackartisan (approved):
This looks great I like the plan to move the error handling up to the application controller

PR #1275, opened 2021-04-20
Status and status_label values for Availability by Holding (mfhd) endpoint

### LINKED ISSUE #937: Provide detailed availability values at /bibliographic/:id/holding/:id/availability endpoint
child of #573; follow-up to #1179

## Implementation

For this endpoint, we need to send item-level availability statuses. The status and status_label should be taken from:
1. work_order_type if it exists
2. else process_type if it exists
3. else base_status

`status` field should return "value"; `status_label` field should return "desc"

Note: An item will never have more than one `work_order_type` at once.

An example "at preservation" item:
MMS Id: 99116973763506421
Holding ID: 22245146270006421
Item ID: 23245146260006421

```
"base_status":
  {
    "value": "0",
    "desc": "Item not in place"
  }, 
“process_type”:
  {
    “value”: “WORK_ORDER_DEPARTMENT”,
    “desc”: “In Process”
  },
“work_order_type”:
  {
    “value”: “Pres”,
    “desc”: “At Preservation”
  },
 “work_order_at”:
  {
    “value”: “DEFAULT_CIRC_DESK-Pres”,
    “desc”: “Default Circulation Desk”
  },
```

## Background / Discussion

Previously we computed a wide variety of labels based on statuses that came out of voyager. We won't do that any longer. Below is discussion we had on the way to figuring out the new procedure.

Status labels [currently in marc liberation](https://github.com/pulibrary/marc_liberation/blob/090e90bf02cdbd80564c14bd9e8b01952f4a8a64/app/models/voyager_lookup.rb#L109-L146):
- [ ] 'All Items Available' (Now `more_items == true && status.start_with?('on-site')`)
- [ ] 'Available' (Now `['Not Charged', 'On Shelf']`)
- [ ] 'Long overdue' (Now `['Lost--system applied', 'On-site - lost--system applied']`)
- [ ] 'Lost' (Now `['Lost--library applied', 'On-site - lost--library applied']`)
- [ ] Returned (Now `['Discharged']`)
- [ ] In transit (Now `['In transit discharged']`)
- [ ] In process (Now `['In process', 'On-site - in process']`)
- [ ] Checked out (Now `['Charged', 'Renewed', 'Overdue', 'On hold', 'In transit', 'In transit on hold', 'In transit discharged', 'At bindery',  'Remote storage request', 'Hold request', 'Recall request']`)
- [ ] Missing (Now `['Missing', 'Claims returned', 'Withdrawn', 'On-site - missing', 'On-site - claims returned', 'On-site - withdrawn']`)
- [ ] See front desk (Starts with 'on site - ')
- [ ] On-site access ('on-site')
- [ ] Pending order ('Pending Order')
- [ ] Order received ('order received')
- [ ] On-order ('on-order') - see #630 for alma details
- [ ] ? (There's a fallback to whatever Voyager has if the above doesn't fit)

### Alma implementation

There is a list of process types at https://developers.exlibrisgroup.com/alma/apis/docs/xsd/rest_item.xsd/. The list shows demo process types but is supposed to show your institution's actual process types if you log in. The only type of process_type our records currently have is "technical". Everything with one of those statuses will have availability "not available".

The records are migrated with process_type "technical - migration" and the voyager data in local_notes_3. pul staff will manually move them into the various other process types in batches. This means they will create a "work order" for each process type and then create statuses for that work order.

We should be able to find fixtures for items with the checked out status. These will have a process_type "loan".

We can also create a fixture for an item with a process type of "technical - migration". The next step will be converting a handful of these to use real work orders / process types.

#### Impact on Requests app

3 Broad categories of status values (not status label values!), as far as Requests is concerned. Behavior culminates in the [requestable form](https://github.com/pulibrary/requests/blob/main/app/views/requests/request/_requestable_form.html.erb).
- Charged or missing (`borrow_direct?`)
  - user is prompted to check borrow direct
- On-order or Pending (`request?`)
  - In-process and on-order items are sent through an email workflow so catalogers can bump up its priority. 
- Available (requests logic is that it's available if it's not in the other categories) (`pick_up?`)
  - user can schedule a pickup.

related migration ticket in requests: https://github.com/pulibrary/requests/issues/891

hectorcorrea:
First pass at picking the correct status text from the work order, process type, and base status.

Fixes #937 

This is still a work in progress.

### DIFF
--- a/app/adapters/alma_adapter/alma_item.rb
+++ b/app/adapters/alma_adapter/alma_item.rb
@@ -167,18 +167,20 @@ class AlmaAdapter
     end
 
     def availability_summary(marc_holding:)
+      status = calculate_status()
+
       {
         barcode: item_data["barcode"],
         id: item_data["pid"],
         copy_number: holding_data["copy_id"],
-        status: nil, # ?? "Not Charged"
+        status: status[:code],        # Available or ?? "Not Charged"
+        status_label: status[:label], # available
         on_reserve: "N",
         item_type: item_type, # e.g., Gen
         pickup_location_id: holding_location, # stacks
         pickup_location_code: holding_location, # stacks
         location: composite_location, # firestone$stacks
         label: holding_library_name, # Firestore Library
-        status_label: status_label(marc_holding), # available
         description: item_data["description"], # "v. 537, no. 7618 (2016 Sept. 1)" - new in Alma
         enum_display: enumeration, # in Alma there are many enumerations
         chron_display: chronology # in Alma there are many chronologies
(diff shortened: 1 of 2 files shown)


app/adapters/alma_adapter/alma_item.rb
```diff
@@ -203,26 +206,50 @@ def item_type
...
+             end
+      { code: code, label: desc, source: "work_order" }
+    end
+
+    def status_from_process_type
+      value = item_data.dig("process_type", "value")
+      desc = item_data.dig("process_type", "desc")
+
+      # Source for values: https://developers.exlibrisgroup.com/alma/apis/docs/xsd/rest_item.xsd/
+      code = if value == "ACQ"
```
kevinreiss, line 235: Knowing the value of the process type would be useful. Currently we have "special" behaviors for titles that are "On Order" which is what the process type "ACQ" would map to.

christinach (dismissed):
It looks great. I'm not sure if you would like to add the mapping that @kevinreiss commented about https://github.com/pulibrary/bibdata/pull/1275/files#r617913506 or if you want to add this in a different pr.

app/adapters/alma_adapter/alma_item.rb
hectorcorrea: Good to know. I've added a `process_type` property to the response.

hectorcorrea:
@christinach I've added the property that Kevin requested.

christinach (approved):

PR #1281, opened 2021-04-22
Implements PER_SECOND_TRESHOLD check for patron endpoint.

hectorcorrea:
Implements PER_SECOND_TRESHOLD check for patron endpoint.

Moved some of the logic for the Alma code to the adapter to hide the logic to deal with Alma errors since the gem sometimes hides the errors, sometimes lets them bubble up, and some other times buries the real error alongside a generic error. 

Notice that method `alma_preserve_exception` that is now duplicated in both AlmaAdapter and AvailabilityStatus. We could refactor this to a single method in a helper class or update the AvailabilityStatus to funnel all calls to the Alma gem through the adapter but that would couple both classes both ways or remove the Alma calls from AvailabilityStatus. My suggestion would be to refactor AvailabilityStatus so that it does not need to call the Alma gem and instead it receives the data that it needs, but that should be handled via a separate issue.

Part of #1094

### DIFF
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -139,6 +139,17 @@ class AlmaAdapter
     holding
   end
 
+  def find_user(patron_id)
+    alma_preserve_exception do
+      return Alma::User.find(patron_id)
+    rescue Alma::StandardError => e
+      # The Alma gem throws "not found" for all errors but only error code 401861
+      # really represents a record not found.
+      raise Alma::NotFoundError, "User #{patron_id} was not found" if e.message.include?('"errorCode":"401861"')
+      handle_alma_error(client_error: e)
+    end
+  end
+
   private
 
     def apikey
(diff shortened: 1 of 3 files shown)


app/adapters/alma_adapter.rb
```diff
@@ -139,6 +139,17 @@ def holding_by_id(mms_id:, holding_id:)
     holding
   end
 
+  def find_user(patron_id)
+    alma_preserve_exception do
+      return Alma::User.find(patron_id)
+    rescue Alma::StandardError => e
+      # The Alma gem throws "not found" for all errors but only error code 401861
+      # really represents a record not found.
```
hackartisan, line 147: 😂

hackartisan (approved):
This looks great. Thank you for all the detailed attention to the various ways these errors come through.

hackartisan:
Created https://github.com/pulibrary/bibdata/issues/1285 as a followup

PR #1283, opened 2021-04-22
Adds check for PER_SECOND_THRESHOLD to item_discharge endpoint

hectorcorrea:
Moves the calls to Alma to the adapter and adds a check for PER_SECOND_THRESHOLD errors.

Related to #1094


hectorcorrea:
I am closing this PR since we won't be needing the item_discharge endpoint anymore but I wanted to keep the work that would be needed to add the check if the end up keeping (or resuscitating) the endpoint.

Issue #1190, opened 2021-03-31
Add `temp_loc` to bibliographic/[:id]/availability.json endpoint


eliotjordan:
We need `temp_loc` for stackmap and for logic to display the request button in OL.

https://github.com/pulibrary/bibdata/blob/alma/app/adapters/alma_adapter/availability_status.rb#L29

hectorcorrea:
@eliotjordan what's the value that goes in `temp_loc` ? true/false or an actual location?

hackartisan:
orangelight is expecting an actual location.

mzelesky:
Our temp locations are not currently included in our publishing profiles. This needs to be corrected before temp locations can be accessed. To fix this, we need to assign 'Current library subfield' and 'Current location subfield.' If those values are different than the locations in the 852 field, then they are temp values.

tpendragon:
Blocked by current work on availability.

hectorcorrea:
This has been unblocked (https://github.com/pulibrary/bibdata/pull/1211)

hectorcorrea:
Per Mark: the following records are in temporary locations and are not ETAS: 9959958323506421 and 9938793603506421

PR #1299, opened 2021-05-04
Availability endpoint and temp locations

hectorcorrea:
@hackartisan @eliotjordan This PR makes some changes to the Availability endpoint to handle holdings in temporary locations.

Notice that I changed a few keys/values of the attributes returned in the API (i.e. it will break Orangelight) but I think these keys/values make more sense now that we understand what Alma does.

I am also adding a boolean `temp_location` that we can use in Orangelight to add something in the web page to request more Availability information for these records (see [notes on 4/29](https://docs.google.com/document/d/1sq_DmkRVLWEQdwf0YY1Dl3KGYlPd_0ldCYYc6PTcOuY/edit) about this)

Addresses #1190 

## Example of usage:
A call like this: http://localhost:3000/bibliographic/9959958323506421/availability.json will return a flag indicating whether a holding is in a temporary location. When that is the case the id for the holding will be a fake one:

```
{
    "9959958323506421": {
        "fake_id_1": {
            "on_reserve": "N",
            "location": "lewis$resterm",
            "label": "Lewis Library: Term Loan Reserves",
            "status_label": "Available",
            "copy_number": null,
            "cdl": false,
            "temp_location": true,
            "id": "fake_id_1"
        }
    }
}
```

When the client detects that an item is in a temporary location it can issue a separate call like this http://localhost:3000/bibliographic/9959958323506421/availability.json?deep=true (notice the new `deep=true`) parameter and then we will fetch the actual holding id(s) for those items in temporary locations:

```
{
    "9959958323506421": {
        "22272063570006421": {
            "on_reserve": "N",
            "location": "lewis$resterm",
            "label": "sciresp: Lewis: Term Loan",
            "status_label": "Available",
            "copy_number": "1",
            "cdl": false,
            "temp_location": true,
            "id": "22272063570006421"
        },
        "22272063520006421": {
            "on_reserve": "N",
            "location": "lewis$resterm",
            "label": "sciresp: Lewis: Term Loan",
            "status_label": "Available",
            "copy_number": "2",
            "cdl": false,
            "temp_location": true,
            "id": "22272063520006421"
        }
    }
}
```

Notice that is it possible to get more than ONE holding for these records since at this point Alma takes into account "multiple copies" of the same item.


spec/adapters/alma_adapter_spec.rb
```diff
@@ -311,33 +313,38 @@
...
-      expect(portfolio[:status_label]).to eq "Available"
+      empty_availability = { "99122426947506421" => {} }
+      expect(availability).to eq empty_availability
     end
 
-    it "reports availability for other" do
-      # This kind of resource is new in Alma and still needs some work,
-      # but for now we at least test that we are executing the branch of
-      # code that handles them.
+    it "reports availability (without holding_id) for items in temporary locations" do
```
hectorcorrea, line 332: This test and the one below it are the only relevant test for handling temporary locations. The rest of the changes in the tests are to account for the slightly different response that we returning now that we are only reporting on physical items (i.e. ignoring electronic portfolios and digital items)

app/adapters/alma_adapter.rb
```diff
@@ -36,10 +36,22 @@ def get_availability(ids:)
...
-  def get_availability_one(id:)
+  def get_availability_one(id:, deep_check: false)
     bibs = Alma::Bib.find(Array.wrap(id), expand: ["p_avail", "e_avail", "d_avail", "requests"].join(",")).each
     return nil if bibs.count == 0
-    { bibs.first.id => AvailabilityStatus.new(bib: bibs.first).bib_availability }
+
+    av_info = AvailabilityStatus.new(bib: bibs.first).bib_availability
+
+    temp_locations = av_info.any? { |_key, value| value[:temp_location] }
+    if temp_locations && deep_check
```
hectorcorrea, line 46: This is the new behavior that the client can request via the deep_check parameter to drill down for more information if the initial call reports holdings in temporary locations.

app/adapters/alma_adapter/availability_status.rb
```diff
@@ -12,61 +12,71 @@ def initialize(bib:)
...
-    #   "copy_number": item["copy_number"],
-    #   "patron_group_charged": nil,    # TODO: nil
-    #   "status": "???",                # TODO: "Not Charged"
-    # TODO: Confirm that we need all of these values with the Alma implementation.
-    # TODO: Find out if we can get copy_number from Solr (i.e. without hitting ExLibris' API)
-    def holding_status(holding)
+    # Returns availability information for each of the holdings in the Bib record.
+    # Notice that although we return the information by holding, we drill into item
+    # information to get details.
+    def bib_availability_from_items
```
hectorcorrea, line 27: This method is brand new: `item_data` makes a call to the Alma API to fetch item-level information that we use when the client asks for a "deep_check".

app/adapters/alma_adapter/availability_status.rb
```diff
@@ -12,61 +12,71 @@ def initialize(bib:)
...
+        raise StandardError.new, "Multiple items found under the same key" if value.count > 1
+        alma_item = AlmaAdapter::AlmaItem.new(Alma::BibItem.new(value.first.item))
+        status = holding_status_from_item(alma_item)
+        raise StandardError.new, "Holding found more than once" if availability[alma_item.holding_id]
+        availability[alma_item.holding_id] = status
+      end
+      availability
+    end
+
+    def holding_status(holding:, sequence:)
```
hectorcorrea, line 40: This method is now greatly simplified because we are ONLY handling physical inventory and ignoring electronic and digital.

app/adapters/alma_adapter/availability_status.rb
```diff
@@ -12,61 +12,71 @@ def initialize(bib:)
...
+      status[:id] = "fake_id_#{sequence}" if holding["holding_id"].nil?
+
+      # Notice that we only check if a holding is available via CDL when necessary
+      # because it requires an extra (slow-ish) API call.
+      status[:cdl] = cdl_holding?(holding["holding_id"]) if status[:status_label] == "Unavailable"
+
+      status
+    end
+
+    def holding_status_from_item(alma_item)
```
hectorcorrea, line 69: This method is new. 

This method builds a response that is identical to what the method `holding_status` (just above this one) returns.

This new method builds the response from the data coming from Alma's `/bib/{id}/holdings/all/items` API as opposed to the information in the AVA fields in the Alma's `/bib/{id}` endpoint.

hackartisan (changes requested):
Just one small request

app/adapters/alma_adapter/availability_status.rb
```diff
@@ -12,61 +12,71 @@ def initialize(bib:)
...
```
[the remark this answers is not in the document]
hectorcorrea: Good idea. Done.

hackartisan (approved):

PR #1302, opened 2021-05-04
Update locations gem with remote_storage attribute

### LINKED ISSUE #1301: Update holding locations using the new remote_storage attribute from the locations gem
following issue from https://github.com/pulibrary/locations/pull/88

christinach:
closes #1301 

First https://github.com/pulibrary/locations/pull/88 needs to be ready and merged, create a tag in locations and update this PR to point to the tag. This PR is currently pointing to the `87-add-remote-storage` branch.


christinach:
This is ready for review. I noticed that the remote storage is not added in the edit view but I don't think it is needed since we said that moving forward we are not going to do any edits on the holding locations table using the UI.

christinach:
PR is updated.

hackartisan (approved):

Issue #1255, opened 2021-04-15
SubmitCollection to ReCAP needs to support bound-withs.


tpendragon:
Sibling of #1251

We need more information here, I'm going to fill in my understanding and ask @mzelesky edits this if I get anything wrong. This functionality is already implemented for BarcodesController.

When a host record appears in the incremental dump:

1. Pull the updated constituent records.
2. Copy all holding/item data from the host record over to the constituent records
3. Send the host records as well as the constituents.

When a constituent record appears in the incremental dump:

1. Pull the host record (there may not be a link back from constituent to host - this is something that will be added in the future, maybe we can count on it existing?)
2. Copy all holding/item data from the host record to constituent records
3. Send the host records as well as the constituents.

The architecture here needs to try to avoid doing a ton of single-bib API calls. We may be able to do something like pre-scan for all constituents/hosts we're going to need to query for and grab them in bulk API calls.

The records we have set up correctly are listed on https://github.com/pulibrary/orangelight/issues/2431

hackartisan:
@mzelesky Will these dumps included deleted records? I assume they will. Do we need to enrich deleted boundwiths to add item / holding data from the host record?

hackartisan:
@mzelesky If the incremental dump doesn't include one of the records for the boundwith, we'll have to pull that from the API to include it in the file. However, the SCSB records must include suppressed records. The API doesn't return suppressed records. I'm not sure there's a way to fulfill this requirement for that edge case.

hackartisan:
@mzelesky we're considering whether it might work to create a nightly dump of all boundwith host and constituent records to reference as we create the scsb file. We messed around with sets a little, though, and couldn't figure out how to generate a dump like this. Do you think it is possible?

christinach:
I haven't tried it but have you looked into https://knowledge.exlibrisgroup.com/Alma/Knowledge_Articles/Create_Users_Set_for_expired_patrons_based_on_Analytics? but this would be a one time set.

hackartisan:
We think we will actually get suppressed bib records from the bibliographic endpoint, and suppressed holdings from the holdings/ALL/items endpoint. But we think you wouldn't get an AVA field from the bibliographic endpoint for a holdings record that is suppressed. We have not tested these cases. We aren't sure there exist suppressed holdings.

Implementation plan:
- Sweep through the dump files to collect all the host and constituent records
- Insert them in a database cache for marc records, and add them to a pair of in-memory arrays (hosts, constituents)
[...]

PR #1303, opened 2021-05-06
Process SCSB Boundwith Records

eliotjordan:
- Adds CachedMarcRecord model
- Extracts boundwith records from SCSB dumpfiles
- Processes and saves records in a new scsb dumpfile

Progress towards #1255


app/jobs/recap_dump_file_processing_job.rb
```diff
@@ -22,7 +22,20 @@ def extract_records(path)
         records = MARC::XMLReader.new(content, external_encoding: 'UTF-8').to_a.map do |record|
           # ScsbDumpRecord will handle converting from the dump's MARC-XML to
           # the proper format for submitCollection
-          AlmaAdapter::ScsbDumpRecord.new(marc_record: record)
+          scsb_record = AlmaAdapter::ScsbDumpRecord.new(marc_record: record)
+          # If boundwith param is true; return record if it is a boundwith or
+          # skip record if it is not a boundwith.
+          # If boundwith param is false; return record if it is not a boundwith
+          # or skip record if it is a boundwith.
+          if boundwiths && scsb_record.boundwith?
```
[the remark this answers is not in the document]
hackartisan: yesss that looks so much nicer

app/adapters/alma_adapter/scsb_dump_record.rb
```diff
@@ -17,18 +17,124 @@ def transformed_record
...
+    end
+
+    def cache
+      CachedMarcRecord.find_or_create_by(bib_id: id).tap do |record|
+        record.marc = marc_record.to_xml
+        record.save
+      end
+    end
+
+    # 852/866/867/868 fields which have a subfield "8" are all copied from
```
hectorcorrea, line 95: Perhaps indicate that the subfield "8" represents the holding_id ?

app/jobs/recap_boundwiths_processing_job.rb
```diff
@@ -0,0 +1,97 @@
...
+      end.flatten.compact
+    end
+
+    # @return [DumpFileType]
+    def dump_file_type
+      DumpFileType.find_by(constant: "RECAP_RECORDS")
+    end
+
+    # Find related host and constiuent records and add
+    # to set of boundwith records.
```
christinach, line 56: I'm sorry a small spelling mistake. constiuent

app/jobs/recap_boundwiths_processing_job.rb
```diff
@@ -0,0 +1,97 @@
...
+        r.constituent? ? r.host_id : r.id
+      end
+
+      # Iterate through each key in grouped_records
+      grouped_records.each do |host_id, records|
+        host_record = records.find(&:host?)
+        constituent_records = records.find_all(&:constituent?)
+
+        unless host_record
+          # Get host record id from a constituent[773] and
```
christinach, line 70: can we specify 773w ?

app/jobs/recap_boundwiths_processing_job.rb
```diff
@@ -0,0 +1,97 @@
...
+          # retrieve from Alma or cache
+          host_record = constituent_records.first.host_record
+          # Add missing host record to group
+          grouped_records[host_id] << host_record
+        end
+
+        # Fetch constituent record ids from host[774] and retrieve from Alma or
```
christinach, line 77: can we specify in the comment 774w?

christinach (changes requested):
It looks great. Some comments updates.

hackartisan (changes requested):
I wanted to record a couple of blocking concerns that have come out of slack discussion:

1. This process as written is not idempotent. That is, if the code errors in the process of creating / saving the boundwith file, the records will no longer be in the other files in order to rerun it. Such an error seems likely to happen at some point, given that we're doing network stuff and working with marc records.
2. There is a race condition in the way it's written. With each dump file being stripped of its boundwith records at the same time as they are scanned for boundwith records, we will undoubtedly loose some.

We need to make adjustments to address these concerns.

hackartisan (changes requested):
Also we should add a test for the case where there were no boundwiths in the dump

hackartisan (changes requested):
Also we should add a test for the case where there were no boundwiths in the dump

hackartisan (approved):
They're all temp files now. Should get around both the race condition and make it idempotent. I can't think of a use case that requires us to store them, as long as they get to the s3 drive. Seems great! 🎉

christinach (approved):

Issue #1265, opened 2021-04-16
bibdata timeouts returning 400


hackartisan:
There are examples in the logs of bibdata getting a Net::ReadTimeout and returning a 400.

- [x] Send I think a 504 here
- [x] but also increase our timeout value for Alma calls made from the multiple availability endpoint
- [x] Also increase timeouts for alma calls to the /holdings/ALL/items endpoint. These can take 7-8 seconds to load the first page.

example log entry:
```
E, [2021-04-16T14:24:01.117195 #17784] ERROR -- : HTTP 400. Failed to retrieve availability for IDs: ["9918888023506421", "9929723813506421", "998968313506421", "9911466953506421", "99116916533506421", "99647623506421", "9988274863506421", "9937714773506421", "992643323506421", "9961661563506421", "9958479363506421", "99620463506421", "9974132093506421", "99115371133506421", "9970871483506421", "9919068383506421"] Net::ReadTimeout
I, [2021-04-16T14:24:01.117750 #17784]  INFO -- : {"method":"GET","path":"/bibliographic/availability.json","format":"json","controller":"BibliographicController","action":"availability_many","status":400,"duration":10246.28,"view":0.0,"ddsource":["ruby"],"params":{"bib_ids":"9918888023506421,9929723813506421,998968313506421,9911466953506421,99116916533506421,99647623506421,9988274863506421,9937714773506421,992643323506421,9961661563506421,9958479363506421,99620463506421,9974132093506421,99115371133506421,9970871483506421,9919068383506421","format":"json"},"@timestamp":"2021-04-16T14:24:01.117Z","@version":"1","message":"[400] GET /bibliographic/availability.json (BibliographicController#availability_many)"}
```

PR #1312, opened 2021-05-11
Uses a longer timeout value 

hectorcorrea:
Uses a longer timeout value (10 seconds) for calls to the Alma Items API

(partially) addresses #1265


hackartisan (approved):
👍🏻  logging the elapsed time is a really good idea

PR #1338, opened 2021-05-19
Require rubygems/package to fix cache_file rake task

eliotjordan:
This is needed to run the service in a rake task.


hackartisan (commented):
Weird that the rake task needs this but the spec didn't?

hackartisan (approved):
I see it's required in other places, though. 

👍🏻

Issue #1336, opened 2021-05-19
Incremental recap dump files are not uploaded to S3 


eliotjordan:
Incremental recap dump files are not being uploaded to S3 correctly. In a republishing job with 26 dump files, only one file is uploaded to the S3 bucket.

eliotjordan:
This looks to be a performance & memory issue. On the alma staging machine (5 sidekiq threads), this is the data from `top`:

25 minutes:   99.3% cpu  73.7% memory
40 minutes:   99.3% cpu  90.0% memory

Best guess, the sidekiq jobs saturate the available memory. It's strange and annoying that the jobs are considered "complete" when the threads terminate.

`perf` wasn't that helpful

```
   5.60%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000ae361
   5.11%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000ae358
   3.52%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000ae33f
   2.85%  util.rb:23     libc-2.27.so          [.] 0x00000000000bb4ea
   2.77%  util.rb:23     libc-2.27.so          [.] 0x00000000000bb4e6
   2.77%  util.rb:23     libc-2.27.so          [.] 0x00000000000bb4f2
   2.70%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000b124d
   2.62%  util.rb:23     libc-2.27.so          [.] 0x00000000000bb4ee
   2.27%  util.rb:23     libruby-2.6.so.2.6.6  [.] rb_wb_protected_newobj_of
   2.08%  util.rb:23     libc-2.27.so          [.] cfree
   1.55%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000b1154
   1.52%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000b2488
   1.47%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000b1e14
   1.45%  util.rb:23     libruby-2.6.so.2.6.6  [.] 0x00000000000af172
   1.25%  util.rb:23     libc-2.27.so          [.] malloc
```

tpendragon:
It's not that they're considered complete, it's that the OS terminates the thread to free memory and the job doesn't make it back. Sidekiq-pro fixes that.

eliotjordan:
[...]

PR #1339, opened 2021-05-19
Reimplement RecapTransferJob as RecapTransferService

eliotjordan:
Hopefully this refactor will solve the issues with scsb record uploads, or, at the least, provide some insight as to the job behavior on the staging sidekiq workers.

Work towards #1336


tpendragon:
What happens now if the transfer fails? Will it re-process the SCSB file?

eliotjordan:
@tpendragon The error should cause the dumpfile / boundwiths-file processing jobs to restart and re-proccess the file.

tpendragon:
@eliotjordan And that's safe?

eliotjordan:
@tpendragon The original dump file is never altered. The processed file is saved as a tempfile and sent to S3. That was happening before this PR. If the transfer fails, the only penalty should be the extra time to run the job again. It was the way to get these to be idempotent.

tpendragon (approved):

PR #1341, opened 2021-05-20
Improve ReCAP guard in AlmaItem methods

### LINKED ISSUE #1277: Update guard for recap location on item enrichment
This line will break when many recap items are switched to locations under their home libraries to use the remote storage feature of alma:

https://github.com/pulibrary/bibdata/blob/41dc462cd2a8d1d1c05daa1b2d13d43bde35f1c0/app/adapters/alma_adapter/alma_item.rb#L51

Keep the guard but use the location code not the library. Location code should be a member of one of these arrays: https://github.com/pulibrary/bibdata/blob/41dc462cd2a8d1d1c05daa1b2d13d43bde35f1c0/app/adapters/alma_adapter/alma_item.rb#L145-L155

eliotjordan:
Updated recap guards in AlmaItem.

Closes #1277

### DIFF
--- a/app/adapters/alma_adapter/alma_item.rb
+++ b/app/adapters/alma_adapter/alma_item.rb
@@ -64,7 +64,7 @@ class AlmaAdapter
     end
 
     def recap_876_fields
-      return [] unless item.library == "recap"
+      return [] unless recap_item?
       [
         MARC::Subfield.new('h', recap_use_restriction),
         MARC::Subfield.new('x', group_designation),
@@ -135,13 +135,13 @@ class AlmaAdapter
     end
 
     def recap_customer_code
-      return unless item.library == "recap"
+      return unless recap_item?
       return "PG" if item.location[0].casecmp("x").zero?
       item.location.upcase
     end
 
     def recap_use_restriction
-      return unless item.library == "recap"
+      return unless recap_item?
       case item.location
       when *in_library_recap_groups
         "In Library Use"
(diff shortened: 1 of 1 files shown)


app/adapters/alma_adapter/alma_item.rb
```diff
@@ -151,7 +151,7 @@ def recap_use_restriction
     end
 
     def group_designation
-      return unless item.library == "recap"
+      return unless recap_item?
       case item.location
       when 'pa', 'gp', 'qk', 'pf'
```
eliotjordan, line 156: Are these still recap locations? If so they should be added to the "all_recap_groups" method.

app/adapters/alma_adapter/alma_item.rb
```diff
@@ -135,13 +135,13 @@ def cdl?
     end
 
     def recap_customer_code
-      return unless item.library == "recap"
+      return unless recap_item?
       return "PG" if item.location[0].casecmp("x").zero?
```
eliotjordan, line 139: Are there recap locations that still start with "x"? If so, how many and why?

app/adapters/alma_adapter/alma_item.rb
christinach: @eliotjordan they are look https://bibdata-alma-staging.princeton.edu/locations/holding_locations but I think we should adjust the location name to be recap$pa, recap$gp etc

christinach: https://api-na.hosted.exlibrisgroup.com/almaws/v1/conf/libraries/recap/locations?apikey=apikey&format=json

eliotjordan: You mean all locations? 

```
 d['pj', 'pk', 'pl', 'pm', 'pn', 'pt']  
["pb", "ph", "ps", "pw", "pz", "xc", "xg", "xm", "xn", "xp", "xr", "xw", "xx"]  
['jq', 'pe', 'pg', 'ph', 'pq', 'qb', 'ql', 'qv', 'qx'] 
```

These should all be prefixed by "recap$"?

kevinreiss: Some will be prefaced by recap$, those are the ones that will still remain a part of the "Recap" library in Alma, but those are moving to be filed under a branch library but flagged for "remote_storage"  an example of this would be "pk" which will come through as "mendel$pk" under the new remote_storage set-up.

kevinreiss: For now I think the way to handle this is to prefix all the locations in these arrays with "recap$". I'll give you a list of the new "remote_storage" at the branch library locations.

hackartisan: Are you sure? I thought the point of doing it this way was so that we would catch things that don't have a recap location because of the switch to remote storage.

hackartisan (approved):

PR #1360, opened 2021-05-25
Move dump type generation into seeder class

### LINKED ISSUE #1352: load dump types on deploy
Duplicate logic from seeds.rb into a rake task. (Don't remove it because it's used in the tests)

Run the rake task on deploy.

eliotjordan:
Move dump type generation into seeder class and load dump types on deploy.

Closes #1352


eliotjordan:
@hackartisan Moved.

spec/utils/data_seeder_spec.rb
```diff
@@ -1,16 +1,20 @@
...
     expect(DumpType.count).not_to be_zero
     expect(DumpFileType.count).not_to be_zero
   end
 
   it 'is idempotent' do
-    Rails.application.load_seed
+    seeder = described_class.new
+    seeder.generate_dump_types
+    seeder.generate_dump_file_types
     expect { Rails.application.load_seed }.not_to change { DumpType.count }
```
hackartisan, line 18: Could we try this with `expect {seeder.generate_dump_types}.not_to`? I think it would be more specific to the class now under test.
Also for the line below.

hackartisan (changes requested):

spec/utils/data_seeder_spec.rb
eliotjordan: Thanks. Missed that. Tired at the end of the day.

hackartisan (approved):
🎉

PR #1383, opened 2021-06-03
Allow scsb full dumps to index

hackartisan:
refs #1382

### DIFF
--- a/docs/indexing.md
+++ b/docs/indexing.md
@@ -57,3 +57,18 @@ as deploy user, in `/opt/marc_liberaton/current`
 $ RAILS_ENV=production UPDATE_LOCATIONS=false SET_URL=http://lib-solr8-staging.princeton.edu:8983/solr/catalog-alma-staging-reindex bin/rake liberate:full --silent >> /tmp/full_reindex_[YYYY-MM-DD].log 2>&1
 
 This step takes about 21 hours
+
+### Index Partner SCSB records
+
+If needed, pull the most recent SCSB full dump records into a dump file:
+
+- as deploy user, in `/opt/marc_liberaton/current`
+- `$ RAILS_ENV=production bundle exec rake scsb:import:full`
+- It kicks off an import job
+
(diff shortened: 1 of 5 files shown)


spec/models/dump_file_spec.rb
```diff
@@ -28,11 +28,16 @@
...
+    let(:recap_record_full_type) { DumpFileType.find_by(constant: 'RECAP_RECORDS_FULL') }
     let(:other_type) { DumpFileType.create }
     it 'returns true for RECAP_RECORDS type' do
       df = DumpFile.new(dump_file_type: recap_record_type)
       expect(df.recap_record_type?).to eq true
     end
+    it 'returns true for RECAP_RECORDS_FULL type' do
+      df = DumpFile.new(dump_file_type: recap_record_full_type)
+      expect(df.recap_record_type?).to eq true
+    end
```
christinach, line 40: should this be `expect(df.recap_record_full_type?).to eq true` ?

docs/indexing.md
```diff
@@ -57,3 +57,18 @@ as deploy user, in `/opt/marc_liberaton/current`
...
 
 This step takes about 21 hours
+
+### Index Partner SCSB records
+
+If needed, pull the most recent SCSB full dump records into a dump file:
+
+- as deploy user, in `/opt/marc_liberaton/current`
+- `$ RAILS_ENV=production bundle exec rake scsb:import:full
+- It kicks off an import job
```
christinach, line 67: can you add a ` at the end of line $ RAILS_ENV=production bundle exec rake scsb:import:full

christinach (changes requested):

spec/models/dump_file_spec.rb
hackartisan: That's the name of the method; I extended its use to cover the new recap dump type as well.

docs/indexing.md
hackartisan: yes thanks for catching this -- done.

christinach (approved):

PR #1387, opened 2021-06-03
Adds logic to calculate on_reserve

### LINKED ISSUE #1363: Set on_reserve flag to correct value
I did a search in alma for `where (In temporary location equals "Yes")` and found this item that is in a reserve location - https://catalog-alma-qa.princeton.edu/catalog/99106282523506421

I then looked that item up in bibdata:
https://bibdata-alma-staging.princeton.edu/bibliographic/99106282523506421/holdings/22158920250006421/availability.json

You can see the on_reserve in the metadata is `N`, not `Y`

There are a few use cases here:
- requests: on-reserve items cannot be requested
- show page: on-reserve items should not have a request button

Additional investigation is needed for both:
- For requests we should be able to use the info we get from the item-level API.
- For the show page we likely want to use the location code, either by adding a new flag to the locations table or by doing an API lookup to Alma's location endpoint. Either way we prefer to hold the logic in bibdata, not requests or orangelight. 

The alma locations API returns this information in the “fulfillment_unit” key. Value would be “reserves”. Mark will follow up to ensure the list of reserves locations is correct.

hectorcorrea:
Adds logic to calculate `on_reserve` based on the library/location rather than returning a hard-coded "N".

Closes #1363

### DIFF
--- a/app/adapters/alma_adapter/alma_item.rb
+++ b/app/adapters/alma_adapter/alma_item.rb
@@ -9,7 +9,7 @@ class AlmaAdapter
 
     def self.reserve_location?(library_code, location_code)
       return false if library_code.nil? || location_code.nil?
-      Rails.cache.fetch("library_#{library_code}_#{location_code}", expires_in: 30.minute) do
+      Rails.cache.fetch("library_#{library_code}_#{location_code}", expires_in: 30.minutes) do
         record = Alma::Location.find(library_code: library_code, location_code: location_code)
         record.response.dig("fulfillment_unit", "value") == "Reserves"
       end
--- a/app/adapters/alma_adapter/alma_item.rb
+++ b/app/adapters/alma_adapter/alma_item.rb
@@ -10,6 +10,7 @@ class AlmaAdapter
     def self.reserve_location?(library_code, location_code)
       return false if library_code.nil? || location_code.nil?
       Rails.cache.fetch("library_#{library_code}_#{location_code}", expires_in: 30.minutes) do
+        # We could get this information from our location table if we want to avoid the Alma API call.
         record = Alma::Location.find(library_code: library_code, location_code: location_code)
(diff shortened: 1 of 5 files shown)


app/adapters/alma_adapter/alma_item.rb
```diff
@@ -7,6 +7,14 @@ def initialize(item)
       super(item)
     end
 
+    def self.reserve_location?(library_code, location_code)
+      return false if library_code.nil? || location_code.nil?
+      Rails.cache.fetch("library_#{library_code}_#{location_code}", expires_in: 30.minutes) do
```
hectorcorrea, line 12: Notice that I am caching this Alma value for 30 minutes since we want to reduce our calls to Alma and these values don't change often.

hectorcorrea: Also, notice that there are two "reserve" methods in this class, one is a class method and the other is an instance method. This is because we need this same logic from outside this class (i.e. in `alma_adapter/availability_status.rb`)

app/adapters/alma_adapter/alma_item.rb
```diff
@@ -36,6 +44,10 @@ def composite_location_label_display
       end
     end
 
+    def in_reserve?
```
hackartisan, line 47: Maybe we should call it `on_reserve?` I think this is more conventional and matches the property name.

hackartisan (changes requested):
Just one comment and otherwise this looks good to me!

I think another option would be to consult our own location tables instead of the alma API, but I'm happy to merge it this way and if it seems like we want to change it later we can.

hectorcorrea:
@hackartisan I renamed the method and added a comment to indicate that we could also get the value from the table in the future.

hackartisan (approved):

Issue #1381, opened 2021-06-02
On order items are being marked as available


carolyncole:
See the very last item in this list: https://bibdata-alma-staging.princeton.edu/bibliographic/9939075533506421/holdings/22194161030006421/availability.json

It is an item that is on order.  I do not believe its status should be Available.  Is there a reason it is marked as available?

kevinreiss:
Also the case for "In Process" items https://bibdata-alma-staging.princeton.edu/bibliographic/99122455086806421/holdings/22477860740006421/availability.json

In Voyager "On-Order" = Alma "Acquisition"
In Voyager "In Process" = Alma "Acquisitions and Cataloging"

PR #1401, opened 2021-06-08
Ensuring that the SolrDeleter handles cases in which a HTTP network error is encountered when requesting the deleting of a batch of IDs

### LINKED ISSUE #1389: Index deletions failing from timeout errors
I ran the full scsb load and 2 incremental alma dumps on bibdata-alma-staging1. On every deletion there was a timeout error. It's very consistent.

Here's a backtrace
https://gist.github.com/hackartisan/bf7c906eb3fe871e6e8d09f3184e42b5

I ran them through the solr url that goes through the load balancer: "http://lib-solr8-staging.princeton.edu:8983/solr/catalog-alma-staging"

We need to replicate this locally and trace down what's causing the deletions to fail while every other index operation succeeds.

jrgriffiniii:
Advances #1389

### DIFF
--- a/marc_to_solr/lib/solr_deleter.rb
+++ b/marc_to_solr/lib/solr_deleter.rb
@@ -41,15 +41,23 @@ class SolrDeleter
       output.join
     end
 
+    def valid_response?(response)
+      return false if response.nil?
+
+      response.status == 200
+    end
+
     def delete_batch(batch)
       uri = "#{@solr_url}/update?commit=true&wt=json"
       body = build_request_body(ids: batch)
 
       response = request_deletion(uri: uri, body: body)
-      return unless response.nil? || response.status != 200
+      return if valid_response?(response)
 
       # Only retry once
       retry_response = request_deletion(uri: uri, body: body)
-      Honeybadger.notify("Error deleting Solr documents. IDs: #{batch.join(', ')}. Status: #{retry_response.status}. Body: #{retry_response.body}") unless retry_response.nil? || retry_response.status == 200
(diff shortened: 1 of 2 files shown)


marc_to_solr/lib/solr_deleter.rb
```diff
@@ -44,10 +46,10 @@ def delete_batch(batch)
       body = build_request_body(ids: batch)
 
       response = request_deletion(uri: uri, body: body)
-      return if response.status == 200
+      return unless response.nil? || response.status != 200
```
hackartisan, line 49: maybe move this logic to a little helper method since it's used here and below. It would help readability to expand it to a full multi-line conditional.

hackartisan (changes requested):
Please see one change request inline.

I think this is a good step toward ensuring the indexing continues on failed deletes.

But when indexing both scsb and alma record sets deletions failed consistently while regular indexing did not. I don't think we're ready to close #1381. I'll update it to say we need more investigation / replication.

marc_to_solr/lib/solr_deleter.rb
christinach: @hackartisan I think in your comment you meant to type: we are not ready to close issue #1389

christinach:
@hackartisan I think in your comment you meant to type: we are not ready to close issue #1389 (I didn't want to edit your comment)

hackartisan:
Yes. thanks @christinach; I fixed that typo.

christinach (approved):

PR #1405, opened 2021-06-09
1403 holding location label

### LINKED ISSUE #1403: holding location label should take value only from the Alma external name
We want to update the LocationDataService so that the holding location `label` takes the value from the Alma external name only. If there is no external name then keep it blank.

https://github.com/pulibrary/bibdata/blob/52e7e865f488f4db56776b811857bc8f548317fb/app/services/location_data_service.rb#L36

christinach:
closes #1403


spec/services/location_data_service_spec.rb
```diff
@@ -119,9 +131,10 @@
       location_record7 = Locations::HoldingLocation.find_by(code: 'online$cdl')
       location_record8 = Locations::HoldingLocation.find_by(code: 'recap$gp')
       location_record9 = Locations::HoldingLocation.find_by(code: 'recap$pb')
+      location_record10 = Locations::HoldingLocation.find_by(code: 'eastasian$hy')
```
jrgriffiniii, line 134: 👍

jrgriffiniii (approved):

PR #1425, opened 2021-06-16
Adds rake task to register a local PARTNER_RECAP file into the database

### LINKED ISSUE #1422: Add a rake task to create an event-dump_file for local development
It would be helpful to have a rake task for local development to create an event and all the relations based on a given constant and given file.

hectorcorrea:
Fixes #1422

### DIFF
--- a/lib/tasks/scsb.rake
+++ b/lib/tasks/scsb.rake
@@ -36,17 +36,19 @@ namespace :scsb do
     IndexFunctions.process_scsb_dumps([dump], ENV['SET_URL'])
   end
 
-  desc "Adds a local PARTNER_RECAP file to the database"
-  task add_local_partner_recap_dump: :environment do
-    abort "usage: FILE=full-file-path.xml.gz rake add_local_partner_recap_dump" unless ENV['FILE']
+  desc "Adds a local dump file to the database"
+  task add_local_dump_file: :environment do
+    abort "usage: FILE=full-file-path.xml.gz rake add_local_dump_file" unless ENV['FILE']
     ev = Event.new(success: true)
     ev.save
 
-    dump_type_id = DumpType.where(constant: "PARTNER_RECAP").first.id
+    dump_type = ENV['DUMP_TYPE'] || "PARTNER_RECAP"
+    dump_type_id = DumpType.where(constant: dump_type).first.id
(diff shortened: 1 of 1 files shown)


lib/tasks/scsb.rake
```diff
@@ -36,6 +36,21 @@ namespace :scsb do
...
   end
 
+  desc "Adds a local PARTNER_RECAP file to the database"
+  task add_local_partner_recap_dump: :environment do
+    abort "usage: FILE=full-file-path.xml.gz rake add_local_partner_recap_dump" unless ENV['FILE']
+    ev = Event.new(success: true)
+    ev.save
+
+    dump_type_id = DumpType.where(constant: "PARTNER_RECAP").first.id
+    dump = Dump.new(event_id: ev.id, dump_type_id: dump_type_id)
```
christinach, line 46: can we add constant: DUMP_TYPE so that we can pass any constant we want when we run the rake task?

christinach (changes requested):
@hectorcorrea thank you. I was thinking about this rake task to be more flexible across the different dump types and dump file types. I added a couple of questions. Also a mapping between DUMP_TYPE and DUMP_FILE_TYPE as a comment in the beginning of the rake task would be good.

hectorcorrea:
@christinach Made the changes that you requested. Notice that I left it with the defaults for RECAP files if no DUMP_TYPE and DUMP_FILE_TYPE are indicated.

hectorcorrea:
@christinach I am not sure I follow this: "Also a mapping between DUMP_TYPE and DUMP_FILE_TYPE as a comment in the beginning of the rake task would be good."

lib/tasks/scsb.rake
```diff
@@ -36,6 +36,23 @@ namespace :scsb do
     IndexFunctions.process_scsb_dumps([dump], ENV['SET_URL'])
   end
 
+  desc "Adds a local dump file to the database"
+  task add_local_dump_file: :environment do
+    abort "usage: FILE=full-file-path.xml.gz rake add_local_dump_file" unless ENV['FILE']
+    ev = Event.new(success: true)
```
hackartisan, line 42: I think the events page will fail to load in the UI unless it has some start / end dates. I almost never use the UI in development, though; doesn't seem worth blocking over.

hackartisan (approved):

PR #1427, opened 2021-06-16
Timeout values for deletes

### LINKED ISSUE #1389: Index deletions failing from timeout errors

hectorcorrea:
Sets a longer timeout for deletes since we are getting `Net::ReadTimeout` exceptions.

Fixes #1389 

We might still need to tweak these values, but I am hoping that the new longer value of 10 seconds allows at least some of the deletes to succeed. 

For what is worth, I am able to replicate the `Net::ReadTimeout` error if I set the timeout values to 1.

### DIFF
--- a/marc_to_solr/lib/solr_deleter.rb
+++ b/marc_to_solr/lib/solr_deleter.rb
@@ -26,7 +26,7 @@ class SolrDeleter
     def request_deletion(uri:, body:, content_type: 'text/xml')
       @logger&.info "Deleting #{body}"
       Faraday.post(uri) do |req|
-        req.headers = {"Content-Type" => content_type}
+        req.headers = { "Content-Type" => content_type }
         req.body = body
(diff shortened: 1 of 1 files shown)


jrgriffiniii (approved):

jrgriffiniii:
I was not aware of these options, but for reference, please see https://lostisland.github.io/faraday/usage/customize

PR #1442, opened 2021-06-23
Bump time out values for deletes

### LINKED ISSUE #1440: Bump up timeout value
Bump up the timeout values set in PR #1427

We are going to set all of them to 60 seconds. If they timeout again at least we'll have a baseline to use, rather than guess what the default/nil values mean.

Defaults values: https://ruby-doc.org/stdlib-2.7.1/libdoc/net/http/rdoc/Net/HTTP.html

hectorcorrea:
Bump time out values for deletes.

I also updated the code to log the time it took complete or timeout the delete operation. This should help us tune the parameters if the error happens again. I also updated the code to use the proper logger and catch a Faraday exception since Faraday wraps the standard Ruby Net HTTP Timeout exception.

Fixes #1440 

Log example after a successful execution:

```
2021-06-23T14:16:53-04:00  INFO Deleting <delete><id>SCSB-2146180</id><id>SCSB-2440575</id></delete>
2021-06-23T14:16:54-04:00  INFO Delete completed in 1.962307 seconds. URL: http://localhost:8983/solr/puldata/update?commit=true&wt=json
2021-06-23T14:16:54-04:00  INFO finished Indexer#process: 115 records in 3.044 seconds; 37.8 records/second overall.
```

Log example after a timeout exception (I forced the timeout value to 1 second to produce this log, hence the short timeout value reported)

```
2021-06-23T14:15:41-04:00  INFO Deleting <delete><id>SCSB-2146180</id><id>SCSB-2440575</id></delete>
2021-06-23T14:15:42-04:00  WARN Delete timed out after 1.019411 seconds. URL: http://localhost:8983/solr/puldata/update?commit=true&wt=json : <delete><id>SCSB-2146180</id><id>SCSB-2440575</id></delete>
2021-06-23T14:15:42-04:00  INFO Deleting <delete><id>SCSB-2146180</id><id>SCSB-2440575</id></delete>
2021-06-23T14:15:43-04:00  WARN Delete timed out after 1.005086 seconds. URL: http://localhost:8983/solr/puldata/update?commit=true&wt=json : <delete><id>SCSB-2146180</id><id>SCSB-2440575</id></delete>
2021-06-23T14:15:43-04:00  INFO finished Indexer#process: 115 records in 3.157 seconds; 36.4 records/second overall.
```

### DIFF
--- a/marc_to_solr/lib/solr_deleter.rb
+++ b/marc_to_solr/lib/solr_deleter.rb
@@ -25,15 +25,19 @@ class SolrDeleter
     # request payload is XML (even if the response is in JSON via the wt=json param)
     def request_deletion(uri:, body:, content_type: 'text/xml')
       @logger&.info "Deleting #{body}"
-      Faraday.post(uri) do |req|
+      start_time = Time.now
+      response = Faraday.post(uri) do |req|
         req.headers = { "Content-Type" => content_type }
         req.body = body
-        req.options.open_timeout = 10
-        req.options.read_timeout = 10
-        req.options.write_timeout = 10
+        req.options.open_timeout = 60
+        req.options.read_timeout = 60
+        req.options.write_timeout = 60
       end
-    rescue Net::ReadTimeout => net_read_timeout
-      Rails.logger.warn("Failed to transmit the POST request to #{uri}: #{body}")
+      @logger&.info("Delete completed in #{Time.now - start_time} seconds. URL: #{uri}")
+      response
+    rescue Faraday::TimeoutError
+      @logger&.warn("Delete timed out after #{Time.now - start_time} seconds. URL: #{uri} : #{body}")
+      nil
     end
 
     def build_request_body(ids:)
--- a/spec/marc_to_solr/lib/solr_deleter_spec.rb
+++ b/spec/marc_to_solr/lib/solr_deleter_spec.rb
@@ -44,22 +44,21 @@ describe SolrDeleter do
       end
     end
 
-    context 'when the HTTP request to the Solr endpoint' do
+    context 'when the HTTP request to Solr times out' do
       let(:logger) do
         instance_double(ActiveSupport::Logger)
       end
 
       before do
-        allow(Faraday).to receive(:post).and_raise(Net::ReadTimeout)
+        allow(Faraday).to receive(:post).and_raise(Faraday::TimeoutError)
         allow(logger).to receive(:info)
         allow(logger).to receive(:warn)
-        allow(Rails).to receive(:logger).and_return(logger)
 
         solr_deleter.delete(ids, batch_size)
       end
 
       it 'logs a warning' do
-        expect(logger).to have_received(:warn).exactly(2).times.with("Failed to transmit the POST request to #{url}: <delete><id>#{ids.first}</id><id>#{ids.last}</id></delete>")
+        expect(logger).to have_received(:warn).exactly(2).times
       end
     end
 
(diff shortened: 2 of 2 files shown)


marc_to_solr/lib/solr_deleter.rb
```diff
@@ -25,15 +25,19 @@ def delete(ids, batch_size = 100)
...
+        req.options.open_timeout = 60
+        req.options.read_timeout = 60
+        req.options.write_timeout = 60
       end
-    rescue Net::ReadTimeout => net_read_timeout
-      Rails.logger.warn("Failed to transmit the POST request to #{uri}: #{body}")
+      @logger&.info("Delete completed in #{Time.now - start_time} seconds. URL: #{uri}")
+      response
+    rescue Faraday::TimeoutError
+      @logger&.warn("Delete timed out after #{Time.now - start_time} seconds. URL: #{uri} : #{body}")
```
hectorcorrea, line 39: Notice that I am using `@logger` rather than `Rails.logger` since this part of the code does not run as a Rails app.

marc_to_solr/lib/solr_deleter.rb
```diff
@@ -25,15 +25,19 @@ def delete(ids, batch_size = 100)
...
-        req.options.write_timeout = 10
+        req.options.open_timeout = 60
+        req.options.read_timeout = 60
+        req.options.write_timeout = 60
       end
-    rescue Net::ReadTimeout => net_read_timeout
-      Rails.logger.warn("Failed to transmit the POST request to #{uri}: #{body}")
+      @logger&.info("Delete completed in #{Time.now - start_time} seconds. URL: #{uri}")
+      response
+    rescue Faraday::TimeoutError
```
hectorcorrea, line 38: Notice that I am rescuing a Faraday exception since Faraday wraps the native Ruby exception in its own thing.

spec/marc_to_solr/lib/solr_deleter_spec.rb
```diff
@@ -44,22 +44,21 @@
...
         allow(logger).to receive(:info)
         allow(logger).to receive(:warn)
-        allow(Rails).to receive(:logger).and_return(logger)
 
         solr_deleter.delete(ids, batch_size)
       end
 
       it 'logs a warning' do
-        expect(logger).to have_received(:warn).exactly(2).times.with("Failed to transmit the POST request to #{url}: <delete><id>#{ids.first}</id><id>#{ids.last}</id></delete>")
+        expect(logger).to have_received(:warn).exactly(2).times
```
hectorcorrea, line 61: I could not figure out how to write a `expect` that evaluates if the message includes a text (e.g. "time out") which is required now that I am including the elapsed time in the message.

hackartisan: I'm pretty sure you can pass it a regex instead of a string and then it will test there's a match on the regex.

hackartisan (commented):
This looks great. Do you want to try the test with a regex? If it doesn't work I would just merge it, but it seems worth a quick try at least.

spec/marc_to_solr/lib/solr_deleter_spec.rb
hackartisan: example: https://github.com/pulibrary/bibdata/blob/c642e47575620b4f2a2f9e391f302d9bd9f219f0/spec/services/numismatics_indexer_spec.rb#L78

hectorcorrea: @hackartisan Great idea, thanks!

hectorcorrea:
@hackartisan Added the message back to the test via a RegEx, thank for the tip!

hackartisan (approved):

jrgriffiniii (approved):

PR #1445, opened 2021-06-24
Returns a more sensible status for records with no items

### LINKED ISSUE #1314: Items in locations where flag "aeon_location" is true should have availability "Onsite access only"
They currently show as "some items not available" because they have no item records (no barcodes)

Example: https://catalog-alma-qa.princeton.edu/catalog/9917917633506421
 
For these items, locations are indexed so availability calls are not needed / availability status can be set and not updated. Alternately, bibdata could return "Online access only" to keep business logic all in one place.

hectorcorrea:
Fixes #1314


app/adapters/alma_adapter/availability_status.rb
```diff
@@ -64,6 +64,12 @@ def holding_status(holding:, sequence:)
         # Notice that we only check if a holding is available via CDL when necessary
         # because it requires an extra (slow-ish) API call.
         status[:cdl] = cdl_holding?(holding["holding_id"])
+      elsif status[:status_label] == "Some items not available"
+        # Alma gives an incorrect status when records have no items. Here we check
+        # if that is the case and return a more accurate status. This logic would be
+        # better inside the Status class but we keep it here since it requires an
+        # extra API call to get the item data.
+        status[:status_label] = "On-site access" if item_data.empty?
```
eliotjordan, line 72: If `item_data` is NOT empty, then the status label is is nil. Is this what we want?

hackartisan: Won't it still be "Some items not available"?

hectorcorrea: As @hackartisan said, it will only change the value if there are no items, and that is the intended behavior. 

These two tests shows it in action: https://github.com/pulibrary/bibdata/blob/issue_1314_onsite/spec/adapters/alma_adapter_spec.rb#L247-L257

hackartisan (approved):

PR #1458, opened 2021-06-29
1246 update spec for holdings_1display and electronic_access_1display

### LINKED ISSUE #1246: Update fixtures for skipped holdings tests in traject config spec
https://github.com/pulibrary/bibdata/blob/9a9e6ea96019abc180dd83f4ad03289d211562b9/spec/marc_to_solr/lib/config_spec.rb#L413-L448

christinach:
closes #1246

### DIFF
--- a/spec/marc_to_solr/lib/config_spec.rb
+++ b/spec/marc_to_solr/lib/config_spec.rb
@@ -19,6 +19,7 @@ describe 'From traject_config.rb' do
     @sample34 = @indexer.map_record(fixture_record('99105855523506421'))
     @sample35 = @indexer.map_record(fixture_record('9990567203506421'))
     @sample36 = @indexer.map_record(fixture_record('9981818493506421'))
+    @sample37 = @indexer.map_record(fixture_record('9976174773506421'))
     @manuscript_book = @indexer.map_record(fixture_record('9959060243506421'))
     @added_title_246 = @indexer.map_record(fixture_record('9930602883506421'))
     @related_names = @indexer.map_record(fixture_record('9919643053506421'))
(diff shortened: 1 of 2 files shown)


spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -403,43 +404,25 @@ def fixture_record(fixture_name)
...
-      end
+  describe 'holdings_1display' do
+    it 'groups holding info into a hash keyed on the mfhd id' do
+      @holdings = JSON.parse(@sample37["holdings_1display"][0])
+      expect(@holdings.keys).to eq ["22170509880006421", "22170509890006421", "22170509900006421"]
+      expect(@holdings['22170509880006421']['location_code']).to eq 'engineer$stacks'
+      expect(@holdings['22170509890006421']['location_code']).to eq 'lewis$stacks'
+      expect(@holdings['22170509900006421']['location_code']).to eq 'firestone$stacks'
+      # expect(@holding_block[holding_id]['location_note']).to include(holding['852']['z'])
     end
```
christinach, line 416: I would like to keep this comment here until I get an answer on my question regarding the location_note from @mzelesky .

hackartisan (commented):
This looks good to me. I am wondering if we might hold off on merging it until we figure out that one outstanding expectation, so we don't have to make another ticket, etc. If it doesn't seem like that will be possible in the next day or so I would be in favor of merging.

christinach:
I think it will be ok to merge this. The location note as far as a I remember is on the item level so it couldn't be 852. We would need to update traject.

christinach:
I asked @mzelesky to add the 852z and will export the record again.

mzelesky (approved):

hackartisan (approved):

PR #1459, opened 2021-06-29
Reintroducing sortable_call_number 

### LINKED ISSUE #1126: Fix bibliographic#bib_items sortable_call_number field
child of pulibrary/orangelight#2348

The [`add_locator_call_no` method](https://github.com/pulibrary/marc_liberation/blob/4bb139d15e87fc0b0df1ac3744d1185e30260b67/app/controllers/bibliographic_controller.rb#L234-L249) looks for a holding location "f". That will never be there with the new alma locations. If this field is needed (@kevinreiss ?) We need to re-enable the [call-number controller specs](https://github.com/pulibrary/marc_liberation/blob/4bb139d15e87fc0b0df1ac3744d1185e30260b67/spec/controllers/bibliographic_controller_spec.rb#L191-L219) with new fixtures and make sure it works.

This should happen when the holding location is `firestone#stacks`

Re: https://github.com/pulibrary/firestone_locator/blob/00ad7b643305ea055bee34cd717446aa950bb399/includes/item_proc.php#L141

The call number coming out of the Alma endpoint puts subfield k before the call number, so "oversize" and "folio" (and any other subfield k value) would be before the call number.

We'll need to adjust the call number as it comes from the Alma API to put the word "Oversize" at the end. We can do this by calling Alma's holding API, but it would be better to get it out of solr. In Alma we decide to index the subfield k that could include 'Oversize' at the end of the call_number.
We will use a new 'call_number_locator' field https://github.com/pulibrary/bibdata/issues/1136 for the lookup of the firestone locator when the record is in firestone. This new field will not include the 'Oversize' string.

hectorcorrea:
Fixes #1126 

This is a first pass at converting the call number from Alma into a `sortable_call_number` in the `/bibliographic/:mms_id/items` endpoint. This code also takes into consideration the "Oversize" token that Alma puts at the front of some call numbers.

Below is an example of how the response would look like for MMS_ID `9941598513506421`:

```
{
  "firestone$stacks":[ 
    {
      "holding_id":"22172120500006421",
      "call_number":"Oversize RA566.27 .B7544 2003q",
      "items":[an array goes here],
      "sortable_call_number":"RA.056627.B7544.2003--OVERSIZE"
    }
  ]
}
```

~~Notice that the code to reshuffle the call number and move the "Oversize" token at the end it's pretty brittle at the moment, we can improve this (e.g. make it case insensitive) if this is the approach that we want to take.~~


app/controllers/bibliographic_controller.rb
```diff
@@ -295,17 +295,37 @@ def bib_id_param
...
+
+    def make_sortable_call_number(call_no)
+      if call_no.start_with?("Oversize ")
+        call_no.sub("Oversize", "") + " Oversize"
+      else
+        call_no
+      end
+    end
+
     def add_locator_call_no(records)
```
hectorcorrea, line 312: The logic in this method is a bit more complicated than with Voyager because Alma inserts holdings in between and we have to loop through them.

app/controllers/bibliographic_controller.rb
```diff
@@ -295,17 +295,37 @@ def bib_id_param
...
     end
 
+    def needs_locator_call_no(location)
+      # For now only Firestore items need this.
+      return true if (location || "").start_with?("firestone$")
+      false
+    end
+
+    def make_sortable_call_number(call_no)
+      if call_no.start_with?("Oversize ")
```
hackartisan, line 305: There was some discussion that "oversize" isn't the only possible value here, for example we may have Folio or other values. In the ticket there's some discussion of checking solr or doing an API call. Otherwise we may need to see if we can configure a more exhaustive list? Given that these items are only going to be in firestone stacks maybe generating a list of possible values here would work fine; not sure how many special designated areas there might be.

app/controllers/bibliographic_controller.rb
```diff
@@ -295,17 +295,37 @@ def bib_id_param
       sanitize(params[:bib_id])
     end
 
+    def needs_locator_call_no(location)
+      # For now only Firestore items need this.
+      return true if (location || "").start_with?("firestone$")
```
hackartisan, line 300: From looking at the ticket it seems like we can `return true if location == "firestone$stacks`

hackartisan (changes requested):
I left some comments but I think it would be good if @kevinreiss can review at some point.

christinach:
@hectorcorrea I'm quoting what you wrote `This code also takes into consideration the "Oversize" token that Alma puts at the front of some call numbers.` We decided to put $k (that may include Oversize) in front of the call_number_display not Alma. See call_number_display: 
https://github.com/pulibrary/bibdata/blob/5c1f6ed3f4b2a20bb98c8e72ddbbdf3519cf257e/marc_to_solr/lib/traject_config.rb#L1307 vs call_number_browse_s: https://github.com/pulibrary/bibdata/blob/5c1f6ed3f4b2a20bb98c8e72ddbbdf3519cf257e/marc_to_solr/lib/traject_config.rb#L1315 where we keep $k at the end of the call_number_browse_s.

christinach (commented):
I left a comment but forgot to click review first.

app/controllers/bibliographic_controller.rb
christinach: We created https://github.com/pulibrary/bibdata/blob/5c1f6ed3f4b2a20bb98c8e72ddbbdf3519cf257e/marc_to_solr/lib/traject_config.rb#L1321 for this purpose. It does not include $k.

christinach (changes requested):
Yes what comes from the alma api has $k in front but in order to use solr we have a created a new field  where we don't include $k. https://github.com/pulibrary/bibdata/blob/5c1f6ed3f4b2a20bb98c8e72ddbbdf3519cf257e/marc_to_solr/lib/traject_config.rb#L1321

hectorcorrea:
@christinach @hackartisan I made a few adjustments per your review. In particular I narrowed the location check for "firestore$stacks" and made the code to adjust the call number a bit more generic and easier to expand to take other prefixes into consideration if we need to.

Notice that I am adjusting the call number *in place* rather than pulling the values from Solr. I am not totally pleased with this approach, I think it duplicates the logic from Traject, but, I am afraid that for records with more than one call number (e.g. https://catalog.princeton.edu/catalog/722287) I won't be able to match the values in Alma to the ones in Solr to pick the right sortable number. We can discuss this if we really want to use the Solr values.

hectorcorrea:
Oh yeah, I also need to add tests, but I am waiting to see if this is even the approach we want to take :)

hectorcorrea:
Added tests.

spec/controllers/bibliographic_controller_spec.rb
```diff
@@ -308,6 +308,24 @@
       end
     end
 
+    context "sortable call number " do
+      it "handles normal call numbers" do
+        stub_alma_bib_items(mms_id: "9965126093506421", filename: "cdl_9965126093506421_bib_items.json")
+        get :bib_items, params: { bib_id: "9965126093506421" }, format: 'json'
+        expect(response.status).to be 200
+        body = JSON.parse(response.body)
+        expect(body["firestone$stacks"].first["sortable_call_number"]).to eq "PS.3558.A62424.B43--2010"
```
hackartisan, line 317: Where is `--` coming from? It seems like it should just be a space.

hackartisan (changes requested):

spec/controllers/bibliographic_controller_spec.rb
hectorcorrea: That's what the normalized coded does, I would be surprised if the normalized code uses spaces since it would make it harder for humans to compare values (computers wouldn't care, but humans could get confused with spaces)

christinach (approved):

spec/controllers/bibliographic_controller_spec.rb
hackartisan: okay, cool!

hackartisan (approved):

PR #1460, opened 2021-06-29
Ensure legacy IDs for delivery locations come through on regenerate.

### LINKED ISSUE #1447: figgy holding_locations will break on go-live
The holding_locations in figgy use bibdata URIs from the delivery_locations table. The IDs aren't the same between old bibdata and new. Not sure if the entries are otherwise the same.

Need a migration plan here.

tpendragon:
Closes #1447

### DIFF
--- a/spec/services/location_data_service_spec.rb
+++ b/spec/services/location_data_service_spec.rb
@@ -172,6 +172,19 @@ RSpec.describe LocationDataService, type: :service do
       LocationDataService.delete_existing_and_repopulate
       library_record = Locations::Library.find_by(code: 'annex')
     end
+
+    it "sets a static ID" do
+      LocationDataService.delete_existing_and_repopulate
+      # Run a second time to ensure idempotency.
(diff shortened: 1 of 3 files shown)


spec/services/location_data_service_spec.rb
```diff
@@ -172,6 +172,17 @@
       LocationDataService.delete_existing_and_repopulate
       library_record = Locations::Library.find_by(code: 'annex')
     end
+
+    it "sets a static ID" do
+      LocationDataService.delete_existing_and_repopulate
```
hackartisan, line 177: Can we run this twice in the test to validate idempotence? Or if we're sure it's already been done can we add a comment pointing to where it's done during setup or whatever?

hackartisan (changes requested):

tpendragon:
@hackartisan Ready for re-review.

christinach (approved):

hackartisan (approved):
yay!

PR #1467, opened 2021-07-01
Do not index the isbn if there is no subfield $a

### LINKED ISSUE #785: ISBN field label displays when no subfield a
See https://catalog.princeton.edu/catalog/11059941

The label appears, even though no data is displayed in the field. In this case, there is an 020 $z, which I'm guessing is causing the field to display.

### LINKED ISSUE #1205: Correction on the isbn_display field
If there is no subfield $a in 020 then we should not display the isbn in the record page.

christinach:
closes #1205 
closes #785

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -1025,10 +1025,12 @@ to_field 'data_source_display', extract_marc('786at', trim_punctuation: true)
 
 # ISBN:
 #    020 XX a
+# Dont index if subfield $a is not present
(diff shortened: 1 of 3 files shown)


christinach:
@mzelesky can you review?
record 99121576653506421 with one 020 with no subfield $a
Screen Shot 2021-07-01 at 1 25 25 PM

record 99110599413506421 with one 020 and subfield $a
Screen Shot 2021-07-01 at 1 23 48 PM

mzelesky:
Looks good to me.

hackartisan (approved):

PR #1468, opened 2021-07-01
Index scsb records by the date the files were generated rather than the date the db object was created

### LINKED ISSUE #1402: Potential for skipping updated scsb partner files due to timestamps
Some analysis is needed here. We may have a potential for missing scsb partner update files due to the way full dump files are packaged.

The start / end times on the full dump event reflect the creating of the event in bibdata, not the create times of the partner files on S3.

Also, the fact that all the partner dumps from all the partner institutions are grouped together in one full dump, but aren't necessarily generated on the same day, could be an issue.

We may need to split out the full dump into multiple full dumps, one per partner institution, and update the indexing to ensure we get each one.

We may also want to update the event start / end to equal the create time from the s3 bucket, so that we can use that timestamp when locating update files from that point on.

hackartisan:
fixes #1402

### DIFF
--- a/app/models/dump.rb
+++ b/app/models/dump.rb
@@ -66,6 +66,10 @@ class Dump < ActiveRecord::Base
       end
   end # class << self
 
+  def subsequent_partner_incrementals
+    Dump.partner_recap.where(generated_date: generated_date..Float::INFINITY)
+  end
+
   def enqueued?
     index_status == ENQUEUED
   end
--- a/app/models/dump.rb
+++ b/app/models/dump.rb
@@ -40,8 +40,8 @@ class Dump < ActiveRecord::Base
       dump
     end
 
-    def latest
-      order('created_at desc').first
+    def latest_generated
+      order('generated_date desc').first
     end
 
     private
--- a/db/migrate/20210630201908_add_generated_date_to_dumps.rb
+++ b/db/migrate/20210630201908_add_generated_date_to_dumps.rb
@@ -1,0 +1,5 @@
+class AddGeneratedDateToDumps < ActiveRecord::Migration[5.2]
+  def change
+    add_column :dumps, :generated_date, :datetime
+  end
+end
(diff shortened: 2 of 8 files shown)


app/models/dump.rb
```diff
@@ -66,6 +66,10 @@ def partner_recap_full_dump_type
       end
   end # class << self
 
+  def subsequent_partner_incrementals
+    Dump.partner_recap.where(generated_date: generated_date..Float::INFINITY)
+  end
```
christinach, line 71: Why do you add from generated_date to infinity?

christinach: why not now .. Time.now? Isn't this what we want?

christinach: ok we take all the incrementals from the generated date and back.

db/migrate/20210630201908_add_generated_date_to_dumps.rb
```diff
@@ -0,0 +1,5 @@
+class AddGeneratedDateToDumps < ActiveRecord::Migration[5.2]
+  def change
+    add_column :dumps, :generated_date, :datetime
+  end
+end
```
christinach, line 5: Why we don't use the created_at?

christinach: ok generated_date is the datetime the dump file is complete and exists in the db.

hackartisan: created_at reflects the date that the file was downloaded, not the date the file itself was actually generated. This should give us a more accurate view of the age of the data itself.

hackartisan: `generated_date` is taken from the filename, and reflects the date that the file was created by the SCSB system.

christinach (approved):

PR #1475, opened 2021-07-02
Bib data updates for Figgy

### LINKED ISSUE #1271: Provide a way for figgy to check for updated bib records
See https://github.com/pulibrary/figgy/issues/4477 for details of current implementation.

hectorcorrea:
Generated the data that Figgy needs to detect Bib records that were deleted or updated. Fixes #1271 

This version provides a rake task that process the MARC files in a Dump and updates the `delete_ids` and `update_ids` in the Dump record (thanks for the idea @hackartisan). This will allow Figgy to pick up the data using the existing endpoints `/events.json` and `/dumps/xxx.json`. Example of usage:

```
DUMP_ID=23 bundle exec rake liberate:dump_log_ids
```

This version ditches Traject and parses the MARC XML directly to get the list of IDs (thanks for the tip @tpendragon)

Although the code only processes dumps that are for CHANGED_RECORDS this could still be a problem for Figgy if the "changed records" is the entire dataset. Since we'll probably run this rake task manually initially this might not be a big problem on day one, but something that we probably should address at some point.


marc_to_solr/lib/traject_log_ids_config.rb
```diff
@@ -0,0 +1,38 @@
...
+  if record.leader[5] == 'd'
+    delete_ids << id
+  else
+    update_ids << id
+  end
+end
+
+after_processing do
+  LogIdsService.save(delete_ids: delete_ids.to_a, update_ids: update_ids.to_a)
+end
```
hectorcorrea, line 38: This logic could be moved to our "normal" `traject_config.rb` but we will need to make sure that Traject is run within a Rails environment and we currently don't.

tpendragon:
The main question I have is what's the proposed process? I see a rake task, but I don't see it connected to anything. What am I missing?

app/services/log_ids_service.rb
```diff
@@ -0,0 +1,29 @@
...
+    indexer = Traject::Indexer.new
+    indexer.load_config_file(config_file)
+    indexer.process(file)
+  end
+
+  # Note: Our MARC files have 100K records so theorically the maximum number of
+  # deleted_ids or updated_ids will be that.
+  def self.save(delete_ids:, update_ids:)
+    # Create an event...
+    ev = Event.new(start: Time.now, finish: Time.now, success: true)
```
hackartisan, line 13: I'm surprised we're creating a new event and new dumps. I expected to process existing dumps and add the fields to them.

hackartisan (changes requested):
I think we want this as metadata on the existing dumps, so we don't get CHANGED_RECORDS Dumps without any changed records in them.

app/services/log_ids_service.rb
hectorcorrea: Fixed

app/services/dump_log_ids_service.rb
```diff
@@ -0,0 +1,64 @@
...
+      tar_extract = Gem::Package::TarReader.new(Zlib::GzipReader.open(tar_file))
+      tar_extract.tap(&:rewind)
+    end
+
+    def write_chunks(entry, temp_file)
+      while (chunk = entry.read(16 * 1024))
+        temp_file.write chunk
+      end
+      temp_file.tap(&:rewind)
+    end
```
hectorcorrea, line 63: These method were copied from https://github.com/pulibrary/bibdata/blob/alma/app/services/alma/indexer/dump_file_indexer.rb and could be refactored into a share class.

hackartisan (commented):
Yeah, I see a lot of places where we're decompressing files. It would be really good to get that sorted out. Do you think it's worth doing with this PR?

hectorcorrea:
@hackartisan I wouldn't want to mess with the indexing process so close to the release. I created issue #1522 to deal with this later.

hackartisan (approved):

hackartisan:
Makes sense to me.

PR #1503, opened 2021-07-08
update holding_library.rb- remove old recap locations

### LINKED ISSUE #1498: Cleanup the recap$ locations and set the Alma recap_edd flag  correctly
The recap_edd flag is always set to false.  It should be set to true in many recap locations.  

Here are the old location where recap_edd is set to true: etasrcp, rcpgp, rcpjq, rcppa, rcppb, rcppj, rcppk, rcppn, rcppq, rcppt, rcppw, rcpxw, scsbcul, scsbnypl

**Update**
The new locations are something like mendel$pk NOT recap$pk as is set up in bibdata.  The recap$ locations exist AND the mendel$ locations exists.  The flags are set correctly on the recap$ locations, but not on the mendel$ ect locations.  The recap$ locations should be purged and the new locations should have the correct flags set.

### LINKED ISSUE #1500: Update existing voyager-alma mapping file with the 'EDD allowed' column
see https://github.com/pulibrary/bibdata/issues/1498#issuecomment-876557821 for details

### LINKED ISSUE #1501: Update holding_library file with ReCAP value and keys based on the new recap locations
Following the changes in the tickets https://github.com/pulibrary/bibdata/issues/1500 and https://github.com/pulibrary/bibdata/issues/1498 we would need to update the holding_library.rb file so that the new recap locations are grouped under ReCAP in the advanced search.

christinach:
closes #1500,
closes #1501,
closes #1498
updates templates


christinach (commented):
Reviewed in my local environment with @kevinreiss the locations. We had to https://github.com/pulibrary/bibdata/issues/1510 update in bibdata production. I will have to regenerate again the files so that they will pick the new change in production bibdata and push up the changes.

christinach:
I pushed the new changes after regenerating the file because of https://github.com/pulibrary/bibdata/issues/1510

spec/services/location_data_service_spec.rb
```diff
@@ -185,6 +245,45 @@
       new_location = FactoryBot.create(:delivery_location)
       expect(new_location.id).to eq 41
     end
+
+    describe "new recap locations" do
+      before do
+        LocationDataService.delete_existing_and_repopulate
+        @location_engineer_pt = Locations::HoldingLocation.find_by(code: 'engineer$pt')
```
hackartisan, line 252: rubocop doesn't like these; you could just define them directly in the `it` block.

christinach: there are 2 it blocks that uses them

hackartisan: doesn't the 2nd one just use the one variable? I think it's fine to define it again. Or you can do `let`s like rubocop says. whichever you prefer!

christinach: yeah it's only the one in the second it.

hackartisan (approved):

PR #1512, opened 2021-07-12
Update file names - locations fixtures

### LINKED ISSUE #1511: Rename the locations<number>.json files using the location name
We start to have many fixtures locations1,2,3,4 files and it will be easier to know the content of each file (which library locations each one includes) if we rename them as `locations_<location_name>.json` etc. for example `locations_firestone.json`

christinach:
closes #1511


spec/services/location_data_service_spec.rb
```diff
@@ -194,7 +194,7 @@
       location_record10 = Locations::HoldingLocation.find_by(code: 'eastasian$hy')
 
       expect(Locations::Library.count).to eq 11
-      expect(Locations::HoldingLocation.count).to eq 113
+      expect(Locations::HoldingLocation.count).to eq 102
```
hackartisan, line 197: why did the count go down? It seems like the contents of the files hasn't really changed.

christinach: I was looking the same thing before I opened the pr. I I will have an answer soon.

christinach: The previous code included **firestone$pb** in the locations1.json. This file had 11 locations. 10 belonged to arch and 1 to firestone. The same file `locations1.json` was used twice to stub the request for **firestone locations** and for **arch locations**. As a result we were creating 10 firestone locations that do not really exist and 1 arch location that does not really exist. I separated them and moved the firestone location in its own file and used it to stub the firestone request location.

christinach: 113 - arch$pb (one not real arch location) - (ten not real firestone locations) = 102

hackartisan: 👍🏻 thanks for hunting it down.

hackartisan (approved):

PR #1527, opened 2021-07-14
Index incrementals in a Batch

tpendragon:


christinach (dismissed):

christinach (commented):
What happens when there is an error?

tpendragon:
> What happens when there is an error?

The job fails and it restarts the job.

hackartisan (approved):

PR #1539, opened 2021-07-16
Combine locations gem into bibdata

### LINKED ISSUE #1193: Remove locations forms from the bibdata UI
Remove the New, Create, Edit, Destroy, Update actions from:
Libraries
Holding Locations
Delivery Locations

We don't want to do any updates on the data through the UI but only through the files that will be created in https://github.com/pulibrary/bibdata/issues/1103

Discussed with @kevinreiss

eliotjordan:
- Ports logic, views, and specs from locations gem
- Removes Create, Update, Delete functionality

Closes #1193


hackartisan (changes requested):
It seems like we'll have to deploy this with some downtime for the locations tables and I'm not entirely sure about the implications for that.

I think there is a way to add a table prefix in model code. Maybe it would make sense to keep the migrations we have and use a prefix? I think that would mean we could refer to them by their shorter  names in code, but they would still have the prefix in the db.

eliotjordan:
@hackartisan Removed the migration changes so we can migrate to un-prefixed tables in future PRs.

Issue #1531, opened 2021-07-15
SCSB partner full record download / processing does not complete due to `ActiveRecord::StatementInvalid: PG::UnableToSend: no connection to the server`


hackartisan:
https://app.honeybadger.io/projects/54497/faults/80414654
https://app.honeybadger.io/projects/54497/faults/80435006

hackartisan:
This happens when it's run in the background or switched to the foreground via perform_now.

hackartisan:
fixed by #1542

PR #1542, opened 2021-07-19
Attach each SCSB DumpFile as it's processed

hackartisan:
2 reasons:

1. It will hit the database more often, hopefully keeping our connection
   alive
2. Linking the files as we get them will make it easier to clean up a
   failed run -- deleting the Event will clean out the files on disk.

refs #1531


tpendragon (approved):

hackartisan:
The job completed running this way.

PR #1563, opened 2021-07-29
Issue 1557 ark v2

hectorcorrea:
This is a slight enhancement to the work in PR #1559. Now the code detects the invalid URL with a single forward slash (e.g. http:/something), logs the error, and does not produce the invalid link for the `electronic_access_1display` field in Solr. And it does not crash Traject.

Notice that I moved some of the tests that were under `spec/marc_to_solr/lib/uri_ark_spec.rb` to be under `spec/marc_to_solr/lib/princeton_marc_spec.rb` since when a MARC file is being processed these checks happen before the ARK code, i.e. the ARK class will not get the bad URLs.

### DIFF
--- a/marc_to_solr/lib/electronic_access_link.rb
+++ b/marc_to_solr/lib/electronic_access_link.rb
@@ -44,14 +44,12 @@ class ElectronicAccessLink
       if !@url_key&.match?(URI.regexp)
         @logger.error "#{@bib_id} - invalid URL for 856$u value: #{@url_key}"
         @url_key = nil
+      elsif @url_key.start_with?(/http:\/[A-Za-z]/)
+        # Misleading URL "http:/" instead of "http://"
+        @logger.error "#{@bib_id} - invalid URL for 856$u value (http:/): #{@url_key}"
+        @url_key = nil
       else
-        if @url_key.start_with?(/http:\/[A-Za-z]/)
-          # Misleading URL "http:/" instead of "http://"
-          @logger.error "#{@bib_id} - invalid URL for 856$u value (http:/): #{@url_key}"
-          @url_key = nil
-        else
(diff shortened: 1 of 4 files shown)


marc_to_solr/lib/electronic_access_link.rb
```diff
@@ -44,6 +44,10 @@ def url
       if !@url_key&.match?(URI.regexp)
         @logger.error "#{@bib_id} - invalid URL for 856$u value: #{@url_key}"
         @url_key = nil
+      elsif @url_key.start_with?(/http:\/[A-Za-z]/)
```
hectorcorrea, line 47: This is needed because Ruby's built-in `URI.regexp` does not detect that "http:/whatever" is an invalid URL.

marc_to_solr/lib/uri_ark.rb
```diff
@@ -3,6 +3,9 @@
 class URI::ARK < URI::Generic
   attr_reader :nmah, :naan, :name
 
+  ARK_REGEX = /\:\/\/(.+)\/ark\:\/(.+)\/(.+)\/?/.freeze
```
hectorcorrea, line 6: Notice that I reverted to the original RegEx (more strict) version that we were using before PR 1559

marc_to_solr/lib/uri_ark.rb
```diff
@@ -3,6 +3,9 @@
...
+  PRINCETON_ARK_REGEX = /[\/?]ark\:\/88435\/(.+)\/?/.freeze
```
hackartisan, line 7: This is nice to move these to constants

hackartisan (approved):
This looks great; it's nice to have them handled in a way that logs the error.

christinach (approved):

Issue #1564, opened 2021-07-29
Availability slowness in show results


hectorcorrea:
On the production infrastructure the Availability JavaScript call takes a long time to come back. Datadog shows some pages taking up to a minute and a non-insignificanr number of them taking 20-30 seconds ([example](https://app.datadoghq.com/apm/resource/marc_liberation/rack.request/630c154bff820882?query=env%3Anone%20service%3Amarc_liberation%20operation_name%3Arack.request%20resource_name%3ABibliographicController%23availability_many&env=none&index=apm-search&spanID=87359450771159505&start=1627582835858&end=1627586435858&paused=false)).

It looks like the initial call to the `/bibs` endpoint with 20 MMS IDs takes 5 seconds, but since we make a second call to `/bibs/.../holdings/ALL/items` when a record is not available the aggregated response time is getting to be too long.

For now we are going to try to only make the `/bibs` call and let the client make the second call. This will reduce the time that bib data takes to respond on the initial call.

## Example

On this example, the initial `/bibs` call takes 8 seconds, but the `/holding/items` take 5 seconds each so the total request time balloons to 40 seconds.

Initial `bibs` call took 8 seconds:

```
/almaws/v1/bibs?mms_id=99123792583506421%2C99123831363506421%2C99124832743506421%2C99123194243506421%2C99124376113506421%2C99124412473506421%2C99124620593506421%2C99123653083506421%2C99123933753506421%2C99124620603506421%2C99123460183506421%2C99122329203506421%2C99123460243506421%2C99121609463506421%2C99124412493506421%2C99122545793506421%2C99122612103506421%2C99124511593506421%2C99124680603506421%2C99124511603506421&expand=p_avail%2Ce_avail%2Cd_avail%2Crequests
```

But then it executed the following calls and __each__ of them took an additional 4-5 seconds:

```
/almaws/v1/bibs/99123194243506421/holdings/ALL/items?           5 seconds
/almaws/v1/bibs/99123792583506421/holdings/ALL/items?           4 seconds
/almaws/v1/bibs/99124376113506421/holdings/ALL/items?           5 seconds
/almaws/v1/bibs/99124412473506421/holdings/ALL/items?           4 seconds
/almaws/v1/bibs/99124620593506421/holdings/ALL/items?           4 seconds
/almaws/v1/bibs/99124620603506421/holdings/ALL/items?           4 seconds
/almaws/v1/bibs/99124832743506421/holdings/ALL/items?           4 seconds
```

PR #1565, opened 2021-07-30
Makes the CDL check only when client request a deep availability check

hectorcorrea:
Address the slowness in the calls for availability (issue #1564)

Since the biggest performance offender seems to be the extra calls that we make to `/bibs/.../holdings/ALL/items` this PR stops making that extra call automatically for all Unavailable records. Instead, we only make the call for Unavailable records _if the client requests it_ (via the deep parameter). 

For now we are not making any changes to Orangelight. OL will just get "Unavailable" for records that Alma report as Unavailable and no indication whether they are available through CDL.

See also https://github.com/pulibrary/orangelight/pull/2597 for changes to Orangelight to preserve the CDL flag *on the show page*.


hackartisan (commented):
Looks like a good direction to me.

tpendragon (approved):

PR #1579, opened 2021-08-05
Exclude inactive electronic portfolios/ update electronic_portfolio_s logic

### LINKED ISSUE #1576: Dont index electronic portfolios if 951$a has an inactive status

### LINKED ISSUE #1589: Update the electronic_portfolio_s logic
- Current implementation: The logic that indexes the electronic_portfolio_s is depending on an existing collection. An existing collection has a name that comes from 951 $n.
- Defect: If the portfolio is not part of an existing collection, it has no title. The 'Available Online' section includes the active url in the DOM but it appears empty in the UI. Example record: https://catalog.princeton.edu/catalog/9996338243506421

Screen Shot 2021-08-10 at 1 48 21 PM

- Proposal: If the portfolio is not part of an existing collection, use the title_display value as part of the electronic_portfolio_s['title']. 

The above is after reviewing various records in Alma with @mzelesky.
- Impact: This solution may result in large titles in the 'Available Online' section.

christinach:
closes #1576 
closes #1589

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -1349,9 +1349,6 @@ to_field 'call_number_locator_display' do |record, accumulator|
 end
 
 to_field 'electronic_portfolio_s' do |record, accumulator|
-  fields = []
-  dates = []
-  embargoes = []
   # Don't check for scsb
   fields = alma_951_active(record)
   dates = alma_953(record)
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -1352,7 +1352,9 @@ to_field 'electronic_portfolio_s' do |record, accumulator|
   fields = []
   dates = []
   embargoes = []
-  # dont check for scsb
+  # Don't index if 951$a is 'Not Available' - inactive portfolio
+  # Don't check for scsb
+  next if inactive_electronic_portfolio(record)
   MarcExtractor.cached('951knx').collect_matching_lines(record) do |field, _spec, _extractor|
     next unless alma_951(record)
     fields << field
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -321,8 +321,8 @@ to_field 'cataloged_tdt' do |record, accumulator|
   unless /^SCSB-\d+/.match?(doc_id)
     cataloged_date = if alma_876(record) && alma_876(record).map { |f| f['d'] }.compact.present?
                        alma_876(record).map { |f| f['d'] }.sort.first
-                     elsif alma_951(record) && alma_951(record).map { |f| f['w'] }.compact.present?
-                       alma_951(record).map { |f| f['w'] }.compact.sort.first
+                     elsif alma_951_active(record) && alma_951_active(record).map { |f| f['w'] }.compact.present?
+                       alma_951_active(record).map { |f| f['w'] }.compact.sort.first
                      else
                        alma_950(record)
                      end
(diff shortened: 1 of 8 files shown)


christinach:
fixes bad inactive portfolio links that appear in the search results.
Fixes empty available online section in orangelight.

Screen Shot 2021-08-05 at 5 51 47 PM
Screen Shot 2021-08-05 at 5 51 34 PM

christinach:
vs what is in the catalog with the inactive portfolios:

Screen Shot 2021-08-05 at 5 40 03 PM
Screen Shot 2021-08-05 at 5 38 24 PM

marc_to_solr/lib/traject_config.rb
```diff
@@ -1352,7 +1352,9 @@
   fields = []
   dates = []
   embargoes = []
-  # dont check for scsb
+  # Don't index if 951$a is 'Not Available' - inactive portfolio
+  # Don't check for scsb
+  next if inactive_electronic_portfolio(record)
```
tpendragon, line 1357: Is there ever a time when one of the 951s is inactive, but other ones aren't? If so, I wonder if this should go down in 1359 similar to `next unless alma_951(record)`?

christinach: yes there is this possibility. let me check

christinach: I'll try to find a fixture with active and inactive

christinach (commented):
@mzelesky reviewed the pr and I need to do some changes to make it also more efficient. I will also add a custom fixture that has active and inactive portfolios.

christinach:
@mzelesky if you could review this pr with me tomorrow. I updated and refactored the code. I'm including also a fixture that has active portfolios and a custom inactive one.

marc_to_solr/lib/traject_config.rb
```diff
@@ -1352,21 +1352,11 @@
...
-    next unless alma_953(record)
-    dates << field
-  end
-
-  MarcExtractor.cached('954ac').collect_matching_lines(record) do |field, _spec, _extractor|
-    next unless alma_954(record)
-    embargoes << field
-  end
+  # Don't index if 951$a is 'Not Available' - inactive portfolio
+  # Don't check for scsb
```
christinach, line 1356: @mzelesky I also have a question for the dont check for scsb rule. It was not added. Do I want to add it?

christinach:
Screen Shot 2021-08-09 at 4 52 00 PM

vs the empty Available online section:

Screen Shot 2021-08-09 at 4 53 12 PM

christinach:
In this example: https://catalog.princeton.edu/catalog/9996338243506421 the available online section appears empty because the electronic_portfolio url of the sibling record https://catalog.princeton.edu/catalog/9995002873506421/raw does not include a title. It is empty. What are we doing when the title is empty? Do we display the url or a part of it?
The 951 $8 indicates that the sibling_record is an alma record. 
I will review with @mzelesky the 856 of the sibling record to make sure it is indexed as expected.

christinach:
Following up on this comment https://github.com/pulibrary/bibdata/pull/1579#issuecomment-895655636
We reviewed with @mzelesky various alma records and different electronic portfolios. A new issue is that there are active portfolios but with no title. The reason is that the existing electronic_portfolio_s logic depends only in an existing collection. We should update the logic with a new condition for cases where the portfolio is part of a book and not a collection. I will open a different issue for that.

marc_to_solr/lib/traject_config.rb
christinach: @mzelesky said it's ok not to add it.

marc_to_solr/lib/traject_config.rb
```diff
@@ -1352,21 +1352,10 @@
   fields = []
```
hackartisan, line 1352: Should these be deleted? they're being set again directly below.

christinach:
With the change in #1589 we have 'Full Text' as a title for electronic portfolios that are not part of a collection.:

Screen Shot 2021-08-10 at 4 11 30 PM

christinach:
Left side is from catalog-qa (the fixed), right is from production catalog.
https://catalog-qa.princeton.edu/catalog/99123430173506421
https://catalog-qa.princeton.edu/catalog?utf8=%E2%9C%93&search_field=all_fields&q=lux+magazine
https://catalog-qa.princeton.edu/catalog/9996338243506421

lux_magazine

full_text

hackartisan (approved):

Issue #1518, opened 2021-07-12
Index all incremental dumps created since the most recent full dump 


hackartisan:
This would ideally happen as part of the full dump task, but if all the dump files are backgrounded we don't want to risk any of the incrementals running before the fulls so that may not work.

Make sure to update indexing documentation as part of this.

## Technical Implementation

An IndexManager to coordinate indexing alma full and incremental dumps.

```
class IndexManager < ActiveRecord::Base
  solr_collection string
  dump_in_progress foreign key (tentative)
  last_dump_completed foreign key

  # returns the active IndexManager for the given collection
  def self.current(solr_collection)

  # finds and indexes the next incremental dump. no-op if there's a dump_in_progress.
  def next_incremental
```

Running a new full index on catalog-alma-rebuild:
  - Delete `last_dump_completed` from current Index Manager for the rebuild collection
  - empty the rebuild index via the solr API
  - queue up the initial full dump
  - when a full dump finishes
    - adds the dump to last_dump_completed
    - it calls IndexManager.current(collection).next_incremental
    - the full dump index jobs will have to store a reference to the index manager or the collection they're going into for this to work.
    - sidekiq batches make an action like this possible via the batch callback

Swapping the rebuild index into production use
  - Change the solr_collection value
  - do the alias swap via the solr API
  - Alternative: don't use the alias as the solr_collection value? This means we may need to look up what the alias is pointing to using the solr API pretty frequently.

Ensuring all incrementals get indexed, and in the correct order
  - When a new incremental file gets transferred from sftp, call next_incremental on the configured primary `SOLR_COLLECTION`
  - When any incremental file finishes indexing, call next_incremental on the collection it indexed into
    - sidekiq batches make an action like this possible via the batch callback

Instead of configuring `SOLR_URL` and `SOLR_REINDEX_URL` in env, we configure `SOLR_BASE_URL`, `SOLR_COLLECTION`, and `SOLR_REINDEX_COLLECTION`.

tpendragon:
Yeah we can't have one incremental run before another either. To handle this I think we'd have to either check to make sure there aren't started DumpFiles from an earlier dump enqueued or have some sort of managing coordinator in the database.

hackartisan:
Hm. A managing coordinator in the database seems tricky when you want to manage building a new index on catalog-alma-rebuild. A given incremental file may have been indexed into the live prod index but not into the rebuild index, or vice-versa.

hackartisan:
[...]

PR #1590, opened 2021-08-10
IndexManager to manage control of indexing into separate Solr URLs.

tpendragon:
Work towards #1518 

- [x] Rake task to create an IndexManager on first deploy which has the solr URL and the most recently created dump.
- [x] Rake task/class method to do a full reindex.

### DIFF
--- a/spec/models/index_manager_spec.rb
+++ b/spec/models/index_manager_spec.rb
@@ -93,12 +93,13 @@ RSpec.describe IndexManager, type: :model do
       expect(DumpFileIndexJob).to have_received(:perform_async).exactly(2).times
       expect(index_manager).not_to be_in_progress
     end
-    it "indexes the next incremental if the most recent full dump has been done" do
+    it "indexes the previous incremental if the most recent full dump has been done" do
       allow(DumpFileIndexJob).to receive(:perform_async).and_call_original
-      # This incremental is before the full dump, don't run it
+      # This incremental is before the pre-full-dump incremental, don't run it
+      pre_pre_incremental_event = FactoryBot.create(:incremental_dump_event, start: Time.current - 2.days, finish: Time.current - 2.days + 100)
       pre_incremental_event = FactoryBot.create(:incremental_dump_event, start: Time.current - 2.days, finish: Time.current - 2.days + 100)
       full_event = FactoryBot.create(:full_dump_event, start: Time.current - 1.day, finish: Time.current - 1.day + 100)
-      # This should get run on the second call
+      # This should get run on the third call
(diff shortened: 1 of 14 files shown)


tpendragon:
The full reindex logic may not be correct. The first incremental is a full index, as is the first "full index". Do we have some way of identifying an incremental which is everything?

spec/models/index_manager_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+      # There's one record in 1.xml, and one record in 2.xml
+      expect(response['response']['numFound']).to eq 2
+      index_manager.reload
+      expect(index_manager.last_dump_completed).to eq full_event.dump
+      expect(index_manager.dump_in_progress).to be_nil
+      expect(index_manager).not_to be_in_progress
+    end
+    it "doesn't index anything if it's caught up" do
+      allow(DumpFileIndexJob).to receive(:perform_async).and_call_original
+      # This incremental is before the full dump, don't run it
```
hackartisan, line 79: Not sure this comment is accurate?

spec/models/index_manager_spec.rb
```diff
@@ -0,0 +1,181 @@
...
+        index_manager = described_class.find(index_manager.id)
+        index_manager.index_next_dump!
+        run_callback(Sidekiq::BatchSet.new.to_a.last)
+      end
+      solr.commit
+
+      expect(DumpFileIndexJob).not_to have_received(:perform_async).with(pre_pre_incremental_event.dump.dump_files.first.id, anything)
+      expect(DumpFileIndexJob).to have_received(:perform_async).with(pre_incremental_event.dump.dump_files.first.id, anything)
+      expect(DumpFileIndexJob).to have_received(:perform_async).with(incremental_event.dump.dump_files.first.id, anything)
+      expect(DumpFileIndexJob).not_to have_received(:perform_async).with(final_incremental_event.dump.dump_files.first.id, anything)
```
hackartisan, line 126: This one doesn't run just because you didn't run the callbacks manually enough times?

hackartisan (changes requested):
This is really awesome and exciting!

One minor fix inline.

My current thoughts are around
- what it looks like to swap an index - the index managers will have the alias urls so they'll need to be updated.
- what updates we will need to documentation

But none of that needs to be addressed yet in this PR

spec/models/index_manager_spec.rb
tpendragon: Yes.

hackartisan (approved):

PR #1619, opened 2021-08-18
Use new custom docker image for CI.

tpendragon:


tpendragon:
We can't fix this without upgrading Traject (our old version doesn't support HTTP auth.)

PR #1627, opened 2021-08-20
1623 partner location label

### LINKED ISSUE #1623: Recap materials from partners institutions should include "Remote Storage" in the location label.
This is how it displays for our materials, so it should also look like that for partner materials.

This ticket comes from the 8/19 Alma Open House.

christinach:
closes #1623


christinach:
Tested on bibdata_staging and catalog-qa. 
The scsb location is being retrieved from bibdata directly.
If bibdata has no entries then it falls back to the solr_document, holdings_1display field (same as the rest of the pul records.)

The difference between SCSB location display and pul location display is that for the second we use javascript to get the availability through the availability.json and if JS is not working then it falls back to bibdata; if bibdata is nil then it falls back to the indexed solr_document.

Screen Shot 2021-08-20 at 12 05 42 PM

Screen Shot 2021-08-20 at 12 05 25 PM

tpendragon (approved):

Issue #1625, opened 2021-08-19
Perform a full reindex


tpendragon:
This will include Harvard records.

We'll have to manually switch the IndexManager to point to the new collection location when we switch "-rebuild" with the normal index.

Otherwise, follow the directions in `docs/indexing.md`

PR #1634, opened 2021-08-24
Set liberate:full to use IndexManager.reindex!

hackartisan:
refs #1625


hackartisan:
Ran this in staging and it started indexing the correct dump (the most recent All Records dump)
Screen Shot 2021-08-24 at 10 17 31 AM

hectorcorrea (approved):

PR #1638, opened 2021-08-25
Update index url and theses section in the indexing notes

christinach:


hackartisan (changes requested):
Can you use the url with the SHA in it instead of main?

PR #1639, opened 2021-08-25
Document how to swap IndexManagers

hackartisan:
refs #1625


lib/tasks/index_manager.rake
```diff
@@ -8,4 +8,19 @@ namespace :index_manager do
     manager.save!
     puts "Created IndexManager for #{solr_url} pointed to Dump ID #{solr_url}"
   end
+
+  desc "Swap rebuild index manager for production index manager. Do this RIGHT BEFORE swapping the alias for a rebuild."
+  task promote_rebuild_manager: :environment do
+    old_prod_manager = IndexManager.for(Rails.application.config.solr["url"])
+    prod_alias = old_prod_manager.solr_collection
+    new_prod_manager = IndexManager.for("#{Rails.application.config.solr['url']}-rebuild")
```
hackartisan, line 16: In prod they're currently called "catalog-alma-production" and "catalog-alma-rebuild" -- we should change the rebuild index to follow this pattern so that it's the same as in staging, and this will work.

tpendragon: Weird, how did https://github.com/pulibrary/bibdata/blob/1625-document-swap/app/models/index_manager.rb#L20 work?

hackartisan: We passed in the param instead of using that default.

PR #1648, opened 2021-09-03
On-site access for map holdings

hectorcorrea:
Adjusted logic to return "On-site Access" as the availability for certain specific map locations.

Fixes https://github.com/pulibrary/orangelight/issues/2711

No changes needed in Orangelight.

### DIFF
--- a/app/adapters/alma_adapter/status.rb
+++ b/app/adapters/alma_adapter/status.rb
@@ -14,11 +14,21 @@ class AlmaAdapter::Status
   end
 
   def to_s
-    return "On-site Access" if aeon
+    return "On-site Access" if on_site_holding?
     return "Some items not available" if holding["availability"] == "check_holdings"
     return holding["availability"].titlecase if holding["availability"]
 
     # For electronic holdings
     return holding["activation_status"].titlecase if holding["activation_status"]
+  end
+
+  def on_site_holding?
+    return true if aeon
+    on_site_locations.include? "#{holding['library_code']}$#{holding['location_code']}"
+  end
+
+  # Holdings in these location are for on-site use only
+  def on_site_locations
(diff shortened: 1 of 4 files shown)


app/adapters/alma_adapter/status.rb
```diff
@@ -14,11 +14,21 @@ def initialize(bib:, holding:, aeon: false)
...
+
+  def on_site_holding?
+    return true if aeon
+    on_site_locations.include? "#{holding['library_code']}$#{holding['location_code']}"
+  end
+
+  # Holdings in these location are for on-site use only
+  def on_site_locations
+    ["lewis$map", "lewis$maplf", "lewis$maplref", "lewis$mapmc", "lewis$mapmcm"]
+  end
```
christinach, line 33: looks ok to me. We might need to change the list of the map locations and keep only some of them; but this will be decided after we meet with the map team. I think for now it is ok to merge this.

spec/adapters/alma_adapter_spec.rb
```diff
@@ -260,6 +261,13 @@
       expect(item[:status_label]).to eq "On-site Access"
     end
 
+    it "reports on-site access for some specific (map) locations" do
+      FactoryBot.create(:map_location, code: 'lewis$mapmc', label: 'Lewis Library - Map Collection. Map Case')
+      availability = adapter.get_availability_one(id: "9968442613506421")
```
christinach, line 266: @hectorcorrea we will have a meeting with the maps team to update the list of the locations. We should update the fixture after the meeting  as well, unless you want to use a lewis$map fixture which is the code they use for the map location for sure.

hectorcorrea: Yeah, we can totally update this list at anytime.

christinach (approved):

christinach:
Screen Shot 2021-09-07 at 10 51 45 AM
Screen Shot 2021-09-07 at 10 50 55 AM

PR #1660, opened 2021-09-21
First pass at Change The Subject

bess:
We will need a more robust system for managing term replacements over
time, but this should get us started on a proof of concept.

Work toward #595

### DIFF
--- a/app/services/change_the_subject.rb
+++ b/app/services/change_the_subject.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+##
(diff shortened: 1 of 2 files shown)


hectorcorrea (approved):

app/services/change_the_subject.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+##
+# The creation and management of metadata are not neutral activities.
```
hectorcorrea, line 4: 👍🏼

PR #1670, opened 2021-09-27
Display the deployed version in the application footer

### LINKED ISSUE #1668: Add a version footer
Following [this example](https://github.com/pulibrary/oawaiver/pull/49/files), add a version footer to the application

- [x] The application shows me what version has been deployed and when it was last deployed

bess:
Closes #1668 

Screen Shot 2021-09-27 at 1 58 54 PM

### DIFF
--- a/config/initializers/git_sha.rb
+++ b/config/initializers/git_sha.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+revisions_logfile = Rails.root.join("..", "..", "revisions.log")
+
+GIT_SHA =
+  if File.exist?(revisions_logfile)
+    `tail -1 #{revisions_logfile}`.chomp.split(" ")[3].gsub(/\)$/, "")
+  elsif Rails.env.development? || Rails.env.test?
+    `git rev-parse HEAD`.chomp
+  else
+    "Unknown SHA"
+  end
+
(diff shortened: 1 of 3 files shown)


hectorcorrea (approved):

config/initializers/git_sha.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+revisions_logfile = Rails.root.join("..", "..", "revisions.log")
+
+GIT_SHA =
+  if File.exist?(revisions_logfile)
+    `tail -1 #{revisions_logfile}`.chomp.split(" ")[3].gsub(/\)$/, "")
```
hectorcorrea, line 7: Perhaps a future improvement could be to cache this value

```
tail -1 #{revisions_logfile}`.chomp.split(" ")
```

so that we don't `tail` this file three times (which requires a call to the shell). 

This is probably not an issue for bibdata where this page is not loaded very often, but it would be an issue other projects, like Orangelight, where the page is displayed quite often.

hectorcorrea: The values themselves (GIT_SHA, BRANCH, and LAST_DEPLOYED) are calculated only once, right? since they are in UPPERCASE.

Issue #1548, opened 2021-07-22
Index a defined set of topics under "Indigenous Studies" to boost their visibility


kelynch:
![Screen Shot 2021-09-29 at 11 25 41 AM](https://user-images.githubusercontent.com/65608/136403275-ec6f74c7-a0cb-4c86-af23-cee3bf3d3441.png)

![Screen Shot 2021-10-05 at 11 54 28 AM](https://user-images.githubusercontent.com/65608/136403223-715d931d-1c98-453f-a7ac-d03f0f1af23d.png)

Related to #595 

Index topics in [this list](https://docs.google.com/spreadsheets/d/1Y_GdQi7-VPdSzpgLk3Phn2IR6tst-dTO/edit#gid=714979645) to be discoverable under the "Indigenous Studies" keyword search to boost their visibility. 

Example: [This search](https://catalog.princeton.edu/?f%5Bsubject_topic_facet%5D%5B%5D=Kootenai+language&q=Kootenai%20language&search_field=subject) for "Kootenai language" with "Subject (Browse)" selected from the search dropdown should be discoverable under the term "Indigenous Studies" searched from the keyword dropdown.

This will build on #595 and may require a one-to-many term mapping, depending on the approach to #595.

**ACCEPTANCE CRITERIA**
- [x] Bibdata has a list of terms that trigger the addition of "Indigenous Studies"
- [x] At index time, this term gets added to lc_subject_facet, lc_subject_display, and subject search fields
- [x] I can search for "Indigenous Studies" in the "Subject (Keyword)" drop down and retrieve these records
- [x] I can use the "Subject > Topic" facet to limit to "Indigenous Studies" and retrieve these records
- [x]  Add a flag CHANGE_THE_SUBJECT. If there is this env variable then this code will change.
- [x] Test this in staging

minjiec:
Background information of the 2400~ terms in the spreadsheet:
- The terms are extracted from the LCSH database
- are about "the indigenous peoples of the Western Hemisphere, including Eskimos"
- but restricted to those organized hierarchically under the umbrella heading "Indians"
- and go only two levels deep.

Related projects:
1.	A [count](https://docs.google.com/spreadsheets/d/1CoJdjimDBP_Iys1_xW1aw0d6EolJ7Sd0rRqBgMKoyUw/edit#gid=971920615) of how many titles PUL has for each subject heading (left-match) as of August 2021
[...]

PR #1676, opened 2021-09-28
Augment subjects with Indigenous Studies where appropriate

bess:
Using a list of subject terms provided by PUL catalogers, at index time
add the term "Indigenous Studies" where appropriate.

Work toward #1548 

Screen Shot 2021-09-29 at 11 25 41 AM


christinach:
It looks great. We exported a fixture form Alma and tested locally the changes. The subject keyword search works and the show page display the desired term. I will approve the pr but it should not be merged before https://github.com/pulibrary/orangelight/issues/2755 is done and tested on staging.

christinach (approved):

christinach (changes requested):
After indexing some records in staging we noticed that there needs to be a match on the 650a subfield.

PR #1707, opened 2021-10-15
Logging for liberate::full rake task

hectorcorrea:
Logs whether a new reindex was started or not when running `liberate::full` rake task.

This will help to diagnose when running `rake liberate::full` does not queue any new jobs like the issue that @christinach and I ran into last night.

### DIFF
--- a/app/models/index_manager.rb
+++ b/app/models/index_manager.rb
@@ -9,11 +9,12 @@ class IndexManager < ActiveRecord::Base
   def self.reindex!(solr_url: nil)
     solr_url ||= rebuild_solr_url
     manager = self.for(solr_url)
-    return if manager.in_progress?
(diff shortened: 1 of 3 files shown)


app/models/index_manager.rb
```diff
@@ -9,11 +9,12 @@ def self.for(solr_url)
   def self.reindex!(solr_url: nil)
     solr_url ||= rebuild_solr_url
     manager = self.for(solr_url)
-    return if manager.in_progress?
+    return false if manager.in_progress?
```
tpendragon, line 12: Seems fine, but if you're depending on this return value you should add a test for it.

tpendragon: Whoop, I just missed it.

PR #1729, opened 2021-11-01
Adds a stricter validation to detect OCLC numbers

### LINKED ISSUE #1708: Re-Examine OCLC Number Normalization Rules
We should make the oclc normalization routine that drives the related titles feature. Example problem records:

https://catalog.princeton.edu/catalog/9912405253506421 has OCLC Number (OCoLC)ocm00112267

https://catalog.princeton.edu/catalog/9970335923506421 has OCLC Number (OCoLC)TGPSM11-B2267

both get normalized to "112267" which leads to the incorrect match. @mzelesky says the second number if bad data from a vendor loaded record, but we should make the routine more robust and numbers that don't really follow a valid OCLC number pattern should not be considered at all for OCLC matching purposes. Since the related titles look up also driving our electronic linking features that bring e-links from equivalent titles into record pages this has cascading problems in that space which you also see here since an erroneous e-link is display on 9970335923506421 because of this.

hectorcorrea:
Fixes #1708 

The code now checks for more than "string starts with (OCoLC)" and makes sure the format adheres better to the OCLC number specification.

### DIFF
--- a/marc_to_solr/lib/princeton_marc.rb
+++ b/marc_to_solr/lib/princeton_marc.rb
@@ -141,9 +141,9 @@ def other_versions record
     field.subfields.each do |s_field|
       linked_nums << StdNum::ISBN.normalize(s_field.value) if (field.tag == "020") || ((field.tag == "776") && (s_field.code == 'z'))
       linked_nums << StdNum::ISSN.normalize(s_field.value) if (field.tag == "022") || ((field.tag == "776") && (s_field.code == 'x'))
-      linked_nums << oclc_normalize(s_field.value, prefix: true) if s_field.value.start_with?('(OCoLC)') && (field.tag == "035")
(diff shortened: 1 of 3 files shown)


christinach:
@hectorcorrea I reindexed in staging using your branch:
 tested https://catalog-staging.princeton.edu/catalog/9970335923506421/raw vs
https://catalog.princeton.edu/catalog/9970335923506421/raw. With your branch the oclc_s is not indexed as expected. Your pr adds the stricter oclc rule but does not fix the issue with the unrelated links. We might to break issue #1708 in two issues. I will add it in the agenda for the orangelight/bibdata meeting today.

christinach:
https://catalog-staging.princeton.edu/catalog/9912405253506421 it looks ok now.

christinach (approved):

PR #1742, opened 2021-12-06
Retrieve firestone$ssrcfo and firestone$ssrcdc locations

### LINKED ISSUE #1741: Add Two New Holding Locations
Two new firestone locations have been created, we need to add them to the holding_locations. They are non-circulating items so are not available for pick-up service. 

firestone$ssrcfo
firestone$ssrcdc

christinach:
closes #1741


kevinreiss (approved):
@christinach this looks correct to me.

carolyncole (approved):
Will ticket why we need extra code...

PR #1751, opened 2021-12-18
1750 primary language field

### LINKED ISSUE #1750: Create new field for the primary language.
Create a new language_iana_primary_s to use in the 'lang' attribute in the h1 element of the record page. Following a discussion in orangelight with @tventimi and Lia Contursi https://pulibrary.slack.com/archives/CA891BNR0/p1639681431249700, @mzelesky and I drafted the logic for this new field so that we get one primary language and not more than one like we do in the language_iana_s field https://github.com/pulibrary/bibdata/blob/522fe147f05bc3e5cfdfba0bbcf3e73ba8a78b24/marc_to_solr/lib/traject_config.rb#L726:
per @mzelesky : 
008 35-37 should be used unless ###,zxx, mul, sgn, und, |||
if 008 is one of those (###,zxx, mul, sgn, und, |||) values then look first 041 field and first $a, 
if there is no subfield $a then use $d
Apply the same logic for 041 and exclude values (###,zxx, mul, sgn, und, |||)
The default is english. 

_Update_: _[aaa] is any other alphanumeric code that will be in the field; this is used._

christinach:
closes #1750
advances https://github.com/pulibrary/orangelight/issues/1129


marc_to_solr/lib/traject_config.rb
```diff
@@ -735,6 +738,32 @@
...
+  accumulator << primary_language_result
+  # 008 35-37 should be used unless ###,zxx, mul, sgn, und, |||
+  # \s{3,} three spaces
+  # if 008 is one of those (###,zxx, mul, sgn, und, |||) values then look first 041 field and first $a,
+  # if there is no subfield $a then use $d
+  # Apply the same logic for 041 and exclude values (###,zxx, mul, sgn, und, |||)
+  # The default is english.
+  # There can also be multiple 041a subfields
+  # Mark thinks the old cataloging style allowed for multiple language codes to be all put into one $a. sth like :freengita update the 041 method with scan(/.{1,3}/)
+end
```
christinach, line 765: this can also be in 041d

christinach:
closing in favor of https://github.com/pulibrary/bibdata/pull/1828

PR #1770, opened 2022-02-01
Code to allow for suppressed records to be shown for the scsb endpoint

### LINKED ISSUE #1745: The barcode/[barcode]/scsb route should return a record even if it is suppressed
https://github.com/pulibrary/bibdata/blob/27ce102fa72468467732dfe5d236d3d76b7b0d06/app/controllers/barcode_controller.rb#L22

Many records in ReCAP are suppressed. This route should not be checking to see if the bib is suppressed.

carolyncole:
fixes #1745


kevinreiss (approved):
This looks good to me. @mzelesky any comments?

mzelesky:
Looks good. I like that by default it still only returns unsuppressed records, but it makes an exception for the barcode route.

PR #1787, opened 2022-02-15
stop cron for campus access

### LINKED ISSUE #1776: Stop the cron for campus access
https://github.com/pulibrary/bibdata/blob/main/config/schedule.rb#L42

christinach:
closes #1776

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -38,11 +38,6 @@ every 1.day, at: "6:00am", roles: [:cron_production] do
   rake "marc_liberation:partner_update", output: "/tmp/cron_log.log"
 end
 
-# process the access file daily at 10:30am Eastern (the machine is in Z time, which is why this is 2pm)
-every 1.day, at: "2:30pm", roles: [:hr_cron] do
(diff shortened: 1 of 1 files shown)


config/schedule.rb
```diff
@@ -39,9 +39,9 @@
 end
 
 # process the access file daily at 10:30am Eastern (the machine is in Z time, which is why this is 2pm)
-every 1.day, at: "2:30pm", roles: [:hr_cron] do
-  rake "campus_access:load", output: "/tmp/campus_access_log.log"
-end
+# every 1.day, at: "2:30pm", roles: [:hr_cron] do
```
carolyncole, line 42: Should we just delete the lines?

carolyncole (changes requested):

config/schedule.rb
christinach: yes I will delete them. I wasnt sure if we would want to put it back at some point. but we can always check the history

carolyncole (approved):

PR #1793, opened 2022-02-23
Update indexing to include scsb in browse lists

### LINKED ISSUE #1790: In traject include scsb records in call_number_browse_s and call_number_display
We reviewed with @mzelesky traject and we need to consider scsb records for call_number_browse_s and call_number_display. This will allow for keyword search of the call _numbers and will allow for call_number browse.

carolyncole:
fixes #1790

### DIFF
--- a/marc_to_solr/lib/princeton_marc.rb
+++ b/marc_to_solr/lib/princeton_marc.rb
@@ -641,8 +641,32 @@ def alma_code_start_53?(code)
   code.to_s.start_with?("53") && code.to_s.end_with?("06421")
 end
 
+def scsb_code_start?(code)
+  code.to_s.start_with?("scsb")
+end
+
 def alma_852(record)
   record.fields('852').select { |f| alma_code_start_22?(f['8']) }
+end
+
+def scsb_852(record)
+  record.fields('852').select { |f| scsb_code_start?(f['b']) }
+end
+
+def browse_fields(record, khi_key_order: ['k', 'h', 'i'])
+  result = []
+  fields = if scsb_doc?(record['001']&.value)
+             scsb_852(record)
+           else
+             alma_852(record)
+           end
+  fields.each do |field|
+    subfields = call_number_khi(field)
+    next if subfields.empty?
+    values = [field[khi_key_order[0]], field[khi_key_order[1]], field[khi_key_order[2]]].compact.reject(&:empty?)
+    result << values.join(" ") if values.present?
+  end
+  result
 end
 
 def alma_876(record)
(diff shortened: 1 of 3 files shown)


spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -251,7 +251,10 @@ def fixture_record(fixture_name)
       expect(@record_call_number2['call_number_browse_s']).to eq(["CD- 50000"])
     end
     it 'returns the enrichment call_number_browse with k at the end' do
-      expect(@sample41['call_number_browse_s']).to eq(["GA 2015.00160 A Middle 30/Drawer 11/GC024/Full Folio/20th-21st c./Artists A"])
+      expect(@sample41['call_number_browse_s']).to contain_exactly("GA 2015.00160", "GA 2015.00160 A Middle 30/Drawer 11/GC024/Full Folio/20th-21st c./Artists A")
```
carolyncole, line 254: @christinach I believe the `GA 2015.00160` being repeated here is unexpected: https://catalog.princeton.edu/catalog/99106471643506421 There is only one item.  It does not have an 8 field, but neither does the SCSB record.

marc_to_solr/lib/princeton_marc.rb
```diff
@@ -641,10 +641,18 @@ def alma_code_start_53?(code)
   code.to_s.start_with?("53") && code.to_s.end_with?("06421")
 end
 
+def scsb_code_start?(code)
```
carolyncole, line 644: I had to add this filter of we ended up getting another call number for https://catalog.princeton.edu/catalog/99106471643506421, which is @sample41 in the tests.

marc_to_solr/lib/princeton_marc.rb
```diff
@@ -641,10 +641,18 @@ def alma_code_start_53?(code)
...
+  code.to_s.start_with?("scsb")
+end
+
 def alma_852(record)
   record.fields('852').select { |f| alma_code_start_22?(f['8']) }
 end
 
+def alma_and_scsb_852(record)
+  record.fields('852').select { |f| alma_code_start_22?(f['8']) || scsb_code_start?(f['b']) }
+end
```
christinach, line 654: We have a check for scsb in princeton_marc, which is to check the value of the document id. SCSB starts with `SCSB-`, `scsb_doc?(record['001'].value)`. 
@mzelesky is it a robust check using subfield $b to check for `scsb` text? If yes then I think it's ok to have one more check besides the id.

mzelesky: I think it's good to check the subfield b of an 852 for SCSB records. However, I think that the 'alma_and_scsb_852' needs to be separate checks. So, like we have 'alma_852', we should also have 'scab_852'. Then, if the record is identified as a SCSB record, go the scsb_852 route. If the record is identified as an All record, go the alma_852 route.

carolyncole: @mzelesky is it possible to have a record that has both recap and non recap 852 fields?  I was unsure, which is why I put them together.

christinach (approved):

christinach (approved):
@carolyncole thank you

PR #1794, opened 2022-02-24
Document how to get past bad jobs

carolyncole:


docs/bad_dump
```diff
@@ -0,0 +1,8 @@
+1. sssh worker box
```
sandbergja, line 1: sssh should be ssh

sandbergja (commented):
Thank you for these, Carolyn!  Should this have the .md extension so it displays nicer in Github?

carolyncole:
Thanks @sandbergja for the review! @christinach I fixed up a couple changes @sandbergja asked about.  Feel free to rearrange when adding in the actual commands.

christinach (dismissed):

docs/bad_dump.md
```diff
@@ -0,0 +1,12 @@
+1. `ssh deploy@bibdata-alma-worker1`
+1. `cd /opt/marc_liberation/current`
+1. `RAILS_ENV=production bundle exec rails c`
+1. Find the IndexManager instance that uses the solr production collection. Currenlty is:`catalog-alma-production`
+`index_manager_current = IndexManager.all.first`
+1. `index_manager_current.solr_collection` => `http://lib-solr8-prod.princeton.edu:8983/solr/catalog-alma-production`
```
carolyncole, line 6: @christinach why is this a step?  What is it doing?

christinach:
@sandbergja can you re-review?

docs/bad_dump.md
christinach: its not a step. You make sure that you are checking the correct index manager instance by checking the solr_collection. I'm sorry I added it as a step.

christinach: there are two instances one has as a solr_collection: catalog-alma-production-rebuild but it is not in production.

christinach: I will update it

docs/bad_dump.md
```diff
@@ -0,0 +1,12 @@
+1. `ssh deploy@bibdata-alma-worker1`
+1. `cd /opt/marc_liberation/current`
+1. `RAILS_ENV=production bundle exec rails c`
+1. Find the IndexManager instance that uses the solr production collection. Currenlty is:`catalog-alma-production`
+`index_manager_current = IndexManager.all.first`
```
christinach, line 5: We should update this. We need to find the current index manager which is in production. It might not be the first.

sandbergja (approved):

PR #1797, opened 2022-02-25
Capture UTF8 errors and ignore them

carolyncole:
Also updated to traject 3.7 from 2.3.1 to get better error handeling options

### DIFF
--- a/spec/jobs/alma_dump_transfer_job_spec.rb
+++ b/spec/jobs/alma_dump_transfer_job_spec.rb
@@ -75,7 +75,7 @@ RSpec.describe AlmaDumpTransferJob, type: :job do
 
         expect(dump.dump_files.count).to eq 2
         expect(dump.dump_files.map(&:dump_file_type).map(&:constant).uniq).to eq ["BIB_RECORDS"]
-        expect(dump.dump_files.first.path).to eq File.join(MARC_LIBERATION_CONFIG['data_dir'], filename1)
+        expect(dump.dump_files.map(&:path)).to contain_exactly(File.join(MARC_LIBERATION_CONFIG['data_dir'], filename1), File.join(MARC_LIBERATION_CONFIG['data_dir'], filename2))
 
         expect(IndexRemainingDumpsJob).not_to have_received(:perform_async)
       end
--- a/spec/marc_to_solr/lib/config_spec.rb
+++ b/spec/marc_to_solr/lib/config_spec.rb
@@ -928,7 +928,7 @@ describe 'From traject_config.rb' do
     it "ignores errors and allows the indexer to continue" do
       indexer = IndexerService.build
       sample = indexer.map_record(fixture_record('99125119454006421', indexer: indexer))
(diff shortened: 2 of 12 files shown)


spec/jobs/alma_dump_transfer_job_spec.rb
```diff
@@ -75,7 +75,7 @@
 
         expect(dump.dump_files.count).to eq 2
         expect(dump.dump_files.map(&:dump_file_type).map(&:constant).uniq).to eq ["BIB_RECORDS"]
-        expect(dump.dump_files.first.path).to eq File.join(MARC_LIBERATION_CONFIG['data_dir'], filename1)
+        expect(dump.dump_files.last.path).to eq File.join(MARC_LIBERATION_CONFIG['data_dir'], filename1)
```
carolyncole, line 78: @christinach the order of the files changed here.  I am unsure if this order is important to processing.  If it is then I can work to correct the order

spec/system/devise_spec.rb
```diff
@@ -25,12 +25,15 @@
...
 
+    scenario "unauthenticated users can not delete events" do
+      visit "/events"
+      expect(page).not_to have_link "Delete"
+    end
+
     scenario "only authenticated users can delete events" do
+      sign_in FactoryBot.create(:admin), scope: :user
       visit "/events"
-      accept_alert do
```
carolyncole, line 30: This test was failing locally and not in CI.  I do not think this was due to the upgrade, but I fixed it anyway.

spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -4,862 +4,874 @@
   let(:leader) { '1234567890' }
   let(:online) { @indexer.map_record(fixture_record('9990889283506421')) }
 
-  def fixture_record(fixture_name)
+  def fixture_record(fixture_name, indexer: @indexer)
     f = File.expand_path("../../../fixtures/marc_to_solr/#{fixture_name}.mrx", __FILE__)
-    @indexer.reader!(f).first
```
carolyncole, line 9: I created a context for valid records, so all of these lines are still part of the file.  They were just indented.  The diff looks much larger because of this.  This allowed me to test my invalid record without having to load all the valid ones.

spec/jobs/alma_dump_transfer_job_spec.rb
carolyncole: I seems that this order changes based on the machine the test is run on.  CI still sees file1 first.  I changed it to contain_exactly, but can look into the order.

carolyncole:
As we discussed in standup, we will wait to update traject until after we do the full reindex next week.

christinach (approved):
@carolyncole thank you it looks great. I will also trigger an incremental job in the alma sandbox to test it in bibdata-staging.

christinach:
The indexing test process completed successfully until the last dump in the events page with dump_id=700.

PR #1814, opened 2022-03-04
Adding qa configuration for capistrano

carolyncole:
Also allowing the application name to be updated per server


kevinreiss (approved):
Good to merge if the machine names are current.

PR #1828, opened 2022-03-23
Updates iana field to exclude invalid values and when nil to default to "en"

christinach:


christinach (commented):
the new tests are https://github.com/pulibrary/bibdata/pull/1828/files#diff-c06309004557ab56e2f1de7a00a776f2e78c19a097462cdfa800c3435076b2fbR307-R330

The rest of the config_spec is updated because I noticed that there was a duplication and a nested test; I fixed it. I must have messed this up in the past when I rebased the temp locations pr.

marc_to_solr/lib/traject_config.rb
```diff
@@ -736,16 +736,18 @@
...
 
+# # import PrimaryLanguage method from ./primary_laguage.rb
 to_field 'language_iana_s', extract_marc('008[35-37]:041a:041d') do |_record, accumulator|
-  codes = accumulator.compact.map { |c| c.length == 3 ? c : c.scan(/.{1,3}/) }.flatten.uniq
-  codes = codes.reject { |c| ISO_639.find(c).nil? }.map do |c|
-    if ISO_639.find(c).alpha2.empty?
-      c
+  codes = accumulator.compact.map { |m| m.length == 3 ? m : m.scan(/.{1,3}/) }.flatten.uniq
+  codes_iso_639 = codes.reject { |m| ISO_639.find(m).nil? }.map do |m|
+    if ISO_639.find(m).alpha2.empty? || ["zxx", "mul", "sgn", "und", "|||"].include?(m)
```
carolyncole, line 743: Where do there codes in the include list come from?  It would be nice to understand why you are ignoring them.  Maybe it is as simple as naming them?

spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -731,6 +758,20 @@ def fixture_record(fixture_name, indexer: @indexer)
       end
     end
 
+    describe 'series 490 dedup, non-filing' do
+      let(:s490) { { "490" => { "ind1" => "", "ind2" => " ", "subfields" => [{ "a" => "Series title" }] } } }
+      let(:s830) { { "830" => { "ind1" => "", "ind2" => " ", "subfields" => [{ "a" => "Series title." }] } } }
+      let(:s440) { { "440" => { "ind1" => "", "ind2" => "4", "subfields" => [{ "a" => "The Series" }] } } }
+      let(:record) { @indexer.map_record(MARC::Record.new_from_hash('fields' => [s490, s830, s440], 'leader' => leader)) }
```
carolyncole, line 765: 🎉 this is a cool way to create test data!

carolyncole (changes requested):
@christinach just one small question about `["zxx", "mul", "sgn", "und", "|||"]`. Otherwise looks good.

sandbergja (approved):

marc_to_solr/lib/traject_config.rb
christinach: @carolyncole good point. I looked with @mzelesky case by case and listed the values that should be excluded based on the section `35-37 - Language` in https://www.loc.gov/marc/bibliographic/bd008a.html. I will add a comment. Some of them are excluded from the ISO-639 gem but some are not. This is why I listed all of them so that I make sure I ignore the invalid ones.

christinach:
@carolyncole can you check again? I added a comment.

carolyncole (approved):

PR #1875, opened 2022-05-20
Ensure empty terms do not raise errors

### LINKED ISSUE #1829: Allow the Change the subject to also allow just masking
Instead of substituting the value we would like to also just hide certain values.

maxkadel:
- Use environment for configuration to allow for test fixtures

Fixes #1829


carolyncole (approved):

carolyncole:
I feel like there should be a better way to add a term to the config for testing.  There is not a good seam for adding a term though.  Looking good!

PR #1878, opened 2022-05-26
Indicators for the new 876 field should be spaces

### LINKED ISSUE #1877: Indicators for the new 876 field should be spaces instead of zeroes
https://github.com/pulibrary/bibdata/blob/4c2f7b2df482af4ce3e9508683c1e3454601a48c/app/adapters/alma_adapter/alma_item.rb#L72

christinach:
closes #1877


mzelesky (approved):
This looks like a simple fix that will create valid MARC fields.

PR #1879, opened 2022-05-26
Update method linked_record_ids to check for field[t] and that is an alma id

### LINKED ISSUE #1876: Make this method more efficient
https://github.com/pulibrary/bibdata/blob/4c2f7b2df482af4ce3e9508683c1e3454601a48c/app/adapters/alma_adapter/marc_record.rb#L21

This method should become

```
    # retrieves constituent IDs from the record
    def linked_record_ids
      linked_record_fields = marc_record.fields('774').select do |field|
        field['w'] =~ /^99[0-9]+6421$/ && field['t']
      end
     linked_record_fields.map { |field| field['w'] }
    end
```

I believe there is already a test for Alma MMS IDs, which can be used in place of the regex.

christinach:
Update comment in barcode controller to remove suppressed.

closes #1876

### DIFF
--- a/app/adapters/alma_adapter/marc_record.rb
+++ b/app/adapters/alma_adapter/marc_record.rb
@@ -19,8 +19,8 @@ class AlmaAdapter
     end
 
     def linked_record_ids
-      linked_record_fields = marc_record.fields.select do |field|
-        field.tag == "774"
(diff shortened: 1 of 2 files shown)


mzelesky (approved):
Looks good.

Issue #1881, opened 2022-06-01
Seeding figgy cache in development seems to block indexing to Solr


maxkadel:
Trying to add a couple fixture objects to Solr in development kicks off seeding the cache for figgy, which runs for a very long time, and seems to be getting the seed for all the production data, which is not needed in development. 

Reproduction:
* Bring up bibdata solr - `bundle exec rake servers:start` 
* Copy the url for lando's development solr
* Using a fixture xml with MARC data, run 
```bash
traject -c marc_to_solr/lib/traject_config.rb PATH/TO/FILE.xml -u LANDO_SOLR_URL/solr/marc-liberation-core-development -w Traject::JsonWriter
```

christinach:
We should include the above in the documentation.

christinach:
documentation was added

PR #1908, opened 2022-06-28
Document how to deal with FIGGY_ARK_CACHE_PATH in development

maxkadel:
Connected to https://github.com/pulibrary/bibdata/issues/1881 - change based on comment on that issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,11 +126,9 @@ One may seed the cache using the following Rake Task:
 rake liberate:arks:seed_cache
 ```
 
-In development, when running traject command, set the `FIGGY_ARK_CACHE_PATH` to point to `spec/fixtures/marc_to_solr/figgy_ark_cache` in the local environment. e.g. 
+In development, when running commands that utilize the cache, such as commands indexing via traject, set the `FIGGY_ARK_CACHE_PATH` to point to `spec/fixtures/marc_to_solr/figgy_ark_cache` in the local environment.
 ```bash
 export FIGGY_ARK_CACHE_PATH=spec/fixtures/marc_to_solr/figgy_ark_cache
-bundle exec traject -c marc_to_solr/lib/traject_config.rb pathToFile -u http://localhost:solr-port/solr/marc-liberation-core-development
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -126,6 +126,13 @@ One may seed the cache using the following Rake Task:
 rake liberate:arks:seed_cache
 ```
 
+In development, it may be best to copy the cache from a production or production-like environment into `spec/fixtures/marc_to_solr/figgy_ark_cache/` and, when running traject command, set the `FIGGY_ARK_CACHE_PATH` in the local environment. e.g.
```
christinach, line 129: In order to index in the dev environment, we don't need to copy all the time the figgy_ark_cache from a production box. What we have right now in `spec/fixtures/marc_to_solr/figgy_ark_cache/` should work. If it doesn't work then copy from prod the `figgy_ark_cache`.

christinach (changes requested):

README.md
maxkadel: There is only a single file in `spec/fixtures/marc_to_solr/figgy_ark_cache/` - is that what you're expecting to see there? The last time I tried with only that single file it did not seem to work - my local environment started spending a *lot* of time re-building the cache.

README.md
```diff
@@ -126,6 +126,13 @@ One may seed the cache using the following Rake Task:
 rake liberate:arks:seed_cache
 ```
 
+In development, when running traject command, set the `FIGGY_ARK_CACHE_PATH` to point to `spec/fixtures/marc_to_solr/figgy_ark_cache` in the local environment. e.g. 
+```bash
+export FIGGY_ARK_CACHE_PATH=spec/fixtures/marc_to_solr/figgy_ark_cache
+bundle exec traject -c marc_to_solr/lib/traject_config.rb pathToFile -u http://localhost:solr-port/solr/marc-liberation-core-development
```
christinach, line 132: @maxkadel I dont think we need the traject command to index records in the section Seeding the cache. I just viewed the full file and realized that its in this section. Can we remove lines 132 and 133?

christinach (approved):

PR #1928, opened 2022-07-27
Augment The Subject refactor

### LINKED ISSUE #1863: Update Augment the Subject List with new Recommendations
Per Minjie Chen there is an updated list of terms we should use for the augment the subject list for indigenous studies. 

Updated term list is at: https://docs.google.com/spreadsheets/d/1jFsLExGilXWmXEikliuYUtdipF5HnRbNMZn-rgJYflY/edit#gid=1600276075

Add these to this file: https://github.com/pulibrary/bibdata/blob/main/marc_to_solr/lib/augment_the_subject/indigenous_studies.txt

Once added if it is possible we should do a selective publishing job in Alma that will force any records containing these terms through the updated term list. If that isn't practical given the large number of new terms we will have to catch up with a full re-index.

maxkadel:
- Load new csv for augment the subject
- Create list of terms that only require subfield a to match
- Create list of terms that only require subfield x to match
- Create list of terms that require both subfield a and specific lists of subfields to match
- Add rake task to re-create fixtures

Remove deprecation warning for URI.escape

- I believe we are only escaping / unescaping urls from trusted sources. If not, we might want a different approach.

Closes #1863 

Co-authored-by: Ryan Laddusaw <rladdusaw@users.noreply.github.com>


marc_to_solr/lib/augment_the_subject.rb
```diff
@@ -5,35 +5,55 @@
 ##
 # The creation and management of metadata are not neutral activities.
 class AugmentTheSubject
-  LCSH_TERMS_FILE = File.join(File.dirname(__FILE__), 'augment_the_subject', 'indigenous_studies.txt')
-  LCSH_SUBFIELDS_FILE = File.join(File.dirname(__FILE__), 'augment_the_subject', 'indigenous_studies_subfields.json')
+  LCSH_TERMS_CSV_FILE = File.join(File.dirname(__FILE__), 'augment_the_subject', 'indigenous_studies.csv')
+  # Can be re-created using `bundle exec rake augment:recreate_fixtures`
+  LCSH_STANDALONE_A_FILE = File.join(File.dirname(__FILE__), 'augment_the_subject', 'standalone_subfield_a.json')
+  # Must be created by hand from file provided by metadata librarians
```
christinach, line 11: What exactly are we creating by hand? can we add some instructions?

marc_to_solr/lib/augment_the_subject.rb
```diff
@@ -72,40 +95,70 @@ def indigenous_studies?(terms)
...
+    subfield_a = normalize(term.split(SEPARATOR).first).gsub(/\.$/, '')
+    standalone_subfield_a_terms.include?(subfield_a)
   end
 
   ##
-  # For some subject terms, we need to match on all subfields
-  # E.g., Russia (Federation)-Civilization-Indian influences
-  def full_subject_term_match?(term)
-    indigenous_studies_terms.include? term.downcase.gsub(/[^\w\s#{SEPARATOR}]/, '')
+  # For some subfield terms, only a single subfield needs to match.
```
maxkadel, line 103: Since the standalone subfield x's were in a separate, small worksheet on Google docs, we created the json version of this file manually. If this list becomes larger, we may need to revisit this process.

christinach (dismissed):
@maxkadel @rladdusaw thank you for this work!

Issue #1925, opened 2022-07-22
Index homosaurus subject headings ($2homoit)


sandbergja:
User story 1: as a catalog user, I would like to search for records using terms in the homosaurus vocabulary.
User story 2: as a catalog user, I would like to see homosaurus vocabulary terms display in record show pages.

homosaurus vocabulary terms can be identified in the marc record by the combination of:

- Tag 650
- Second indicator 7
- Subfield 2 with a value "homoit"

@kevinreiss is going to get some more clarity and prioritization of this ticket.

This will probably require a full re-index.

Companion ticket in Orangelight: https://github.com/pulibrary/orangelight/issues/3078 -- we may also need a ticket in pul_solr to weight this search field appropriately.

kevinreiss:
Example record in production: https://catalog.princeton.edu/catalog/99125407041106421

christinach:
@mzelesky suggested to add in bibdata the `homoit_s` so that we indicate in the record page the source of the term. previous example for lcgft_shttps://github.com/pulibrary/bibdata/pull/471/files. 
@mzelesky also suggests to revisit https://github.com/pulibrary/orangelight/pull/3029 as part of this work.

PR #1937, opened 2022-08-02
Index Homosaurus terms to homoit_s

maxkadel:
This vocabulary follows the pattern set by the Siku and local subject headings.

Connected to #1925

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -840,12 +840,19 @@ to_field 'local_subject_display' do |record, accumulator|
   accumulator.replace(subjects)
 end
 
-# Adds lc and siku subject unstem_search fields
+# Homosaurus controlled vocabulary
+to_field 'homoit_subject_display' do |record, accumulator|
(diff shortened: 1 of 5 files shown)


maxkadel:
Is it correct that this is only adding the terms to a single Solr field? Should we be adding them to any of the other subject fields for search?

sandbergja:
@maxkadel that's what the `homoit_subject_unstem_search` should be for.  It will just need to be configured in the [solrconfig](https://github.com/pulibrary/pul_solr/blob/main/solr_configs/catalog-production-v2/conf/solrconfig.xml) as well via a PR to the pul_solr repo.

marc_to_solr/lib/traject_config.rb
```diff
@@ -840,20 +840,27 @@
   accumulator.replace(subjects)
 end
 
-# Adds lc and siku subject unstem_search fields
+# Homosaurus controlled vocabulary
```
christinach, line 843: @maxkadel great! can you include in the comment the link to the specific vocabulary? https://homosaurus.org/

maxkadel: Done!

sandbergja (approved):
Thanks @maxkadel

PR #1940, opened 2022-08-05
Reorganized the README into docs files

### LINKED ISSUE #1914: Organize the documentation in the docs file
We should try to keep in the readme how to set up and run the application and move the different scenarios and other information in individual files in the `docs` directory. 
We could use a link from the Readme to a docs/file_documentation if it's needed to be referenced. 
This will make the README easier to read and navigate.

rladdusaw:
Resolves #1914


README.md
```diff
@@ -48,131 +44,6 @@ bundle install
...
-
-One may seed the cache using the following Rake Task:
-```bash
-rake liberate:arks:seed_cache
-```
-
-In development, when running commands that utilize the cache, such as commands indexing via traject, set the `FIGGY_ARK_CACHE_PATH` to point to `spec/fixtures/marc_to_solr/figgy_ark_cache` in the local environment.
-```bash
-export FIGGY_ARK_CACHE_PATH=spec/fixtures/marc_to_solr/figgy_ark_cache
-```
```
sandbergja, line 144: It would be nice to keep these steps in the README, since they can trip up setting a dev environment.

README.md
```diff
@@ -48,131 +44,6 @@ bundle install
...
 1. To stop: `bundle exec rake servers:stop` or `lando stop`
 
-## Alma
-
-### Configure Alma keys for Development
-
-1. `lpass login emailhere`
-1. `bundle exec rake alma:setup_keys`
-
-This will add a .env with credentials to Rails.root
```
sandbergja, line 58: It would be good to keep these steps in the README, as part of setting up a dev environment.

sandbergja (changes requested):
Looks good, @rladdusaw !  Just two things that I think we should keep in the README so devs don't have to go digging through the docs to set up a good dev environment.

README.md
rladdusaw: Moved back to README

README.md
rladdusaw: Copied to README as I think it makes sense to have in both locations.

sandbergja (approved):
Thank you @rladdusaw

PR #1941, opened 2022-08-05
Remove hours_locations (never fully implemented)

### LINKED ISSUE #1165: Remove the Hours Location feature
https://bibdata.princeton.edu/locations/hours_locations

The Hours Location is a nice feature that we never actually implemented and is not used in the catalog. 
Do we want to implement it? If not can we remove it?

maxkadel:
closes #1165


sandbergja (approved):
thanks @maxkadel

PR #1943, opened 2022-08-08
Temp location cleanup

### LINKED ISSUE #1802: refactor for PR were skipped
comments in this PR were not taken into account https://github.com/pulibrary/bibdata/pull/1734/files

We should do these refactors before we make additional changes to this code.

maxkadel:
Closes #1802

### DIFF
--- a/marc_to_solr/lib/process_holdings_helpers.rb
+++ b/marc_to_solr/lib/process_holdings_helpers.rb
@@ -46,8 +46,13 @@ class ProcessHoldingsHelpers
   end
 
   # Build the permanent location code from 852$b and 852$c
+  # Do not append the 852c if it is a SCSB - we save the SCSB locations as scsbnypl and scsbcul
   def permanent_location_code(field_852)
-    "#{field_852['b']}$#{field_852['c']}"
+    if alma?(field_852)
+      "#{field_852['b']}$#{field_852['c']}"
+    else
+      field_852['b']
+    end
   end
 
   # Select 876 fields (items) with permanent location. 876 location is equal to the 852 permanent location.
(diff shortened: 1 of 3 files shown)


marc_to_solr/lib/process_holdings_helpers.rb
```diff
@@ -73,54 +78,35 @@ def in_temporary_location(field_876, field_852)
...
-    if field_852['l']
-      holding_current['shelving_title'] ||= []
-      holding_current['shelving_title'] << field_852['l']
-    end
-    if field_852['z']
-      holding_current['location_note'] ||= []
-      holding_current['location_note'] << field_852['z']
+  def build_call_number(field_852)
+    call_number = []
+    call_number << [field_852['h'], field_852['i'], field_852['k'], field_852['j']].compact.reject(&:empty?)
```
sandbergja, line 83: can these two lines be combined?

Also, would `reject(&:blank?)` have the same effect as `.compact.reject(&:empty?)`

marc_to_solr/lib/process_holdings_helpers.rb
```diff
@@ -73,54 +78,35 @@ def in_temporary_location(field_876, field_852)
...
+    call_number.flatten!
+    return call_number.join(' ').strip if call_number.present?
+    []
```
sandbergja, line 86:
```suggestion
    call_number.present? ? call_number.join(' ').strip : []
```

sandbergja (changes requested):
Thanks @maxkadel !  Just a few small questions and readability suggestions.

marc_to_solr/lib/process_holdings_helpers.rb
```diff
@@ -154,24 +138,23 @@ def build_item(item:, field_852:, field_876:)
   end
 
   def process_866_867_868_fields(fields:, all_holdings:, holding_id:)
+    return all_holdings if all_holdings[holding_id].nil?
```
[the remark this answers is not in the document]
maxkadel: I got the idea from your review back in February. 

I think previously we were essentially doing the same thing, but it was just sort of being hidden in the `fields.each` loop where it says `next unless all_holdings[holding_id]`. 

The idea is that there's no need to loop through the fields if `all_holdings[holding_id]` is nil, since `all_holdings[holding_id]['location_has'] << parse_textual_holdings(field, '866')` would just error out. 

Looking into it more, it seems that sometimes `all_holdings` at this point is just an empty hash, which also leads to `all_holdings[holding_id]` being nil, which seems surprising. 

I think I'm going to try to refactor the `process_holdings` method in princeton_marc to be able to reason about it more easily.

marc_to_solr/lib/process_holdings_helpers.rb
maxkadel: Yup and Yup! And we can get rid of the `.flatten!`

marc_to_solr/lib/process_holdings_helpers.rb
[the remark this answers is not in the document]
maxkadel: I ended up moving this up to the `process_holdings` method, since it doesn't make sense to run the `process_866_867_868_fields` method unless `all_holdings` has the `holding_id` as a key

sandbergja (approved):
Thanks for this refactor, @maxkadel !

PR #1953, opened 2022-08-18
Add rake task to request full records from SCSB

### LINKED ISSUE #1544: Rake task to request full partner records from SCSB API
It would remove a bottleneck on Kevin to have this as a rake task. 

The API accepts an email address to report when the export to s3 is finished. Maybe use the drds libanswers email address.

maxkadel:
@kevinreiss  - I'm I right in thinking that the response:

```
Can't run export now as another process is in progress. We have saved your request.We will initiate and notify once the exiting process completes
```

means that requests after the first one are in the queue and will be processed once the first one is done?

If I'm mistaken, we should change the rake task to only request from one institution at a time, which could be passed in from the environment as well.

Closes #1544

### DIFF
--- a/app/models/scsb_full_records_request.rb
+++ b/app/models/scsb_full_records_request.rb
@@ -36,7 +36,7 @@ class ScsbFullRecordsRequest
       req.headers['api_key'] = scsb_auth_key
     end
     expected_response_body = "Export process has started and we will send an email notification upon completion"
-    Rails.logger.error("Receved unexpected response: #{response.body}") unless response.body == expected_response_body
+    Rails.logger.error("Received unexpected response: #{response.body}") unless response.body == expected_response_body
     response
   rescue Faraday::ConnectionFailed, Faraday::TimeoutError => connection_failed
     Rails.logger.warn("#{self.class}: Connection error for #{scsb_server}")
--- a/app/models/scsb_full_records_request.rb
+++ b/app/models/scsb_full_records_request.rb
@@ -27,7 +27,7 @@ class ScsbFullRecordsRequest
       req.params['collectionGroupIds'] = '1,2'
       req.params['emailToAddress'] = email
       req.params['fetchType'] = 10
-      req.params['imsDepositoryCodes'] = 'RECAP,HD'
+      req.params['imsDepositoryCodes'] = repository_codes(institution_code)
       req.params['institutionCodes'] = institution_code
       req.params['outputFormat'] = 0
       req.params['requestingInstitutionCode'] = 'PUL'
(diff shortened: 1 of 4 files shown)


app/models/scsb_full_records_request.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  def scsb_request(institution_code)
+    response = scsb_conn.get do |req|
+      req.path = '/dataDump/exportDataDump'
+      req.params['collectionGroupIds'] = '1,2'
+      req.params['emailToAddress'] = email
+      req.params['fetchType'] = 10
+      req.params['imsDepositoryCodes'] = 'RECAP,HD'
```
kevinreiss, line 30: I wonder if this should be a parameter. Harvard is the only partner that has items in both facilities, for CUL and NYPL you'd only ask for "RECAP".

kevinreiss:
Yes, we would want to run the task one institution at a time. Yes, if make the calls for each partner close together you'll get the text response you posted. The application will start your job when the first partner dump requested concludes. Notifications will go to the email address passed in the API call.

kevinreiss:
I'm wondering if ultimately we'd want to create a scheduled task to generate a full dump from each partner say, every two weeks so we'll always have a relatively fresh copy available to us. @mzelesky and @pmgreen also both refer to this dataset for reporting and collection analysis reports so it might beneficial for them to have this data be available in a relatively current form.

kevinreiss (changes requested):

maxkadel:
> I'm wondering if ultimately we'd want to create a scheduled task to generate a full dump from each partner say, every two weeks so we'll always have a relatively fresh copy available to us. @mzelesky and @pmgreen also both refer to this dataset for reporting and collection analysis reports so it might beneficial for them to have this data be available in a relatively current form.

In that case, would this code more appropriately live in lib_jobs?

config/schedule.rb
```diff
@@ -42,3 +42,15 @@
 every 1.week, roles: [:cron_staging, :cron_production] do
   rake "marc_liberation:delete:events"
 end
+
+every 2.weeks, at: '7:00am', roles: [:cron_production] do
+  rake "scsb:request_records[production,mk8066@princeton.edu,CUL]", output: "/tmp/cron_log.log"
```
kevinreiss, line 47: Unfortunately the output from the API is pretty minimal, so having the email notifications is really the only way to get a record of what the job does. I'm thinking we should get a service account like Carolyn suggested to receive them so we don't have one our addresses hardcoded in the config.

kevinreiss (approved):

PR #1967, opened 2022-08-25
Update SCSB indexing docs

christinach:


docs/indexing.md
```diff
@@ -171,17 +171,14 @@ $ SET_URL=http://lib-solr8-prod.princeton.edu:8983/solr/catalog-alma-production-
 
 ### Index the latest SCSB changes
 
-There might be new SCSB updates from Alma between the time the SCSB export started and the index swap. Set the TIMESTAMP to be the next day of when the SCSB export started (as long as the export started after 6am EST). Index these latest SCSB changes:
-example: If the SCSB reindex started on '2021-10-15'. Set the TIMESTAMP= '2021-10-15'.
-
-SSH to the bibdata alma worker machine that is used for indexing https://github.com/pulibrary/bibdata/blob/7284a2364a8c1eb5af70f8e79b80a44eb546a4bc/config/deploy/production.rb#L11-L12 and start a tmux session.
-
+There will be new SCSB updates that need to be indexed. We import these updates to bibdata daily at 6:00am from an AWS S3 that belongs to SCSB.
+SSH to the bibdata alma worker machine that is used for indexing https://github.com/pulibrary/bibdata/blob/7284a2364a8c1eb5af70f8e79b80a44eb546a4bc/config/deploy/production.rb#L11-L12
```
maxkadel, line 175: This is pinned to a specific commit - do we want to link to "whatever's in the spot on main" or something?

https://github.com/pulibrary/bibdata/blob/main/config/deploy/production.rb

maxkadel (changes requested):
Thanks for updating the docs - I think pointing to an old commit to determine current state could be fragile

christinach:
> Thanks for updating the docs - I think pointing to an old commit to determine current state could be fragile

yes definitely. I missed that. I will update

maxkadel (approved):

PR #1975, opened 2022-09-01
Upgrade Honeybadger gem

### LINKED ISSUE #1904: Upgrade honeybadger gem
From the honeybadger UI:

Breadcrumbs are available for our Honeybadger Ruby client, but it looks like you need to upgrade first.

This notice was sent using version 3.3.1. We introduced Breadcrumbs in version 4.4.0. Be sure to upgrade and enable to send breadcrumb data along with your errors!

maxkadel:
Closes #1904


sandbergja (approved):
Thanks @maxkadel

PR #1982, opened 2022-09-07
Update location data service to populate location tables using config json; Remove voyager_to_alma_mapping file; Update templates;

### LINKED ISSUE #1241: Manage locations via a config file instead of using the database
This is part of #1103 
I'm able to update the locations and the flags using the voyager mapping file, the voyager data and the alma locations. 
I thought it would be good to dump everything locally in a yaml file and check it in the repo, so that I can upload them in the location tables anytime, locally or in the VM.

### LINKED ISSUE #1633: Remove voyager to alma mapping file
Removing the file check will simplify the way we add locations and populate the tables.
part of #1241

christinach:
closes #1241,
closes #1633


app/services/locations_to_file_service.rb
```diff
@@ -12,20 +12,27 @@ def initialize(base_path: Rails.root.join('config', 'locations'))
```
christinach, line 12: - LocationsToFileService should be deleted in a different PR so that we keep the history of the json files generation.

- The delivery_locations.json was updated manually with specific ids that figgy is using.

config/locations/delivery_locations.json
```diff
@@ -13,9 +13,8 @@
         "library": {
```
christinach, line 13: We don't use 'order' and this is why it was left 0 since we moved to Alma.

config/locations/libraries.json
```diff
@@ -0,0 +1,110 @@
...
+    {
+        "label": "Electronic Access",
+        "code": "online",
+        "order": 0,
+        "url": "https://bibdata.princeton.edu/locations/libraries/online.json"
+    },
+    {
+        "label": "Engineering Library",
+        "code": "engineer",
+        "order": 0,
```
christinach, line 23: We don't use 'order' and this is why it was left 0 since we moved to Alma.

spec/fixtures/files/alma/locations/delivery_locations.json
```diff
@@ -0,0 +1,393 @@
+[
+    {
+        "id": 1,
```
christinach, line 3: The ids in this file were added manually. Figgy is using these ids.

spec/services/location_data_service_spec.rb
```diff
@@ -215,19 +41,13 @@
       rare_scathx = HoldingLocation.find_by(code: 'rare$scathx')
       mudd_scamudd = HoldingLocation.find_by(code: 'mudd$scamudd')
 
-      expect(Library.count).to eq 13
-      expect(HoldingLocation.count).to eq 136
+      expect(Library.count).to eq 18
+      expect(HoldingLocation.count).to eq 309
```
christinach, line 45: The fixture file includes many holding_locations but not all.

sandbergja (approved):
Thanks @christinach !

PR #1987, opened 2022-09-08
Add Committed and Uncommittable records to dump

maxkadel:
Closes pulibrary/orangelight#3172

### DIFF
--- a/app/models/scsb_full_records_request.rb
+++ b/app/models/scsb_full_records_request.rb
@@ -21,10 +21,11 @@ class ScsbFullRecordsRequest
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/models/scsb_full_records_request.rb
```diff
@@ -21,10 +21,11 @@ def scsb_conn
     end
   end
 
+  # See https://scsb.recaplib.org:9093/swagger-ui.html#/data-dump-rest-controller/exportDataDump for documentation
```
sandbergja, line 24: Thanks for adding the docs URL!

sandbergja (approved):

PR #1992, opened 2022-09-13
Add fields for uniform title-only display

### LINKED ISSUE #1990: Index uniform title without author to uniform_title_1display
Uniform titles for works with authors are currently being indexed to `name_uniform_title_1display`. We should follow the same format for works without authors (in the 130 field, rather than the 240 field), for display in Orangelight.

maxkadel:
- Existing fields only include author-title display

Closes #1990

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -1292,10 +1292,30 @@ to_field 'uniform_240_vern' do |record, accumulator|
   end
 end
 
+to_field 'uniform_130' do |record, accumulator|
+  MarcExtractor.cached('130apldfhkmnorst', alternate_script: false).collect_matching_lines(record) do |field, spec, _extractor|
+    field.subfields.each do |s_field|
+      next if (!spec.subfields.nil? && !spec.subfields.include?(s_field.code))
+      accumulator << s_field.value
+    end
+    break
+  end
+end
+
+to_field 'uniform_130_vern' do |record, accumulator|
+  MarcExtractor.cached('130apldfhkmnorst', alternate_script: :only).collect_matching_lines(record) do |field, spec, _extractor|
+    field.subfields.each do |s_field|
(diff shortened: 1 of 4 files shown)


maxkadel:
Records to check:

- 9941520283506421
- 9924703803506421
- 9911499533506421
- 997697413506421
- 991415603506421
- 9947310053506421
- 9946864703506421
- 9946333583506421
- 9912557153506421
- 9946093213506421

maxkadel:
Disregard any location issues - it's an unrelated issue with experimenting with rare book locations in the Alma sandbox.

christinach:
> Disregard any location issues - it's an unrelated issue with experimenting with rare book locations in the Alma sandbox.

To avoid confusion, I think @maxkadel is referring to the traject log errors referencing rare locations that are not in bibdata because they are in an experimenting stage in the alma sandbox. The records that have holdings in these locations will not get indexed. I'm assuming some of the mmsids that @maxkadel is referencing for testing are in these locations so don't expect to see them getting indexed in catalog-staging.

maxkadel:
Some of the locations are also displaying strangely, which I believe is because of the locations experimentation as well, rather than this code.

marc_to_solr/lib/traject_config.rb
```diff
@@ -1352,11 +1375,27 @@
...
+  doc = context.output_hash
+  uniform_t = []
+  search_field = []
+
+  uniform_t << doc['uniform_130'] if doc['uniform_130']
+  uniform_t << doc['uniform_130_vern'] if doc['uniform_130_vern']
+
+  context.output_hash['uniform_title_1display'] = [uniform_t.to_json.to_s] unless uniform_t.empty?
+
+  # these fields are no longer necessary
```
christinach, line 1396: do you need to remove this comment?

maxkadel: I didn't think so - it's explaining why the fields are being removed (same as 1375)

kevinreiss (approved):

PR #2012, opened 2022-10-20
Cleanup location_data_service_spec and delete locations from spec holding locations file that are not used.

christinach:


sandbergja (dismissed):
Thanks, @christinach !

sandbergja (approved):
Thanks for reducing the size of this fixture file, @christinach !

PR #2024, opened 2022-11-10
Dont print private 583 notes

### LINKED ISSUE #2023: Action Note (583) with first idicator of 0 should not be displayed
### Expected behavior
When 583 fields have a first indicator of 0, that indicates they are private and should not be displayed.

### Actual behavior
The fields are being displayed

### Steps to replicate
See https://catalog.princeton.edu/catalog/99106353023506421

### Impact of this bug
Information that is private is currently visible to the public

### Re-index required?

- [ ] Full re-index
- [x] Partial re-index
- [ ] No re-indexing required

christinach:
closes #2023


sandbergja (approved):
Thank you, @christinach !

PR #2037, opened 2022-12-01
Adds monthly scsb:import cron job

### LINKED ISSUE #2017: Add a cron job to import the partner records once a month.
When we need to run a full reindex we usually run an ad hoc job to import the full partners records from SCSB. The job takes around 26-27 hours to complete.

When this cron job is running, bibdata should not be deployed or restarted. (This is why it's better to schedule this job on a  Friday night or Saturday early am hours.)

The rake task is: `bundle exec rake scsb:import:full`

maxkadel:
- crontab does not have a concept of "Every 4th Saturday" or "Monthly on Saturday", so we give a one-week range, and then use logic in the rake task to either run or not run the job.

Closes #2017

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -54,3 +54,8 @@ end
 every 2.weeks, at: '11:00pm', roles: [:cron_production] do
   rake "scsb:request_records[production,mk8066@princeton.edu,HL]", output: "/tmp/cron_log.log"
 end
+
+# Runs on the second Saturday of the month at 9:00 am
+every '0 9 8-14 * *', roles: [:cron_production] do
+  rake "scsb:import:full_saturdays_only", output: "/tmp/cron_log.log"
+end
--- a/lib/tasks/scsb.rake
+++ b/lib/tasks/scsb.rake
@@ -74,5 +74,11 @@ namespace :scsb do
     task full: :environment do
       ScsbImportFullJob.perform_later
     end
+
+    desc "Used for monthly cron job - downloads files for full partner record set only on Saturdays"
+    task full_saturdays_only: :environment do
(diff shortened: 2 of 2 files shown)


kevinreiss (approved):

config/schedule.rb
```diff
@@ -54,3 +54,8 @@
 every 2.weeks, at: '11:00pm', roles: [:cron_production] do
   rake "scsb:request_records[production,mk8066@princeton.edu,HL]", output: "/tmp/cron_log.log"
 end
+
+# Runs on the second Saturday of the month at 9:00 am
+every '0 9 8-14 * *', roles: [:cron_production] do
```
christinach, line 59: I think the second asterisk should be 6 to be on a Saturday. Right now it will run it every day.

christinach (changes requested):

lib/tasks/scsb.rake
```diff
@@ -74,5 +74,11 @@ namespace :scsb do
     task full: :environment do
       ScsbImportFullJob.perform_later
     end
+
+    desc "Used for monthly cron job - downloads files for full partner record set only on Saturdays"
+    task full_saturdays_only: :environment do
+      abort "This task will only run on Saturdays, to facilitate monthly cron job. If you want to run this job manually on another day, use scsb:import:full" unless Date.today.saturday?
```
christinach, line 80: My understanding is by having 'unless Date.today.saturday' cron will try to run the job and then abort it. Why to do that since we can set 6 (for Saturday) in the cron job. Between 8-14 there is always a Saturday and cron will run once without a need to abort daily until it's a Saturday.

maxkadel: Crontab is weird, and it uses an "OR" for correct day OR correct day of the week, so that would make it run every day in that week, and every Saturday. see https://stackoverflow.com/questions/34357126/why-crontab-uses-or-when-both-day-of-month-and-day-of-week-specified

PR #2057, opened 2023-01-04
Allow linked fields to index when id includes 99

### LINKED ISSUE #2055: MMS IDs containing 99 not linking correctly
Best explained by examples:
9924830023506421
Two AVA, including one with host record number 99124984**99**3506421 from 773 $$w.  It does not appear in the display.

9913692683506421
Three AVA.  The one with 9912548**99**25606421 does not display.

99125489931206421
Constituent (774) $$w 997**994903506421** (title: Béhémoth) displays as The Westminster gazette [microform] which has MMS ID **994903506421**.

It appears that indexing for host/constituent linking is skipping to the second "99" and either failing because there's no corresponding record with the partial MMS ID or returning a false match.  Users can't request or can't properly identify contents.

Record relations are OK in Alma.

christinach:
closes #2055


maxkadel (approved):
Thanks @christinach!

Issue #2069, opened 2023-01-24
Indexing updates for ISO 693-3 three-letter codes


escowles:
CaMS is now including ISO 693-3 three-letter language codes where appropriate (sample record: https://catalog.princeton.edu/catalog/9930372423506421/). DSSG discussed and approved two changes to how these are indexed:

1. We should display the more specific language names from ISO 639-3, e.g., the record above should display "Language: Wu Chinese".
2. We should index both the more specific name and the more general/family name in the Language facet, so both "Chinese" and "Wu Chinese".

mzelesky:
See https://catalog.princeton.edu/catalog/9930372423506421/staff_view

The 041s that have the ISO 639-3 have a $2 of `iso639-3`

So traject can handle the 041s with $2 iso639-3 separately from 041s without that $2.

christinach:
Add the specific language name in the 'language_facet' .

sandbergja:
We aren't always guaranteed to be able to find the more general language name (the "macrolanguage" using ISO 639-3 terminology) given a certain ISO 639-3 code.

The provided example, `wuu`, is included in the [ISO 639-3 macrolanguage table](https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3-macrolanguages.tab) with a macrolanguage of `zho` (Chinese).

However, another example: `nuz` (Tlamacazapa Nahuatl) isn't mapped to a macrolanguage of Nahuatl, since no such macrolanguage exists in ISO-639-3 (it does in ISO-639-2 and ISO-639-1).  It seems that we would want both "Tlamacazapa Nahuatl" and "Nahuatl" in the facets -- will CaMS catalogers always put in both the ISO-639-3 and MARC code in these records?

sandbergja:
A possible gem: https://github.com/bbenno/languages -- it does not currently include macrolanguages.

sandbergja:
Also, for the facet display, do we want them to be presented as a list, or as a "pivot facet" like the call number facet?

maxkadel:
Trying to understand this issue better - these descriptions of the relationship between the iso-639-2/iso-639-5 and iso-639-3 standards were helpful to me - https://www.loc.gov/standards/iso639-2/iso639jac.html and https://iso639-3.sil.org/about/relationships 

It looks like Nahuatl is a "collective", rather than a "macrolanguage" (see https://iso639-3.sil.org/code/nah vs. https://iso639-3.sil.org/code/zho). 

[...]

PR #2092, opened 2023-03-10
i2069: Add a new field for specific languages on the Show page

sandbergja:
This adds a new solr field, `language_name_display`, which we could then add to the Show page.

Part of #2069

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -13,6 +13,7 @@ require_relative './solr_deleter'
 require_relative './access_facet_builder'
 require_relative './augment_the_subject'
 require_relative './language_service'
+require_relative './language_extractor'
 require_relative './pul_solr_json_writer'
 require 'stringex'
 require 'library_stdnums'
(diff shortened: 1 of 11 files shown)


marc_to_solr/lib/traject_config.rb
```diff
@@ -727,6 +728,10 @@
 #    546 XX 3a
 to_field 'language_display', extract_marc('5463a')
 
+to_field 'language_name_display' do |record, accumulator|
```
christinach, line 731: @sandbergja can you add a comment that this field will be used to show specific languages in the record page?

spec/marc_to_solr/lib/language_extractor.rb
```diff
@@ -0,0 +1,66 @@
...
+
+RSpec.describe LanguageExtractor do
+  let(:fields) do
+    [
+      { '008' => '120627s2012    ncuabg  ob    001 0 gre d' }
+    ]
+  end
+  let(:extractor) { described_class.new(LanguageService.new, MARC::Record.new_from_hash('fields' => fields)) }
+  it 'finds a language from the 008' do
+    expect(extractor.specific_names).to eq(['Modern Greek (1453-)'])
```
christinach, line 12: 🇬🇷

christinach (approved):
Thank you @sandbergja 🍄 . Just one comment in traject_config.

As part of a new ticket can we add a document in the `docs/` file describing the need and the logic of the language display field ? I'm thinking more of the meeting we had with the catalogers and the info they provided.

marc_to_solr/lib/traject_config.rb
christinach: I created one. https://github.com/pulibrary/bibdata/issues/2100

christinach (approved):
Thank you @sandbergja 🍄 🍄

PR #2099, opened 2023-03-13
Add retention reasongs IPLCBrill and ReCAPSACAP #2094

### LINKED ISSUE #2094: Two new Retention Reasons to be added to the Committed CGD array
There are 2 new retention reasons that should make ReCAP items 'Committed' for the CGD.

1. IPLCBrill
2. ReCAPSACAP

The ReCAPSACAP items are already at PUL, but the IPLCBrill items will likely not arrive before April.

christinach:
closes #2094

### DIFF
--- a/spec/adapters/alma_adapter/alma_item_spec.rb
+++ b/spec/adapters/alma_adapter/alma_item_spec.rb
@@ -21,7 +21,7 @@ RSpec.describe AlmaAdapter::AlmaItem do
     )
   end
 
-  let(:item_committed_to_retain) do
+  def build_item_committed_to_retain(retention_reason:)
     Alma::BibItem.new(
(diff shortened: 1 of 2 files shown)


spec/adapters/alma_adapter/alma_item_spec.rb
```diff
@@ -21,7 +21,7 @@
     )
   end
 
-  let(:item_committed_to_retain) do
+  def build_item_committed_to_retain(retention_reason:)
```
sandbergja, line 24: Nice!  I like having this as a helper method.

sandbergja (approved):
🐬  Thanks, @christinach !

sandbergja (approved):

PR #2110, opened 2023-03-27
Update indexing documentation

### LINKED ISSUE #2107: Update indexing documentation
Add as a first step to turn off sneakers in catalog-qa
The reason of doing that is that we usually hook the catalog-production-rebuild in catalog-qa and we dont want coins from figgy-staging to get indexed during the full-reindex.

christinach:
closes #2107

### DIFF
--- a/docs/indexing.md
+++ b/docs/indexing.md
@@ -10,13 +10,13 @@ Alma MMS ids start with 99 and end with 3506421
 Frequency of updates: 4x per day: 4am, 10am, 4pm, 10pm see: https://github.com/pulibrary/bibdata/blob/main/docs/alma_publishing_jobs_schedule.md
 
 ## Source: SCSB
-Around 10 million items shared by our ReCAP Partners, Columbia, NYPL, and Harvard, pulled through HTC’s shared collection software https://github.com/ResearchCollectionsAndPreservation/scsb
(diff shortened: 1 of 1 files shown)


docs/indexing.md
```diff
@@ -191,17 +192,43 @@ $ SET_URL=http://lib-solr8-prod.princeton.edu:8983/solr/catalog-alma-production-
...
+Turn off sneakers in catalog-qa:
+- cd in your local princeton_ansible directory
+```
+$ pipenv shell   
+$ ansible orangelight_qa -u pulsys -m shell -a "sudo service orangelight-sneakers stop"
+```
+
+`$ ssh deploy@catalog-qa1`
+
+Use your preffered editor -nano or vi- and manually edit `app_configs/orangelight`
```
maxkadel, line 207: If you do this manually, isn't there a change that running the Orangelight playbook against QA would set it to the wrong value? Would it be better to change it in Princeton Ansible and run the Orangelight playbook against QA with the `site_config` tag?

christinach: Using catalog-qa to review the index is not something critical. It's a suggestion so that we can do a QA and review the results at the and of the indexing process. You can use ansible to change it if you want but it should not be a permanent change; By default catalog-qa should be pointing to solr-staging and catalog-qa index.

PR #2115, opened 2023-03-30
Add aria-labels to tables in documentation, since github removes <caption> tags

sandbergja:
#2113 added some tables to the documentation, and used `<caption>` tags to provide accessible names.  However, github removes these tags, so this PR uses an `aria-label` to provide the accessible names instead.

Without `aria-label`, the Voiceover table rotor lists:
- 4 columns, 5 rows
- 3 columns, 3 rows
- 3 columns, 5 rows

With the `aria-label`, Voiceover can let the user know what each table is:

- Codes and other data included in language standards 4 columns, 5 rows
- Language codes included in MARC records 3 columns, 3 rows
- Language-related Solr fields 3 columns, 5 rows


christinach (approved):
Thanks @sandbergja

PR #2124, opened 2023-04-04
Add ISO639-3 languages to facets

### LINKED ISSUE #2069: Indexing updates for ISO 693-3 three-letter codes

sandbergja:
closes #2069


christinach (approved):
Thanks @sandbergja !

PR #2131, opened 2023-04-12
Return "Unavailable" not "Not Available"

### LINKED ISSUE #2006: Bibdata Availability Response API should return only the "Unavailable" status
The bibdata availability API sometimes returns "unavailable" as the status of an item, in other cases it returns "not available" we should update it to be consistent and only return "unavailable".

Reference points: 
https://github.com/pulibrary/bibdata/blob/bfb518b999f072b7d690577bad708b3a76ce91ca/app/adapters/alma_adapter/status.rb#L19

https://github.com/pulibrary/bibdata/blob/e97b8911e50c10a2b6d2279cbddc2317d9f67b91/app/adapters/alma_adapter/availability_status.rb#L42

Related issue: https://github.com/pulibrary/orangelight/issues/3234

maxkadel:
Closes #2006


christinach (approved):
Thanks @maxkadel 🦄

Issue #2167, opened 2023-05-25
Add indexing changes since Sept 2022 


kevinreiss:
Check the following file change histories to make sure our docs are up to date.

This is the documentation that needs to be updated: https://lib-confluence.princeton.edu/display/ALMA/MARC+fields+indexed+in+Blacklight

- [x] Orangelight [catalog_controller.rb](https://github.com/pulibrary/orangelight/blob/main/app/controllers/catalog_controller.rb)
- [x] Pul_solr [solr_config.xml](https://github.com/pulibrary/pul_solr/blob/main/solr_configs/catalog-production-v2/conf/solrconfig.xml) [schema.xml](https://github.com/pulibrary/pul_solr/blob/main/solr_configs/catalog-production-v2/conf/schema.xml)
- [x] Bibdata traject configuration [traject_config.rb](https://github.com/pulibrary/bibdata/blob/main/marc_to_solr/lib/traject_config.rb), [princeton_marc.rb](https://github.com/pulibrary/bibdata/blob/main/marc_to_solr/lib/princeton_marc.rb),

sandbergja:
- [x] language indexing changes
- [ ] access note changes

sandbergja:
@maxkadel and I will pair to look into the action note questions.

sandbergja:
Finished with updates to confluence!

PR #2172, opened 2023-06-20
Add action note documentation

sandbergja:
helps with #2167

### DIFF
--- a/docs/action_notes.md
+++ b/docs/action_notes.md
@@ -5,7 +5,7 @@ Action notes are stored in Solr as a JSON field called
 583 field.  It displays as a link in the catalog if
 a 583$u is present, otherwise it displays as plain text.
 
-Action notes are only indexed if the second indicator is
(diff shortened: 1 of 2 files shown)


docs/action_notes.md
```diff
@@ -0,0 +1,13 @@
+# Action notes
+
+Action notes are stored in Solr as a JSON field called
+`action_notes_1display`.  Its data come from the MARC
+583 field.  It displays as a link in the catalog if
+a 583$u is present, otherwise it displays as plain text.
+
+Action notes are only indexed if the second indicator is
```
mzelesky, line 8: It's the first indicator, not the second indicator.

mzelesky (changes requested):

mzelesky (approved):
It looks good. I'm assuming the languages markdown has nothing to do with the 583, per se, that it's just correcting another issue.

PR #2191, opened 2023-07-19
Fix circleci failure

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #2225, opened 2023-07-29
maint: removed deprecated endpoints

### LINKED ISSUE #2192: Remove deprecated endpoints
We have several deprecated endpoints, which we should now remove:

https://github.com/pulibrary/bibdata/blob/3e22170f0c46869f80e8d340a9317d2b68f3fa18/config/routes.rb#L33-L45

- [ ] Remove these from config/routes.rb
- [ ] Remove the update and item_discharge methods from the Bibliographic controller

haiderashfaq:
resolves #2192


sandbergja (approved):
Thanks, @haiderashfaq !

sandbergja (changes requested):
@haiderashfaq -- now that the tests have run, it looks like there are several failing tests.  Could you please address those as part of your branch?  I believe most of them can just be deleted.

haiderashfaq:
> @haiderashfaq -- now that the tests have run, it looks like there are several failing tests. Could you please address those as part of your branch? I believe most of them can just be deleted.

all of the failed cases were for removed deprecated routes and actions.  So, removed test cases for them as well.

spec/routing/bibliographic_routing_spec.rb
```diff
@@ -2,8 +2,5 @@
 
 RSpec.describe BibliographicController, type: :routing do
   describe "routing" do
-    it "routes to #item_discharge" do
```
sandbergja, line 5: Since this file no longer has any tests, could you please delete it outright?

sandbergja (changes requested):
Thanks, @haiderashfaq , it's looking good! Just two small requests.

PR #2241, opened 2023-09-07
Migrate to zeitwerk

sandbergja:
Helps with #2053 

Deletes some classes that are not used and do not conform to zeitwerk autoloading.


christinach:
@sandbergja merging https://github.com/pulibrary/bibdata/pulls and rebasing on main would fix the ci failure.

christinach (approved):

Issue #2074, opened 2023-02-09
Add axe accessibility test to bibdata


christinach:
[Orangelight accessibility spec](https://github.com/pulibrary/orangelight/blob/main/spec/features/accessibility_spec.rb)
Link to check when you start the server locally:

- [ ] http://localhost:3000/ 

Same link of bibdata in production:
- [ ] https://bibdata.princeton.edu/

christinach:
closed by https://github.com/pulibrary/bibdata/pull/2258

PR #2258, opened 2023-09-30
Include axe accessibility tests to bibdata

lizgarciao:
toward #2074

Accessibility issues found:
- [ ] Nav bar links have insufficient contrast.
- [ ] Home page use of ids is not unique.
- [ ] Libraries and holding locations pages: Certain ARIA roles must contain particular children (critical)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en-US">
 <head>
   <title>Bibdata</title>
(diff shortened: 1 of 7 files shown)


app/views/layouts/application.html.erb
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en-US">
```
christinach, line 2: Can we change this to  lang="en" ?

christinach: we can do it later. :)

christinach (approved):
@lizgarciao thank you for the improvements!

PR #2272, opened 2023-11-26
Include 10 events in seeds of database

### LINKED ISSUE #2268: Database seeds should include Events
### User story

As a developer making a change to the events page, I would like to have some Events already existing in the database, so that the events page isn't totally empty.

### Success criteria

- [x] When I run `rake servers:start`, it adds 10 events to the database if they don't already exist
- [x] I can confirm that the 10 events display on the [events page](http://localhost:3000/events)

### Implementation notes

You can add a single event if it doesn't yet exist with something like this:

```
Event.find_or_create_by(id: 12, start: '2020-10-10', finish: '2020-10-11', success: true)
```

lizgarciao:
closes #2268


db/seeds.rb
```diff
@@ -1,3 +1,7 @@
 seeder = DataSeeder.new
 seeder.generate_dump_types
 seeder.generate_dump_file_types
+
+10.times do |i|
+  Event.find_or_create_by(id: i+1, start: "200#{i}-10-10", finish: "200#{i}-10-11", success: true)
```
sandbergja, line 6: I like the year interpolation!

sandbergja (approved):
Thanks, @lizgarciao !

db/seeds.rb
sandbergja: It looks like there are some failing tests.  I think because you are specifying ids here, the database doesn't think to increment the id sequence, so there are failures when it attempts to create a second row with the same ID (e.g. 1).

If you don't specify an id in this line, and just let the database keep track of assigning IDs, it should fix it.

lizgarciao: Thanks for the feedback!

christinach:
There are a couple of things that we need to do for this ticket:
- When we create an event we should also create a dump with a file and mark the event success: true. I'll review the relations in the code and update the logic that needs to be followed for the creation of these records in the seeds file.
- In the failing spec: 1. we can delete all the events (which will also delete the dumps etc that are connected to the specific event.) in a `before do`. However this would mean that someone would have to run the seeds file again in env: dev to see the events in the events page. 2. Or we could run this part of the seeds file only in env: dev so that we don't mess up with the test env. 3. Or we can review the specs and see how we can improve them.

db/seeds.rb
```diff
@@ -1,3 +1,9 @@
 seeder = DataSeeder.new
 seeder.generate_dump_types
 seeder.generate_dump_file_types
+
+if Event.count < 10 && Rails.env.development?
```
christinach, line 5: 👍🏻

spec/controllers/events_controller_spec.rb
```diff
@@ -5,12 +5,25 @@
...
+        index = Event.count # index of first test to be created
         event2 = FactoryBot.create(:event, start: Time.now - 3.months, finish: Time.now - 3.months + 180, error: nil, success: true)
         event3 = FactoryBot.create(:event, start: Time.now - 2.months, finish: Time.now - 2.months + 180, error: nil, success: true)
         event1 = FactoryBot.create(:event)
         event4 = FactoryBot.create(:event, start: Time.now - 1.month, finish: Time.now - 1.month + 180, error: nil, success: true)
+
+        # controller action that assigns events order: asc by start
         get :index
-        expect(assigns(:events)).to eq([event2, event3, event4, event1])
+
```
christinach, line 16: do you still need these changes with your recent update in the seeds file?

christinach (approved):
Thanks @lizgarciao 😃

PR #2284, opened 2023-12-21
Upgrade old gem dependencies

sandbergja:
Removes the rerun gem, since it is no longer used.  Also, remove some leftover files from the spring gem.


christinach (approved):
Thanks @sandbergja !

PR #2293, opened 2024-01-10
add new jammy vms back

kayiwa:


christinach (commented):
@kayiwa can you remove the log/.keep file

christinach (approved):
Thanks @kayiwa !

PR #2300, opened 2024-01-12
Cleanup factories/specs that are not used

### LINKED ISSUE #2299: Cleanup factories in event factory
Event factory includes factories that are not used anymore.
Update specs with correct factories.
Remove specs and fixtures that are not used.

christinach:
closes #2299

Cleanup factories that are not used
Update specs with correct factories
Remove specs and fixtures that are not used because code was removed


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2303, opened 2024-01-29
Add firestone$tech

### LINKED ISSUE #2276: New Alma location firestone$tech
See https://github.com/PrincetonUniversityLibrary/alma-config/issues/780

The location should never have any items sent to bibdata. In this case, does the location need to be added to bibdata?

christinach:
closes #2276


mzelesky (approved):
This location behaves like the other locations that currently have tech equipment, so this looks good.

PR #2327, opened 2024-04-01
Remove outdated content from the home page

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #2344, opened 2024-04-10
Add fixture for record with fast headings

Beck-Davis:


christinach (approved):
Thanks @Beck-Davis and @sandbergja

PR #2351, opened 2024-04-18
Remove ISSUE_TEMPLATE directory so that we use org-wide templates

maxkadel:
Connected to pulibrary/dacs_handbook#170

The hope is that once we remove this directory in main, this repository will use the issue templates from [the pulibrary organization .github repo](https://github.com/pulibrary/.github)


christinach (approved):
Thanks @maxkadel !

Issue #2348, opened 2024-04-17
Refactor `DumpType` to use configuration, not database


maxkadel:
### What maintenance needs to be done?
Refactor `DumpType` to use only configuration, not database tables.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
- This class is mostly used in categorizing dumps, and retrieving them. The `DumpType` table is only 7 lines long and never changes, which means it behaves more like configuration, 
- We are seeing errors in using the database table, including in the SQS Poller (see [Princeton Ansible PR for changes to Postgres config to try to fix this](https://github.com/pulibrary/princeton_ansible/pull/3333)), and with 15 minute database queries against this table (See [Bibdata issue](https://github.com/pulibrary/bibdata/issues/2312))

### Acceptance criteria
 - [ ] The code does not try to retrieve data from the DumpType table in Postgres
 - [ ] New and old events still have the correct "Dump Type" in the UI
 - [ ] Tests continue to pass
 - [ ] Make rollback as easy as possible (aka, don't immediately drop the table on first deploy without a plan for if something goes wrong)

maxkadel:
Investígate changing DumpType to [enum](https://www.honeybadger.io/blog/how-to-use-enum-attributes-in-ruby-on-rails/#:~:text=To%20add%20an%20enum%20attribute%20to%20an%20existing%20table%20in,by%20creating%20a%20new%20migration.&text=Next%2C%20edit%20your%20model%20to,mappings%20of%20strings%20to%20integers.&text=That's%20it!,to%20actually%20use%20the%20enum.)

PR #2352, opened 2024-04-21
Move towards using config for DumpType

maxkadel:
- Replace instances of DumpType with dump_type_id - these ids are consistent across environments.

Connected to #2348


christinach (commented):
Thanks @maxkadel

PR #2357, opened 2024-04-25
Use enum for DumpType

### LINKED ISSUE #2348: Refactor `DumpType` to use configuration, not database

maxkadel:
We have seen many issues with the postgres database queries around DumpTypes, so trying putting them into an enum. The number in the database table should be the same whether it's a `dump_type_id` or the enum. 

Enums come with some nice stuff for free, like scopes and booleans (e.g. `dump.changed_records?`). 

Closes #2348

### DIFF
--- a/app/jobs/delete_events_job.rb
+++ b/app/jobs/delete_events_job.rb
@@ -1,18 +1,11 @@
 class DeleteEventsJob < ApplicationJob
-  # @param dump_type [String] the dump type of the dumps you want to clean up
-  # @param older_than [int] the Time before which dumps should be deleted, serialized as an integer
+  # @param dump_type [Symbol] the dump type of the dumps you want to clean up
+  # @param older_than [int] the Time before which dumps should be deleted
   def perform(dump_type:, older_than:)
-    start = Event.arel_table[:start]
-    ids = Event
-          .joins(dump: :dump_type)
-          .where('dump_types.constant': dump_type)
-          .where(start.lt(deserialize_time(older_than)))
-          .map(&:id)
-
-    Event.destroy(ids)
-  end
-
-  def deserialize_time(time)
-    Time.zone.at(time)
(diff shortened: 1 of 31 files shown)


app/jobs/delete_events_job.rb
```diff
@@ -1,18 +1,11 @@
...
-    Event.destroy(ids)
-  end
-
-  def deserialize_time(time)
-    Time.zone.at(time)
+    event_ids = Event.joins(:dump)
+                     .where('dump.dump_type': dump_type)
+                     .where(start: ..older_than)
+                     .map(&:id)
+    Event.destroy(event_ids)
```
maxkadel, line 9: Do we actually get anything by mapping to ids and then deleting the events?

sandbergja (dismissed):
Thanks, @maxkadel 🦄

sandbergja (approved):
🦄

PR #2358, opened 2024-04-25
SQS Poller - Create dump in one transaction

maxkadel:
We have seen long-running postgres queries in this area of the code, trying making the Dump object in a single transaction, to see if that helps. Add a test for an untested class.


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #2362, opened 2024-04-29
Ensure poller is restarted on poller box

maxkadel:
- We should only run the poller on one server per environment
- Princeton Ansible installs the daemon on the first worker box (see https://github.com/pulibrary/princeton_ansible/pull/4875)


christinach (approved):
thanks @maxkadel

Issue #2363, opened 2024-04-30
Make DNS names of servers consistent


maxkadel:
### What maintenance needs to be done?
We need to re-build the Bibdata VMs to ensure the SQS Poller is only installed on one VM per environment. 

Right now, the VMs have inconsistent DNS names, sometimes including "alma", sometimes not, sometimes including the environment name, sometimes not. While we are rebuilding the VMs anyway, we should give them consistent names. 

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
Consistency will allow us to work more quickly, efficiently, and confidently, and reduce mistakes.

### Acceptance criteria
 - [x] DNS names in Documentation match those in the [Prancible New VM issue](https://github.com/pulibrary/princeton_ansible/issues/4877)
 - [x] DNS names for deployment match those in the [Prancible New VM issue](https://github.com/pulibrary/princeton_ansible/issues/4877)
 - [x] You have checked for and updated any other instances of DNS names and updated them appropriately
 - [x] Do it in two separate PRs, so we can start with the Staging rename without impacting the Production service.

PR #2365, opened 2024-04-30
Update production DNS names

maxkadel:
Connected to #2363

Drafted until the new VMs are built.


maxkadel:
Going to add all boxes, then remove old boxes once we've switched to the new ones.

PR #2366, opened 2024-05-01
Add new boxes for deploy

maxkadel:
- My understanding is that only one box should do database actions per environment
- There are no longer any hr_cron jobs


sandbergja (approved):
Thanks, @maxkadel 🦄

Issue #2383, opened 2024-05-23
Don't index Private SCSB records


christinach:
### What maintenance needs to be done?
The following are the CSV files that are being generated with each full partner dump:
[ExportDataDump_Full_HL_20240515_230000.csv](https://github.com/pulibrary/bibdata/files/15421991/ExportDataDump_Full_HL_20240515_230000.csv)
[ExportDataDump_Full_NYPL_20240515_150000.csv](https://github.com/pulibrary/bibdata/files/15421994/ExportDataDump_Full_NYPL_20240515_150000.csv)
[ExportDataDump_Full_CUL_20240515_070000.csv](https://github.com/pulibrary/bibdata/files/15421995/ExportDataDump_Full_CUL_20240515_070000.csv)

We want to make sure that we download full dumps from partners with the attached CSV to have the following codes `collection group Id(s): 1*2*5*6`

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
- Full dumps from partners in S3 may include private records (collection group Id: 3). We don't want to index them.

### Acceptance criteria
 - [ ] Don't import a full dump that the attached CSV includes '3' in the collection group Id(s) column.

maxkadel:
Requested private CUL records to be sent from / to UAT 5/29/2024, waiting for them to be marked delivered.

sandbergja:
This has been tested in staging, and it works great!  Closing.

PR #2388, opened 2024-05-29
Do not process full dumps with private records

maxkadel:
- Add another dump file type for metadata CSVs
- Naming format is different for CSVs, so need to get date out different, factor out that method
- Create method to validate the metadata CSV, aka, make sure that it is present and does not indicate that the associated file includes private records
  - Adds error messages if the file is missing or indicates that there are private records
- If the metadata CSV does not validate, do not download and process the associated full dumps for that institution
- Still download and process full dumps for other institutions, if their metadata CSVs validate

Connected to #2383


kevinreiss (approved):
this approach looks good to me @maxkadel. If at some point we actually encounter a SCSB file where a private item shows up even though it wasn't requested, we can deal with that. We've never observed that before so relaying on the CSV metadata file seems more than adequate to guard against processing a dump file that might have been requested to include partner private records.

spec/models/scsb/partner_updates/full_spec.rb
```diff
@@ -0,0 +1,99 @@
+require 'rails_helper'
+
+RSpec.describe Scsb::PartnerUpdates::Full, type: :model do
+  include_context 'scsb_partner_updates_full'
```
sandbergja, line 4: Cool!

sandbergja (approved):
Thanks, @maxkadel 🦄

PR #2391, opened 2024-05-31
Fix event seeds, address deprecation warning

maxkadel:
- The seeds were creating events without dates, since they were trying to create the year "20010", which the DB apparently doesn't believe in. This made it so that the events could not be viewed in the UI.
- Address deprecation warning that was distracting from the real issue


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #2392, opened 2024-05-31
Fix json display of dumps

maxkadel:
- The keys for the "files" were displaying as `["bib_records", 1]` which was distracting and difficult to read


sandbergja (approved):
I deployed to staging, then opened the json from a link in the Dump Type column.  That looks much nicer!  Thanks, @maxkadel 🦄

PR #2393, opened 2024-06-04
Do not index SCSB items if they are private

maxkadel:
- If there are only private items, skip indexing the whole record
- If there is a mix of private and non-private items, only index the private items

Connected to #2383

Tested successfully on staging using a file from SCSB that included private items.


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #2395, opened 2024-06-05
Update staging solr url to lib-solr8d-staging

maxkadel:
- We now have separate staging and production environments


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #2396, opened 2024-06-06
Upgrade omniauth, omniauth-cas, and omniauth-rails_csrf_protection gems

sandbergja:
Currently deployed to staging


maxkadel (approved):
Thanks @sandbergja! 🍄

PR #2406, opened 2024-06-20
Remove mailers

### LINKED ISSUE #2404: Remove mailer configuration ~Move to new Mail Relay Server~
Bibdata does not send mail, so mailers and mail configuration should be removed.

~Review the following PRs for each of our application that utilizes the mail relay.  Confirm that the following is true after deploying to prod:~

- ~I can still send and receive emails~
- ~/var/log/mail on lib-ponyexpr-prod shows that the mails are sent successfully~
- ~For a rails app, make sure that `enable_starttls` is false.  See https://github.com/pulibrary/orangelight/pull/4026 and https://github.com/pulibrary/pdc_describe/pull/1843 for examples.~

maxkadel:
Bibdata does not send email. Remove defunct code that sent mail and commented out configuration that is not currently needed. 

Leaving the application mailer class and the commented out config to make it easier to reintroduce email in the future.

Closes #2404


kevinreiss (changes requested):
Do we want to remove this gem reference in this commit? 

https://github.com/pulibrary/bibdata/blob/main/Gemfile#L36

Everything you've removed here is related to early experiment when SCSB came up to utilize the SCSB active MQ message architecture to send notifications about events. We never did use it in production and it's been here all these years.

kevinreiss (approved):

PR #2421, opened 2024-07-24
Add solr, DB, and cache health checks

### LINKED ISSUE #2409: Add a health.json endpoint that determines this application is ready to be live to users

We want the load balancer to never serve sites that are broken. That can happen automatically if every application has a health.json endpoint that the load balancer can check.

### Level of urgency

- [ ] High
- [x] Moderate
- [ ] Low

Right now it's easy to deploy to one of x servers, have that deploy fail, and thus have the site go down, unless we do a laborious load balancer dance.

### Acceptance criteria

- [ ] There's a `/health.json` endpoint. 
- [ ] It checks at least the following services (which will be checked by nginx)
  - [ ] Database
  - [ ] Alma API

An example implementation: https://github.com/pulibrary/lae-blacklight/pull/380/files

### First Step

Install the health-monitor-rails gem to the gem file, as per the above PR.

Health monitor rails gem: https://github.com/lbeder/health-monitor-rails

maxkadel:
Closes #2409


christinach (approved):
Deployed on bibdata-staging1. Changed the solr env 
curl localhost/health.json
The solr health check return 502 bad gateway as expected. 

Thanks @maxkadel

Issue #2413, opened 2024-07-12
Allow 773/774 fields to have institution identifier prefix before MMS ID


mzelesky:
### Current indexing
https://github.com/pulibrary/bibdata/blob/main/marc_to_solr/lib/princeton_marc.rb#L820 only accepts 773 and 774 fields that have $w that starts with `99` and ends with `6421`. 

### Desired Indexing
To allow our records to follow the MARC format (see https://www.loc.gov/marc/bibliographic/bd760787.html for $w), we need to add in Alma the prefix `(NjP)` to all 773 and 774 $w. 

(important:)This will break the current indexing routine unless we support
`valid_linked_fields` to accept a $w that may or may not start with `(NjP)` as long as there is a valid MMS ID that ends the subfield.

Examples:
773  __ $w (NjP)9923506421
773 __ $w 9923506421

This will allow both ways to work while we remediate the bib records.
### Acceptance criteria
- [ ] 773 and 774 linkages work (important:) with or without the presence of the `(NjP)` prefix - 

### Implementation Notes
From @christinach on slack:
```The index will accept 773 and 774 with a $w that may or may not start with (NjP) as long as there is a valid MMS ID that ends the subfield. When NJP is there and the mms_id is valid we will index Njp and the mmsid When Njp is not there and the mms_id is valid we will index the mmsid only.```

sandbergja:
Closed by #2422

PR #2422, opened 2024-07-24
[#2413] Index MMS IDs from 773 and 774 that have (NjP) prefix

sandbergja:
Also, refactor 773 and 774 indexing into its own class so that it's a little easier to reason about and test in isolation.

### DIFF
--- a/marc_to_solr/lib/linked_fields_extractor.rb
+++ b/marc_to_solr/lib/linked_fields_extractor.rb
@@ -1,0 +1,24 @@
+# This class is responsible for extracting and validating
+# Alma MMS IDs from the 773 and 774 MARC fields
+class LinkedFieldsExtractor
+  def initialize(record, field_tag)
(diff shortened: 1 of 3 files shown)


marc_to_solr/lib/linked_fields_extractor.rb
```diff
@@ -0,0 +1,24 @@
+# This class is responsible for extracting and validating
+# Alma MMS IDs from the 773 and 774 MARC fields
+class LinkedFieldsExtractor
+  def initialize(record, field_tag)
+    @record = record
+    @field_tag = field_tag
+  end
+
+  def mms_ids
+    valid_fields.map { |field| field['w'].delete_prefix('(NjP)') }
```
christinach, line 10: 👍🏻

christinach (approved):
Thanks @sandbergja

Issue #1929, opened 2022-07-29
Duplicate events can cause indexing to get stuck


sandbergja:
**Expected behavior**
Bibdata does not create duplicate events

**Actual behavior**
Bibdata created two identical events (1766 and 1767), which had the same timestamp and included files with identical checksums.

**Steps to replicate**

We're not sure what caused the duplicate events in the first place.

**Impact of this bug**

When it encounters these duplicates, bibdata gets into an endless loop indexing the two files over and over again, blocking indexing of later files.

See https://github.com/pulibrary/bibdata/blob/391211a3fc9fbc66b7a23ff2baa6f310ff8a37fe/app/models/index_manager.rb#L92

Check the precision of timestamps in the database.

Check how the SQS poller is polling files when for some reason disconnects and then reconnects.

Make sure when there is a table entry in the database with checksum X don't allow to save another file in an event with the same checksum.

sandbergja:
We also observed this with 5260 and 5261.

sandbergja:
We also saw this with 8338 and 8339

christinach:
See https://pulibrary.slack.com/archives/C031U4XCUH2/p1722278029276899 latest update on this one.

christinach:
- 90% of the events since July 24th were duplicate.
- I checked AWS sqs message queue and there was not a duplicate message that was sent from the webhook.
- I deleted all the duplicate events and load the files back on the bibdata VM mount directory /data/bibdata_files
- Also deleted older changed record events that were created before July 4th.
- With @kevinreiss @sandbergja @maxkadel we deployed to production and we didn't see any errors.
- Asked @mzelesky to Run an Adhoc incremental publishing job and to update the schedule of the jobs to be every 4 hours until further job schedule change.

christinach:
closed by https://github.com/pulibrary/bibdata/pull/2431

PR #2431, opened 2024-07-29
[#1929] Don't allow duplicate events

christinach:
Add migration to change event message_body to be unique
Add Rails model validation to prevent creating a duplicate event


sandbergja (approved):
🐝 🚴🏻 🦄 🍄

PR #2435, opened 2024-07-31
[#2434] Add constraint in Dumps table not to be null

christinach:
closes #2434 

Add constraint in Dumps table not to be null
in the model add validate presence: true
Update event factory where dump with id 1
Add additional spec in dump_spec.rb

PR #2438, opened 2024-08-02
In db seeds, make sure that each dump also has an event

sandbergja:
This is a new constraint introduced in #2435.  For me, it was causing the servers:start rake task to fail locally, so this commit makes our seed data matches our db constraints.


christinach (approved):
thanks @sandbergja

maxkadel (approved):
Thanks @sandbergja!

PR #2447, opened 2024-08-14
Create the event before creating the incremental SCSB and make sure the dump is connected to this event_id

christinach:
This is to fix an error we saw in the sidekiq logs, which was blocking SCSB incrementals from indexing.

```
[2024-08-14T15:05:14.414018 #873] ERROR -- : [ActiveJob] [ScsbImportJob] [ee5aab42-b9ba-4265-afe2-e10378594fd0] Error performing ScsbImportJob (Job ID: ee5aab42-b9ba-4265-afe2-e10378594fd0) from Sidekiq(default) in 0.56ms: ActiveRecord::RecordNotFound (Couldn't find Dump without an ID):
```

### DIFF
--- a/app/models/dump.rb
+++ b/app/models/dump.rb
@@ -39,9 +39,9 @@ class Dump < ActiveRecord::Base
       dump = nil
       timestamp = incremental_update_timestamp
       Event.record do |event|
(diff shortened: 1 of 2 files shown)


sandbergja (approved):
🐬 🍄

PR #2454, opened 2024-08-20
Update Rakefile to require_relative path to config

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #2460, opened 2024-08-21
Add logging when event fails validation

christinach:


maxkadel (approved):
Thanks @christinach 🐬 !

PR #2465, opened 2024-08-22
Rescue error in aws sqs_poller poll so that we delete the duplicate message and not enqueue it again

christinach:
Rescue error in aws sqs_poller poll so that we delete the duplicate message and not enqueue it again


sandbergja (approved):
🐬 🦄 🐈‍⬛ 🚴🏻 🍿 🍄

PR #2479, opened 2024-09-03
[#2433] Resource print materials with ai in position 06 and 07 is a databases format

### LINKED ISSUE #2433: Integrating resource print materials should be classed as 'databases'
## Notes
Records with 'ai' in the leader position 06-07 are currently classed as books. See https://catalog.princeton.edu/catalog/12142765

"In this case, the record represents a database of diary entries. There are also many integrating resources that would be more closely related to 'journals' than books. I am not sure if they should be marked as databases or journals, but classifying them as books seems inaccurate to me."

(Reported from Kevin Connell)

## Acceptance criteria
- [ ] Resources with "ai" in leader position 06-07 are given the format type of "Database"
- [ ] Let the ERM group know when this is going live in the catalog

christinach:
closes #2433 

Resource print materials with ai in position 06 and 07 is a databases format
add byebug in the gemfile

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -75,6 +75,7 @@ end
 
 group :development, :test do
   # bundler and rake come in from the voyager_helpers gemspec
+  gem 'byebug'
   gem 'coveralls_reborn'
   gem "dotenv-rails"
   gem "factory_bot_rails"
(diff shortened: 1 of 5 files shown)


sandbergja (approved):
Thanks, @christinach 🐬

PR #2481, opened 2024-09-04
This is an update on #2433

christinach:
This is an update on #2433
When there is an a in position 06 it doesn't consider an i position 07 from the bibformat_bk condition

https://catalog-staging.princeton.edu/catalog/99121427653506421


maxkadel (approved):
Thanks @christinach! 🐬

PR #2484, opened 2024-09-17
Change staging vms to have the lib private domain

christinach:


kayiwa (approved):
OK! Thanks @christinach

PR #2493, opened 2024-09-18
[#1827] Use the term "description", rather than "enumeration", for the combined enumeration and chronology

### LINKED ISSUE #1827: The build_items method needs to reference description instead of enumeration
https://github.com/pulibrary/bibdata/blob/99f589ae48a2f71ddaf83afa264ed87a05e495ac/marc_to_solr/lib/process_holdings_helpers.rb#L144

In this line, it should say item[:description] instead of item[:enumeration].

sandbergja:
This makes the terminology consistent with the Alma availability response, and the Alma enrichment UI.

It also reduces some potential confusion, since we no longer imply that chronology is not included in this value.

Closes #1827


christinach (approved):
Thanks @sandbergja

PR #2495, opened 2024-09-23
Re-enable https for `_url` helpers.

tpendragon:
These action mailer defaults control the protocol for `_url` helpers, which are used in the JSON endpoints Figgy depends on. When they all changed to `http` it broke Figgy.


christinach:
Deployed to prod and confirmed that this change makes no difference.
closing

PR #2496, opened 2024-09-24
set https protocol explicitly for index delivery locations

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #2501, opened 2024-10-03
Add logging for patron controller authorization

sandbergja:
This was in service of us better understanding an issue in which we were getting the load balancer IP addresses, rather than the catalog IP addresses, when the catalog made requests to bibdata.


christinach (approved):
Thanks everyone

PR #2503, opened 2024-10-04
Pin alma gem to 0.5.x 

christinach:
Pin alma gem to 0.5.x until we fix https://github.com/pulibrary/bibda…ta/pull/2502

Update alma gem
Remove the main branch reference from the Gemfile


sandbergja (approved):
Thanks, @christinach 🐬

PR #2513, opened 2024-10-08
Refactor use of #collect_matching_lines

sandbergja:
The #collect_matching_lines method automatically assembles an array made of the return values of the block, so we don't need a separate array to collect those values.

See the traject documentation: https://www.rubydoc.info/gems/traject/3.4.0/Traject/MarcExtractor#collect_matching_lines-instance_method


christinach (approved):
Thanks @sandbergja

PR #2516, opened 2024-10-10
Upgrading Orangetheses for indexing theses from DataSpace

jrgriffiniii:
Advances https://github.com/pulibrary/orangetheses/issues/85


christinach:
pu field error was fixed with [this orangetheses update](https://github.com/pulibrary/bibdata/pull/2518)

Issue #2034, opened 2022-11-29
New facet: geographic subject


escowles:
Feature request received by DSSG: Add a geographic facet, similar to Subject: Era, based on $z in 6xx fields. This would allow users to limit their results to items ABOUT a place, which is sometimes different than the Place of publication.

### Acceptance Criteria

- [x] Index from Alma Staging a set of records using 6xx $z (Bibdata)
- [x] Add Facet to Orangelight. 
- [ ] Preview results in catalog-staging for POs. 
- [ ] Decide if this shows up on the advanced search page or not. We do not have subject era on the adv. search now.

escowles:
DSSG discussion today: some catalogs have geographic subject facets (Cornell, Stanford, and Yale all display a geographic facet), but we're not sure whether they are using 6xx $z or 043 for the data. More investigation is needed to determine what is the best source of data for this facet.

kevinreiss:
DSSG will investigate and report back.

escowles:
Different people do it different ways. We'd like to try 6xx $z. Can we index that into a geographic facet in staging to see what the data looks like?

christinach:
Yes, we will index a data set in bibdata-staging or bibdata-qa.

sandbergja:
This is done!  Yay!

PR #2517, opened 2024-10-11
Add geographic_facet

maxkadel:
Connected to #2034

Handily, there's a traject macro for this already! See https://github.com/traject/traject/blob/5d720e2ba0a277cf7af455763f520cd6a2d956c7/lib/traject/macros/marc21_semantics.rb#L444

Indexed last two dumps on bibdata-staging and deployed branch on catalog-staging.

image


christinach (approved):
Thanks @maxkadel

PR #2529, opened 2024-10-16
[#2528] Don't include genres in Primary Source facet if only part of the word matches

### LINKED ISSUE #2528: "computer network resources" are being misclassified as primary sources
### Expected behavior
"computer network resources" are not considered to be primary sources unless they have one of the primary source genres in the record.

### Actual behavior

They are included -- I think because the word "resources" contains the word "sources".  See [MMS ID 99125445731306421](https://catalog.princeton.edu/catalog/99125445731306421) as an example.

### Steps to replicate

1. Go to https://catalog.princeton.edu/catalog?f%5Bgenre_facet%5D%5B%5D=Primary+source&search_field=all_fields&q=id%3A99125445731306421 -- this should be empty

### Impact of this bug

Primary source genre facet is less reliable.

sandbergja:
This fixes a bug in which a subdivision 'Computer network resources' matched against the genre 'Sources', since the word 'resources' contains the word 'sources'.

Closes #2528

### DIFF
--- a/marc_to_solr/lib/genre.rb
+++ b/marc_to_solr/lib/genre.rb
@@ -81,7 +81,7 @@ class Genre
 
     def genre_term_indicates_primary_source?(genre)
       normalized_genre = genre.downcase.strip.delete_suffix('.')
-      primary_source_genres.any? { |primary_source_genre| normalized_genre.include? primary_source_genre }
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja !

PR #2531, opened 2024-10-17
Primary source genre: add tests for Personal narratives and Personal narratives, Italian

sandbergja:
Thanks @mzelesky for bringing up that subdivisions like "Personal narratives" can also have forms like "Personal narratives, Italian", and we want to include those as well.

These tests checks that both "Personal narratives" and "Personal narratives, Italian" get the Primary Source genre applied.


christinach (approved):
Thanks @sandbergja ! 🍄

PR #2532, opened 2024-10-17
Use plural form Primary sources

sandbergja:
This makes it more consistent with other values in the genre facet, including the existing LCGFT term 'Primary sources'


christinach (approved):
Thanks @sandbergja !

PR #2533, opened 2024-10-17
Update primary sources genre tests to use the plural form

sandbergja:
This should get the tests green on main again.


christinach (approved):
Thanks @sandbergja 🍄

PR #2535, opened 2024-10-18
Part of https://github.com/pulibrary/orangelight/issues/4095

christinach:
part of #4095

Remove punctuation dot from the end of a fast subject heading string

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -895,7 +895,7 @@ to_field 'fast_subject_display' do |record, accumulator|
   next if subjects.present?
   record.fields.select { |field| field.tag[0] == '6' }.each do |field|
     next unless field['2'] == 'fast'
-    value = field['a']&.delete('.')
+    value = field['a']&.delete_suffix('.')
     accumulator << value
   end
 end


marc_to_solr/lib/traject_config.rb
```diff
@@ -895,7 +895,7 @@
   next if subjects.present?
   record.fields.select { |field| field.tag[0] == '6' }.each do |field|
     next unless field['2'] == 'fast'
-    value = field['a']
+    value = field['a'].delete('.')
```
sandbergja, line 898: Is there any chance `field['a']` could return nil?
```suggestion
    value = field['a']&.delete('.')
```

sandbergja (approved):
Thanks, @christinach !  Just one small question

marc_to_solr/lib/traject_config.rb
sandbergja: Oh, actually one more question sorry!  Wouldn't this remove periods within the subfield a, not just at the end?  I wonder if there are some fast headings that legitimately have "." inside them.

christinach: yes it would remove them. I handled the period failure in orangelight so maybe I can close this PR?

christinach: @sandbergja I will wait to discuss this with @mzelesky .

christinach: @sandbergja I used delete_suffix to delete the . at the end of the string.

sandbergja (approved):
Thanks, @christinach 🐬

PR #2540, opened 2024-10-23
Update indexing.md

christinach:
Latest changes in orangetheses, set the RAILS_ENV that points to a dspace domain accordingly.  Even though the RAILS_ENV in bibdata production is production . It is a good practice to set it when we run the rake task. 
For example when testing theses in bidbata-staging we should still run the theses rake task with RAILS_ENV=production so that it will retrieve theses from the production dspace and not the staging one.


sandbergja (approved):
Thanks, @christinach 🐬

Issue #2554, opened 2024-11-15
Change the label "Manuscript, Book" to "Manuscript, Archival Item" for single-item PULFA records


regineheberlein:
### Description

We designate single-item archival objects as leader 6/7 “t” and “m”: manuscript language materials and monograph/item. Based on those values, Orangelight applies the label “Manuscript/Book.”

For PULFA records with that leader, we’d want “Book” to be “Archival Item.” 

### Example
https://catalog.princeton.edu/catalog/99131250454706421

### Implementation
PULFA records have a 035$a that starts with `(PULFA)` and a 040$e set to [dacs](https://www2.archivists.org/groups/technical-subcommittee-on-describing-archives-a-content-standard-dacs/describing-archives-a-content-standard-dacs-second-). 

We can filter by those so we don't apply the label to actual books:

For any records with 
- 035$a starting with "(PULFA)" and
- 040$e being "dacs" or "appm" or being absent, and
- leader 6 "t" and
- leader 7 "m" : 
label for leader 7 should be "Archival Item" (not "Book")

### Acceptance criteria:
- [x] PULFA records matching those selection criteria display as "Manuscript, Archival Item", e.g. [this record](https://catalog.princeton.edu/catalog/99131250454706421)
- [x] non-PULFA records continue to display as "Manuscript, Book", e.g. [this record](https://catalog.princeton.edu/catalog/99125554155506421)
- [x] PULFA records not matching those selection criteria continue to display as they did before, e.g. [this record](https://catalog.princeton.edu/catalog/99125501041106421/staff_view) would continue to display as "Manuscript"
- [x] The format facet should display the same icon as "Manuscript" for the new format "Archival Item"

### Before merging to prod:
index a sample of data for review 

[Related Slack thread](https://pulibrary.slack.com/archives/C031U4XCUH2/p1731617436315399)

regineheberlein:
I've updated the above criteria for the $040 to include appm as well as an absent 040.

PR #2558, opened 2024-12-06
Changes label to archival item for PULFA record

ishasinha1:
Makes progress on #2554

### DIFF
--- a/marc_to_solr/lib/format/bib_format.rb
+++ b/marc_to_solr/lib/format/bib_format.rb
@@ -12,8 +12,10 @@ class BibFormat
     ldr = record.leader
     type = ldr[6]
     lev  = ldr[7]
+    check_pulfa = record['035'] && record['035']['a'].start_with?('(PULFA)')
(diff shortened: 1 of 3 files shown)


marc_to_solr/lib/format/bib_format.rb
```diff
@@ -9,76 +12,98 @@ class BibFormat
...
+    def lev
+      @lev ||= ldr[7]
+    end
+
+    def check_pulfa
+      @check_pulfa ||= record['035'] && record['035']['a'].start_with?('(PULFA)')
+    end
+
+    def check_dacs
+      @check_pulfa ||= record['035'] && record['035']['a'].start_with?('(PULFA)')
```
christinach, line 107: What is the difference between check_pulfa and check_dacs? 🤔 @sandbergja

sandbergja: Ooops, my mistake!  I'll fix that.

regineheberlein (approved):
Thank you @sandbergja !

PR #2562, opened 2024-12-16
[#2552] Don't stop the SCSB import full job if stale file is already deleted

### LINKED ISSUE #2552: Rescue if file is already deleted
### Expected behavior
If we're trying to delete a file and it's already deleted the application doesn't raise an error and doesn't re-try

### Actual behavior
If the code can't delete a file that's already been removed, it raises an error (I'm not sure how this happens, but as long as the file is deleted I don't think it matters - I'm guessing that somehow there's a race condition, or multiple servers are running this against the same filesystem at the same time)

### Impact of this bug
Noise in the logs, have to re-try the job.

### Acceptance criteria
- [x] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
- [x] If a file is deleted while the [delete_stale_files method](https://github.com/pulibrary/bibdata/blob/de797421a9e5980c2c67846b2ba482c4665d313a/app/jobs/scsb_import_full_job.rb#L20-L24) is being called, it does not stop the code. It does log the error.

### Honeybadger link and code snippet, if applicable
https://app.honeybadger.io/projects/54497/faults/114481686

```
line 22 of [PROJECT_ROOT]/app/jobs/scsb_import_full_job.rb: delete_stale_files
line 3 of [PROJECT_ROOT]/app/jobs/scsb_import_full_job.rb: perform
```

Rescue from a `Errno::ENOENT` error, log it, and continue

sandbergja:
closes #2552

### DIFF
--- a/app/jobs/scsb_import_full_job.rb
+++ b/app/jobs/scsb_import_full_job.rb
@@ -18,8 +18,14 @@ class ScsbImportFullJob < ApplicationJob
     end
 
     def delete_stale_files
-      FileUtils.rm Dir.glob("#{ENV['SCSB_PARTNER_UPDATE_DIRECTORY']}/*.zip")
-      FileUtils.rm Dir.glob("#{ENV['SCSB_PARTNER_UPDATE_DIRECTORY']}/*.xml")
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja ! 🍄 🍄

PR #2563, opened 2024-12-19
I2321 Shift SCSB full index tasks into separate background jobs

### LINKED ISSUE #2321: SCSB import job is blocking deployment
We run the SCSB import job to import full dump files from SCSB. The job is scheduled to run on a Saturday so that it does not get interrupted by a deployment.

Some times we need to run the job ad hoc during the week. This will block the deployment because the deploy will restart the worker and the SCSB import job.

Example of a failing SCSB import job because of a bibdata deployment.
 
Screenshot 2024-03-20 at 2 18 56 PM

## Acceptance Criteria

- [ ] Identify a strategy where this job can continue to run and not start from the beginning if the application is deployed or the server is restarted.

maxkadel:
This moves sub-tasks of the SCSB Full import into their own background jobs, so that 
1. They can run in parallel
2. The job will hopefully not block deployment for 12 hours

Closes #2321


rladdusaw (approved):
Thanks @maxkadel!

maxkadel:
Done again under a series of smaller PRs

PR #2564, opened 2024-12-19
Add rake task and cron job to check latest index time

### LINKED ISSUE #2329: Create an alert to check the homepage indexing process timestamp
Create an alert to check daily the bibdata homepage for a timestamp that is one day older. 
If the date on the timestamp is one day older, trigger the alert. Pay attention in the timezone.

rladdusaw:
Resolves #2329

### DIFF
--- a/lib/tasks/index_timestamp.rake
+++ b/lib/tasks/index_timestamp.rake
@@ -1,7 +1,7 @@
 namespace :index_timestamp do
   desc "Check the timestamp for the last index update and raise an error if it has not updated today"
   task check: :environment do
-    active_manager = IndexManager.all.first { |index_manager| !index_manager.solr_collection.match(/rebuild/) }
+    active_manager = IndexManager.all.select { |index_manager| index_manager.solr_collection.match(/catalog-alma-production$/) }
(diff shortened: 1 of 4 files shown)


lib/tasks/index_timestamp.rake
```diff
@@ -0,0 +1,7 @@
+namespace :index_timestamp do
+  desc "Check the timestamp for the last index update and raise an error if it has not updated today"
+  task check: :environment do
+    active_manager = IndexManager.all.first { |index_manager| !index_manager.solr_collection.match(/rebuild/) }
```
christinach, line 4: Can we check for solr-collection `catalog-alma-production`? 
Sometimes we might run an index test with a different solr-collection. This will create an index manager instance and because it won't have the '-rebuild' in the name, it will alert us for the wrong index manager instance.

christinach (changes requested):
@rladdusaw thank you for working on this PR! Can you please check my comment?

lib/tasks/index_timestamp.rake
rladdusaw: Done.

christinach (approved):
Thanks you @rladdusaw !

PR #2571, opened 2024-12-24
Update cimg ruby in bearer job

christinach:


sandbergja (approved):
Thanks, @christinach 🐬 🐻

PR #2573, opened 2024-12-30
Return an index_manager object

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬 🍄

PR #2578, opened 2025-01-02
Update config/test.rb with the correct value for config.action_dispatch.show_exceptions

christinach:
See: https://discuss.rubyonrails.org/t/behaviors-in-action-controllers-raising-error-has-been-changed-in-7-1/83933/2


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2579, opened 2025-01-02
Prep for more background jobs

maxkadel:
In preparation for #2321, change all jobs to Sidekiq jobs

- In order to use batches, we have to use Sidekiq jobs
- Having a more consistent API between jobs is helpful, both in the code and in the tests
- Reorganize jobs directory to make understanding the relationships between jobs clearer
- Remove unused job and queue


sandbergja (approved):
Thanks, @maxkadel 🦄

PR #2581, opened 2025-01-02
Embed Import::Partner::Full in batch with callback

maxkadel:
Embed Import::Partner::Full in batch with callback


christinach (approved):
Thanks @maxkadel

PR #2582, opened 2025-01-02
Remove log of record fixes

maxkadel:
These files consist of a JSON document that lists the type of fix and the ID of the record where the fix was made and no other context.

- As far as I can tell, we never use these, and it complicates breaking the updates into separate background jobs
- rubocop


kevinreiss (approved):

christinach (approved):
Thanks @maxkadel

PR #2586, opened 2025-01-03
Refactor unzipping and xml file cleanup

maxkadel:
Connected to #2321


app/models/scsb/partner_updates/update.rb
```diff
@@ -51,6 +51,18 @@ def process_partner_updates(files:, file_prefix: 'scsb_update_')
         end
       end
 
+      def self.process_record(record)
```
kevinreiss, line 54: This appears to be the same method as the one starting line 66 but called "self.", if we don't need the other one any longer we can remove.

kevinreiss (changes requested):
Thanks @maxkadel

app/models/scsb/partner_updates/update.rb
maxkadel: Done - the incremental update was still using the instance method, so just added that change to the refactor.

spec/jobs/import/partner/download_full_file_job_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+
+  let(:stub_zip) { Zip::File.open('spec/fixtures/scsb_updates/NYPL_20210430_015000.zip') }
+  let(:first_entry) { stub_zip.to_a[0] }
+  let(:second_entry) { stub_zip.to_a[1] }
+
+  before do
+    allow(Zip::File).to receive(:open).and_yield(stub_zip)
+    allow(stub_zip).to receive(:each).and_yield(first_entry).and_yield(second_entry)
```
sandbergja, line 18: Cool!

sandbergja (approved):
Thanks, @maxkadel !  🦄

PR #2588, opened 2025-01-07
Use more performant MARC::XMLReader parser

maxkadel:
In breaking down the SCSB full import into more smaller background jobs, we discovered that processing each XML file was taking up to 9 minutes, and this seems like somewhere we might be able to make improvements.

The Nokogiri parser is much more performant than the default parser for the `MARC::XMLReader`. In addition, `external_encoding` is not a valid option for this class. 

Prior to this change, on my local system, the test was taking about 98 ms, after this change, the test was taking about 71 ms.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -86,6 +86,7 @@ group :development, :test do
   gem 'faker'
   gem 'puma'
   gem 'rails-controller-testing'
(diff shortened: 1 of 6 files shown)


christinach (approved):
Thanks @maxkadel ! 🦄 🦄

Issue #2187, opened 2023-07-14
650 Indigenous Languages aren't always adding the correct facet 


kevinreiss:
From @minjiec I did find some odd cases where bibs should be tagged as "Indigenous Languages (Western Hemisphere)" but are not. 

First, it looks as though **$v Texts** is not working properly.

Among the results of [Algonquian languages—Texts](https://catalog.princeton.edu/catalog/?f[subject_facet][]=Algonquian+languages%E2%80%94Texts) (19)
[6 records](https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&f%5Bsubject_facet%5D%5B%5D=Algonquian+languages%E2%80%94Texts&f1=all_fields&q1=&op2=NOT&f2=all_fields&q2=%22Indigenous+Languages+%28Western+Hemisphere%29%22&op3=AND&f3=title&q3=&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&sort=score+desc%2C+pub_date_start_sort+desc%2C+title_sort+asc&search_field=advanced&commit=Search
) are not tagged

None of [Apache languages—Texts](https://catalog.princeton.edu/catalog/?f[subject_facet][]=Apache+languages%E2%80%94Texts) (3) is tagged.

Among the results of [Zapotec language—Texts](https://catalog.princeton.edu/catalog/?f[subject_facet][]=Zapotec+language%E2%80%94Texts) (42)
[7 records](https://catalog.princeton.edu/catalog?utf8=%E2%9C%93&f%5Bsubject_facet%5D%5B%5D=Zapotec+language%E2%80%94Texts&f1=all_fields&q1=&op2=NOT&f2=all_fields&q2=%22Indigenous+Languages+%28Western+Hemisphere%29%22&op3=AND&f3=title&q3=&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&sort=score+desc%2C+pub_date_start_sort+desc%2C+title_sort+asc&search_field=advanced&commit=Search) are not tagged

Now I know Blacklight does not distinguish between $v Texts and $x Texts, and for our tagging interest we are only interested in the former, but spot-checking found plenty of $v Texts in there.

[...]

PR #2589, opened 2025-01-08
Ensure we match indigenous languages from 650 with $v Texts.

maxkadel:
Match more loosely on subfield v, since many will end with a period.

Connected to #2187


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #2592, opened 2025-01-13
Update ruby version file

christinach:


sandbergja (approved):
Thanks, @christinach 🐬

PR #2595, opened 2025-01-13
Remove ddtrace and use datadog gem.

christinach:
Helps with ruby 3.4.1 upgrade


regineheberlein (approved):
Thanks @christinach !

sandbergja (approved):
Thanks, @christinach ! 🐬 🐶

PR #2605, opened 2025-01-22
Remove obsolete holding locations besides zobsolete$zned and zobsolete$zscl

christinach:
When you review this PR please confirm that I've removed only obsolete locations and that I've kept `zobsolete$zned`
`zobsolete$zscl`


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2613, opened 2025-01-29
[#2612] Lewis Library should not be a pickup for Annex locations

### LINKED ISSUE #2612: Remove Lewis Library as a pickup location for the Forrestal Annex locations
Remove Lewis Library as a pickup location for the Forrestal Annex locations

### Level of urgency
- [x] High

Lewis Library should not be a pickup location for the Forrestal Annex locations

christinach:
closes #2612


mzelesky (approved):
Looks correct to me.

Issue #2618, opened 2025-02-07
lc_facet indexes ":" for weird call numbers


maxkadel:
### Expected behavior
If we cannot parse a real call number for a record, do not index anything into the `lc_facet` field.

### Actual behavior
Some invalid call numbers result in `:` being indexed into the `lc_facet` field. 

### Steps to replicate

From search against production index:
Tunnel into production index: `ssh -L 1234:localhost:8983 pulsys@lib-solr-prod9`
http://localhost:1234/solr/catalog-alma-production/select?facet.field=lc_facet&facet.limit=99999&facet=on&fq=lc_facet%3A%22%3A%22&q=*%3A*&rows=10

A few current records that result in this behavior:
- 99130737156906421
```
050 1 0a| ISSN RECORD
050   4a| AN
```
- 99131315721806421
```
050 0 0a| IN PROCESS
```
- 99126842134206421
```
050 1 0a| ISSN RECORD
```
- 99125251881306421
```
050 1 0a| ISSN RECORD
```

### Impact of this bug
Cannot turn on the Blacklight-Hierarchy gem for call numbers in the catalog until this is remediated and existing values for `:` are removed from the Solr index. See [Orangelight issue](https://github.com/pulibrary/orangelight/issues/4735)

### Acceptance criteria
- [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
- [ ] Invalid call numbers do not result in lc_facets with only `:` in them

maxkadel:
Many thanks to @sandbergja for helping identify this bug

PR #2619, opened 2025-02-07
Do not index : for invalid call numbers

maxkadel:
Connected to #2618

### DIFF
--- a/spec/marc_to_solr/lib/config_spec.rb
+++ b/spec/marc_to_solr/lib/config_spec.rb
@@ -1096,7 +1096,7 @@ describe 'From traject_config.rb', indexing: true do
         expect(lc_facet).to be_nil
       end
 
-      it 'handles cases where the call number is invalid' do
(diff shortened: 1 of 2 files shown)


spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -1092,6 +1095,10 @@ def fixture_record(fixture_name, indexer: @indexer)
         lc_facet = @record_no_call_number['lc_facet']
         expect(lc_facet).to be_nil
       end
+
+      it 'handles cases where the call number is invalid' do
```
sandbergja, line 1099: Could we make this description a little more specific?
```suggestion
      it 'does not index data into the lc_facet field if the call number is invalid' do
```

sandbergja (approved):
Thanks, @maxkadel !  I had one tiny idea, definitely not a blocker.

sandbergja (approved):
Thanks, @maxkadel ! 🦄

Issue #2638, opened 2025-02-25
Create a mapping document that documents where Figgy Ephemera Data fields go into the Orangelight Solr Schema


maxkadel:
Example records that exist in Alma for items that represent what will be entered into the Figgy Epherea Collection can be found in this [dataspace community](https://dataspace.princeton.edu/handle/88435/dsp01g732d9019). Each item in the dspace community has an equivalent Alma MMS ID. Ignore the Research Data sub-community, this data will not be part of this future workflow in Figgy. 

### Example Records from the Staging Collection

- https://figgy-staging.princeton.edu/catalog/36bad44a-955c-48a5-a98c-0de20f0b1811.jsonld
- https://figgy-staging.princeton.edu/catalog/af4a941d-96a4-463e-9043-cfa512e5eddd.jsonld

kelea99:
Both @escowles and myself did a runthrough of the Orangelight Faceting and we both made recommendations on the best fields from ephemera forms in Figgy to map to those Facets. We both came up with the same recommendations, which I share here. We would like the following. Thank you!:

### Orangelight Facet Name | FIGGY EQUIVELANT FIELD NAME
- Access | ONLINE (default for all items. nothing to map here)
- Library | DO NOT MAP
- Format | DO NOT MAP (this will require discussions with POs and @escowles at a future date, do not prioritize)
- Publication Year | DATE CREATED
- Language | LANGUAGE (this is a controlled vocabulary in figgy)
- Subject: Topic | SUBJECT (@kevinreiss this is a nested controlled vocabulary, based off of LC subject headings. will probably require its own ticket)
- Subject: Genre | GENRE (this is a controlled vocabulary in Figgy)
- Subject: Era | DO NOT MAP
- Subject: Geography | GEO SUBJECT (this is a controlled vocabulary in Figgy)
- Recently Added | DATE MODIFIED (We also have date created, but modified might be more accurate here to reflect the changes?)
- Place of Publication | GEO ORIGIN (this is a controlled vocabulary in Figgy)

kelea99:
@escowles "Format" may be a required field. Would the best way forward be getting confirmation from POs/ @kevinreiss  that a temporary placeholder as "Archival Item" is appropriate for this material?

escowles:
@kelea99 Yes, I think it would be good to checkin with POs and Kevin about whether they should all be mapped to one thing, or if we want to have serials mapped to Journal vs. monographs mapped to Book, etc.

[...]

PR #2642, opened 2025-02-28
Map dspace ephemera collection fiels to existing marc record fields

christinach:
Create new fields when needed


christinach:
closing in favor of https://github.com/pulibrary/bibdata/issues/2638

PR #2653, opened 2025-03-11
Add new lc_subject_facet field.

christinach:
Helps with the vocabulary work https://github.com/pulibrary/orangelight/pull/3386 In this new field we index only the lc subject heading and the subdivisions So that when the user searches using the Details section, they can query solr for all the subject headings and their divisions.

This is needed for the Subject browse Vocabulary work.

example:
"lc_subject_facet": [
             "Booksellers and bookselling—Italy—Directories",
             "Booksellers and bookselling-Italy",
             "Booksellers and bookselling"
              ]

### DIFF
--- a/marc_to_solr/lib/princeton_marc.rb
+++ b/marc_to_solr/lib/princeton_marc.rb
@@ -287,6 +287,13 @@ def process_hierarchy(record, fields)
   end.compact
 end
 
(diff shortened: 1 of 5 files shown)


regineheberlein (approved):
Thanks @christinach !

PR #2654, opened 2025-03-12
Set the correct expected lc_subject_facet

christinach:
When the record has different 6xx lc subjects should accumulate the subjects per field.


mzelesky (approved):
This looks to be accomplishing the goal of having LC subjects divided without concatenating all the fields into one long subject.

PR #2655, opened 2025-03-12
Add new facet fields that will be used for the subject browse table and the facet search

christinach:
Helps with vocabulary work https://github.com/pulibrary/orangelight/pull/3386

rbgenr_genre_facet
aat_genre_facet
lcgft_genre_facet
homoit_genre_facet
homoit_subject_facet
local_subject_facet
siku_subject_facet

@mzelesky could not find a record with local subject vocabulary field. We will add a fixture when we have one.


maxkadel (approved):
Thanks @christinach! 🐬

PR #2659, opened 2025-03-19
Index new field fast_subject_facet

christinach:
Helps with https://github.com/pulibrary/orangelight/pull/3386


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2660, opened 2025-03-21
Don't use dotenv gem for test environment

sandbergja:


christinach (approved):
Thank you!

@sandbergja @kevinreiss @maxkadel @rladdusaw @mzelesky

PR #2662, opened 2025-03-21
Only measure code coverage in CI by default

sandbergja:
This reduces performance overhead in TDD cycles, where measuring code coverage is not very useful.  This matches the Orangelight configuration.


christinach (approved):
thanks @sandbergja

Issue #2648, opened 2025-03-10
Make test loading faster by not requiring all files in spec/support for every test


maxkadel:
### What maintenance needs to be done?
Make test loading faster - allows for faster iterating using TestDriven Development

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

### Why is this maintenance needed?
- We have a good number of support files in spec/support, and right now all of them are being loaded for every test; however, many of these support files are only needed for a small subset of tests. We should load support files in a more tightly scoped way. 

Either some of the support files should be moved to other directories, or we should load support files on a per-test basis. See [line where we load all files in support directory](https://github.com/pulibrary/bibdata/blob/03ee0d0c9f1349c43ca94eaaa8cd398b22ba188b/spec/rails_helper.rb#L18) right now.

### Acceptance criteria
 - [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
 - [x] File loading takes less time (when you run a single test, there should be a spot in the response that says something like `(files took 6.81 seconds to load)`
 - [x] Tests still pass

### Implementation notes, if any
See [line where we load all files in support directory](https://github.com/pulibrary/bibdata/blob/03ee0d0c9f1349c43ca94eaaa8cd398b22ba188b/spec/rails_helper.rb#L18) right now.

sandbergja:
On my laptop, the load time for `spec/models/user_spec.rb` has been reduced from a median 7.1 seconds to a median 4.8 seconds.  This was achieved through a combination of:

- Removing unneeded gems
- Selectively loading the traject config
- https://github.com/pulibrary/orangetheses/pull/91
- Removing unneeded support code
- Selectively loading support code (rather than the suggested approach of loading files selectively, this is instead accomplished with RSpec's `when_first_matching_example_defined` feature)

PR #2664, opened 2025-03-21
Remove EZID stub

sandbergja:
It was only used by one test, a pending voyager-specific one. This PR removes the stub and the pending test.

Helps with #2648


maxkadel (approved):
Thanks @sandbergja! 🍄

PR #2667, opened 2025-03-21
Only load global indexer service for tests that need it

sandbergja:
Building this service can be somewhat expensive, since it reads through the entire traject configuration.


kevinreiss (approved):
Thanks @sandbergja

PR #2668, opened 2025-03-21
Address rails deprecations

sandbergja:
* enum has a slightly different API (see https://github.com/rails/rails/pull/50987)
* you now should specify a coder for serialized columns (see https://github.com/rails/rails/pull/47463).  Use YAML for backwards compatibility.
* we weren't actually using one of the serialized columns (`recap_barcodes`), and it doesn't even exist in the database.  Remove it from the model.


christinach (approved):
Thanks @sandbergja 🍄

PR #2676, opened 2025-03-26
Use ChangeTheSubject for geographic subjects

maxkadel:


spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -582,6 +582,15 @@ def fixture_record(fixture_name, indexer: @indexer)
...
       end
+
+      describe 'local overrides' do
+        let(:f650) { { '650' => { 'ind1' => ' ', 'ind2' => '0', 'subfields' => [{ 'a' => 'Oil spills', 'z' => 'America, Gulf of' }] } } }
+
+        it 'creates a facet for local marc geographic locations' do
+          record = @indexer.map_record(MARC::Record.new_from_hash('fields' => [f650], 'leader' => leader))
+          expect(record['geographic_facet'].first).to eq('Mexico, Gulf of')
+        end
+      end
```
sandbergja, line 593: Not a blocker, but I think it would be nice to have a test for a 651$a geographic facet as well.

sandbergja (approved):
Thanks, @maxkadel !  Looks great!  I had one idea for another test that could be added, definitely not a blocker. 🦄

spec/marc_to_solr/lib/config_spec.rb
maxkadel: Done

sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #2680, opened 2025-03-31
Add '(unspecified)' for unknown state/province within known country

### LINKED ISSUE #2678: When Place of Publication is unspecified within larger country, include "[place] (unspecified)"
The `publication_place_hierarchical_facet` has 5 countries where it includes sub-locations within the country (US, UK, Canada, Australia, and USSR), indicated with an 008 code starting with `xx`, e.g. unknown within the us would be `xxu`.

When the country is known but the subdivision is not known, we should label them "Country (unspecified)", e.g. `xxu` would be labeled "United States (unspecified)"

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

- It is confusing to have a top level facet for "United States" with 6 million records, and then a sub-facet within that for "United States" with only 73 thousand records.

### Acceptance criteria
 - [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.
 - [ ] The labels for unknown-within-country locations include "(unknown)" to distinguish them from the top-level facets.

Should change labels in `marc_to_solr/translation_maps/marc_countries_hierarchical.yaml`

maxkadel:
Closes #2678

### DIFF
--- a/marc_to_solr/translation_maps/marc_countries_hierarchical.yaml
+++ b/marc_to_solr/translation_maps/marc_countries_hierarchical.yaml
@@ -375,10 +375,10 @@
 'xs': 'South Georgia and the South Sandwich Islands'
 'xv': 'Slovenia'
 'xx': 'No place, unknown, or undetermined'
-'xxc': ['Canada', 'Canada:Canada']
-'xxk': ['United Kingdom', 'United Kingdom:United Kingdom']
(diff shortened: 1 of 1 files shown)


sandbergja (approved):
Thanks, @maxkadel ! 🦄

PR #2683, opened 2025-04-01
Fix running marc_to_solr tests in isolation

sandbergja:
This fixes a problem I introduced in #2667 -- apologies!  Running a single spec within marc_to_solr would fail locally with an Uninitialized Constant error, while running the whole suite would pass.  This commit fixes this issue by loading those classes earlier in the process (i.e. not waiting for RSpec's `when_first_matching_example_defined` hook to load them).

While I was checking each of these tests to confirm that they load all dependencies correctly, I also switched them to use the spec_helper instead of rails_helper if they don't need all the Rails classes. It makes running an individual test take well under a second, rather than the 6-8 seconds that are typical when using the rails_helper.


kevinreiss (approved):

kevinreiss:
Thanks @sandbergja

Issue #2679, opened 2025-03-31
Bug in Place of Publication Facet Delimiter


kevinreiss:
### Actual behavior
<img width="652" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGw3Q1E9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--1e8d1dba81865116f89f5aaf3eee7a459c517097/Screenshot%202025-03-31%20at%2010.43.48%E2%80%AFAM.png" alt="Screenshot 2025-03-31 at 10.43.48 AM.png" />

### Steps to replicate
Look at the display of facet "China (Republic : 1949- )". This should not be interpreted as a hierarchical facet. 

### Impact of this bug
Confusing UI. 
### Acceptance criteria

- [x] Example facet "China (Republic : 1949- )" does not display as a hierarchical facet after re-index. 
- [x] This update doesn't break the UI while we are waiting for a full re-index. 
- [x] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.

### Implementation notes, if any
We are using a ":" as a delimiter, need to use a character that does not appear in any of the facet string values. It's possible to use multiple characters as a delimiter. We will need to adjust a catalog controller display configuration for the facet to use the new delimiter.

PR #2684, opened 2025-04-01
Create new fields for pipe delimited hierarchical facets

maxkadel:
Since colons were present in the data, they did not make a good delimiter for hierarchical fields.

Connected to #2679

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -982,6 +982,14 @@ to_field 'lc_subject_facet' do |record, accumulator|
   accumulator.replace(lc_subjects)
 end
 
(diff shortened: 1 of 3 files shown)


marc_to_solr/lib/traject_config.rb
```diff
@@ -1074,15 +1082,31 @@
   end
 end
 
+to_field 'lc_pipe_facet' do |record, accumulator|
+  delimiter = '|'
```
sandbergja, line 1086: Can we use a multi-character delimiter (e.g. `|||`)?  I just dread the day when the data gets a value with a pipe, and we have to do this again.

sandbergja (commented):
Thanks, @maxkadel !  One idea, otherwise this looks great.

sandbergja (approved):
Yay, thanks @maxkadel 🦄

PR #2685, opened 2025-04-02
Update orangetheses

sandbergja:
I confirmed that I could still run `bundle exec rake orangetheses:cache_theses` on staging with this deployed, and that it produced a file of comparable size to recent production runs.


christinach (approved):
thanks @sandbergja !

PR #2686, opened 2025-04-02
Only require sidekiq testing for tests that need it

sandbergja:
Helps with #2648


christinach (approved):
nice @sandbergja . thanks!

PR #2687, opened 2025-04-02
Remove redundant require

sandbergja:
Rails requires gems by default, so there is no need to require it again.


christinach (approved):
thanks 🍄

PR #2691, opened 2025-04-15
Improve performance of change_the_subject

christinach:
After recent changes in change_the_subject https://github.com/pulibrary/change_the_subject/commit/d98cb217e46cae75d4b021d398abc8bae264d83d we noticed that the full reindex took double the time to complete. @sandbergja and @christinach used ruby-prof to index 978 records and found that psych parser calls 1001314 times -> *Psych::Visitors::Visitor#visit After creating a new instance of ChangeTheSubject in traject_config we can reuse it in every field that needs to call ChangeTheSubject

Running ruby-prof with this change to index 978 records the psych parser calls 1 time -> *Psych::Visitors::Visitor#visit

command: bundle exec ruby-prof $(bundle show traject)/bin/traject -- -c marc_to_solr/lib/traject_config.rb /denali_gulf.xml -u http://localhost:54709/solr/bibdata-core-development

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -42,6 +42,7 @@ $LOAD_PATH.unshift(File.expand_path('..', __dir__)) # include marc_to_solr direc
 
 augment_the_subject = AugmentTheSubject.new
 language_service = LanguageService.new
+change_the_subject = ChangeTheSubject.new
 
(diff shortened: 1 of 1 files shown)


sandbergja (approved):
Thanks, @christinach 🐬 🍄

kevinreiss (approved):
Thanks for investigating this @christinach and @sandbergja.

Issue #2639, opened 2025-02-25
Create a task that executes a full re-index of the (Formerly Dspace) Born Digital Monographs, Serials, & Series Reports Ephemera Collection


maxkadel:
### Purpose

***Update work on testing the indexing of [this collection](https://figgy.princeton.edu/catalog/3eb73e36-d829-4e39-bb74-3b5eabc76d67) should move to figgy-prod.****

A new workflow is being established to use an [Ephemera collection ](https://figgy.princeton.edu/catalog/3eb73e36-d829-4e39-bb74-3b5eabc76d67)to catalog born digital materials that were previously cataloged in both Dataspace and Alma. New materials in this category will be cataloged and indexed in the catalog with this new indexing routine. 

### Acceptance Criteria

- [x] Task pulls all "complete" Ephemera Folders from the [(Formerly Dspace) Ephemera collection](https://figgy-staging.princeton.edu/catalog.json?f%5Bephemera_project_ssim%5D%5B%5D=Born+Digital+Monographs%2C+Serials%2C+%26+Series+Reports&f%5Bhuman_readable_type_ssim%5D%5B%5D=Ephemera+Folder&f%5Bstate_ssim%5D%5B%5D=complete&per_page=100&q=) in JSON. 
- [x] Tasks loops through each item of the result and then indexes the jsonld document representing each item in the SOLR index using the [mapping worksheet](https://docs.google.com/spreadsheets/d/1UlfW-sb125tnL62kJFQtKCgzz9TuUkT3NFJKlZaLC7c/edit?gid=2039455207#gid=2039455207). Note at present only the first 18 rows are relevant for this initial implementation.  
- [x] Task is implemented and tested in Bibdata Staging with the [Production collection](https://figgy.princeton.edu/catalog/3eb73e36-d829-4e39-bb74-3b5eabc76d67). 
- [x] Task is implemented and tested in Bibdata QA with [Production collection](https://figgy.princeton.edu/catalog/3eb73e36-d829-4e39-bb74-3b5eabc76d67). 
- [x] Task is implemented and tested in Bibdata Production with [Production Collection](https://figgy.princeton.edu/catalog/3eb73e36-d829-4e39-bb74-3b5eabc76d67). 

### Implementation Notes
[...]

PR #2704, opened 2025-05-07
Combine the values from the alternative and transliterated_title - other_title_display

christinach:
towards #2639


sandbergja (approved):
Thanks, @christinach !  🍄 🐬 🦀

PR #2714, opened 2025-05-12
Remove id from EphemeraItem struct. We don't use it. Remove Attributes.

christinach:
towards #2639


sandbergja (approved):
Thanks, @christinach 🐬

PR #2715, opened 2025-05-12
Bring Senior theses indexing into bibdata

sandbergja:
Closes #2522

The behavior of this new Rust indexer is meant to be exactly the same as the orangetheses indexer, except that Dataspace's access restrictions field will be indexed as requested by SC.

Note that, like orangetheses, this indexer logs various messages at the DEBUG and INFO log level.  However, they aren't visible unless you set the `RUST_LOG` environment variable to `debug`.  Maybe we should add environment variable that to the boxes with ansible?

It won't let you Ctrl+C for some reason?

PR #2718, opened 2025-05-13
Add doc for theses

christinach:


christinach:
it was included in https://github.com/pulibrary/bibdata/pull/2715

PR #2723, opened 2025-05-19
Rescue from error for development

maxkadel:
If the developer was off VPN and did not have the proper key in their environment the migrations would error. This file is not necessary for all developers, so rescue and output info for migrations.


christinach (approved):
thanks @maxkadel !

PR #2724, opened 2025-05-20
Only compile regular expression once

sandbergja:
This is a [recommendation from the regex crate documentation](https://docs.rs/regex/latest/regex/#avoid-re-compiling-regexes-especially-in-a-loop).

It speeds up the included microbenchmark substantially, from a median of 26 microseconds to 587 nanoseconds.


christinach (approved):
Thanks @sandbergja !

PR #2725, opened 2025-05-20
Remove lazy_static dependency

sandbergja:
`LazyLock` from the standard library fulfills the same purpose.


christinach (approved):
Thanks @sandbergja !

PR #2733, opened 2025-05-30
Update theses indexing doc with Rust way

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2734, opened 2025-05-30
Don't pass chunk_size in read_ephemera_folders

christinach:
towards #2639 

Don't pass chunk_size in read_ephemera_folders
Make test test_ephemera_folders_iterator to pass


lib/bibdata_rs/src/ephemera/ephemera_folder.rs
```diff
@@ -102,7 +98,6 @@ mod tests {
         .await;
     }
 
-    #[ignore]
```
sandbergja, line 105: 🥳

sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2738, opened 2025-05-31
don't let one preserving_envvar test failure cause additional test failures

sandbergja:
Previously, since we use a `Mutex` to make sure that multiple test threads aren't modifying the same environment variable simultaneously, a failing test caused a Cache Poisoning error, indicating that perhaps the other test failed to put its environment variable back to its original value.

However, we can safely ignore this Error (rather than unwrap and cause a panic), since the original value of the environment variable does not matter for our purposes.

Also, use a type alias and better variable names for clarity.


christinach (approved):
Thanks @sandbergja ! 🍄

PR #2741, opened 2025-06-02
Test that we can post a solr document

christinach:


kevinreiss (approved):
Thanks @sandbergja and @christinach.

PR #2742, opened 2025-06-02
[#2639] add new parameter in the index method

christinach:
part of #2639 

Add parameter: collection in bibdata/lib/bibdata_rs/src/solr/index.rs#index method Make test to pass


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2745, opened 2025-06-04
[#2719] Don't default to english in language_iana_s field if other IANA languages are available

### LINKED ISSUE #2719: language_iana_s is set to English when the first language code does not have a 2-letter equivalent
### Expected behavior

If a record has several language codes with valid IANA equivalents, and English is not one of the languages in the record, we should index one of the languages from the record into the language_iana_s field.  

### Actual behavior

If the first language on the record happens to not have a 2-letter equivalent (for example `ota`), none of the languages from the record will be indexed, they will instead get a language_iana_s of `en`.

### Steps to replicate

1. Go to https://catalog.princeton.edu/catalog/9975186303506421/staff_view
2. Notice that the record has the language codes `ota`, `per`, and `ara`
3. Look at the [raw json of this record](https://catalog.princeton.edu/catalog/9975186303506421/raw.json)
4. Notice that the language_iana_s is `en`

### Impact of this bug

Many records in DPUL erroneously say they are in English, rather than one of the languages from the record.

### Acceptance criteria
- [ ] Records with non-English languages don't get erroneously indexed with language_iana_s `en`
- [ ] If this change touches any of the files referenced in `.rubocop_todo.yml` create a new ticket to address the rubocop  todo.

### See also
* https://github.com/pulibrary/figgy/issues/6697

sandbergja:
Closes #2719

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -76,6 +76,10 @@ Metrics/MethodLength:
   Exclude:
     - 'marc_to_solr/lib/princeton_marc.rb'
 
+Rails/NegateInclude:
+  Exclude:
+    - 'marc_to_solr/**/*' # exclude? is a Rails method that is not included in marc_to_solr
+
 Rails/HasAndBelongsToMany:
   Exclude:
     - 'app/models/delivery_location.rb'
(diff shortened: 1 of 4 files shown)


kevinreiss (approved):

kevinreiss:
Thanks @sandbergja 🐬

PR #2751, opened 2025-06-05
Refactor solr format from a string to an enum

sandbergja:
This means that the compiler can catch us if we try to index a document with an incorrect format.  It also removes a few calls to clone.


christinach (approved):
Thanks @sandbergja !

PR #2753, opened 2025-06-05
part of [#2639] Map provenance value from figgy jsonld

christinach:
part of #2639 
* to provenance display

### DIFF
--- a/lib/bibdata_rs/src/solr/ephemera_solr_mapping.rs
+++ b/lib/bibdata_rs/src/solr/ephemera_solr_mapping.rs
@@ -169,7 +169,7 @@ mod tests {
             Some("Test name".to_owned())
         );
     }
-
+#[test]
     fn it_has_the_publisher_from_the_ephemera_folder_item() {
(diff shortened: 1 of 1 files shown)


lib/bibdata_rs/src/solr/ephemera_solr_mapping.rs
```diff
@@ -154,6 +155,21 @@ mod tests {
...
+            .build()
+            .unwrap();
+        let solr_document = SolrDocument::from(&ephemera_item);
+        assert_eq!(solr_document.id, "abc123");
+        assert_eq!(
+            solr_document.provenance_display,
+            Some("Test name".to_owned())
+        );
+    }
+
```
sandbergja, line 172:
```suggestion
#[test]
```

It looks like the `#[test]` attribute is missing from the next test now

sandbergja (changes requested):
Thanks, @christinach ! 🐬   Just one small thing I noticed

PR #2755, opened 2025-06-05
part of [#2639] Map subject value from jsonld

christinach:
* create a struct for subject and exact_match
* update solrDocument to include lc_subject_display
* Add impl method to check for accepted vocabulary loc or homosaurus


sandbergja (approved):
🐬 🚴🏻 🐈‍⬛ 🍄

PR #2760, opened 2025-06-05
Refactor facet value enums

sandbergja:
- Add a new `AccessFacet` enum to represent the access facet and its two possible values: "Online" or "In the library".  This reduces some allocations and speeds up the `solr_document_from_dataspace_document/from` benchmark by 2%.
- Change the name of the existing `Format` enum to `FormatFacet`, so that it's not so easy to confuse it with the `Format` struct in Ephemera


kevinreiss (approved):
Thanks @sandbergja.

PR #2764, opened 2025-06-06
pub_date_*_sort fields should be single-valued

sandbergja:
This also means we can get rid of some `.clone()`s!


christinach (approved):
thanks @sandbergja !

PR #2766, opened 2025-06-06
[#2639] Map subject field from figgy JSonld to

christinach:
*homoit_subject_facet and homoit_subject_display


sandbergja (approved):
THanks, @christinach ! 🐬

PR #2771, opened 2025-06-09
Add some MARC basics to rust

sandbergja:


christinach (approved):
Thank you @sandbergja ! I love it!

PR #2775, opened 2025-06-10
Deploy: ensure that rust compilation happens before rails assets precompile

sandbergja:
Resolves this failed deployment: https://ansible-tower.princeton.edu/#/jobs/playbook/21414/output


christinach (approved):
thanks @sandbergja  !

PR #2810, opened 2025-08-01
Index the original language of translations as a facet

sandbergja:
This indexing happens in Rust.  To provide a quick lookup by language code, this commit includes an automatically generated Rust function that is based on data from the Library of Congress.  To refresh this function, you can run the provided rake task.

Advances https://github.com/pulibrary/orangelight/issues/4634


lib/tasks/languages.rake
```diff
@@ -1,3 +1,5 @@
+require_relative '../bibdata_rs'
+
 namespace :languages do
   namespace :iso639_5 do
     desc 'refresh the list of collective language codes in this repo from LOC'
```
christinach, line 5: thank you for this test

lib/bibdata_rs/utils.rb
```diff
@@ -0,0 +1,60 @@
+# This file contains utilities that are useful for maintaining Bibdata's Rust code
```
christinach, line 1: 💯

lib/bibdata_rs/src/languages/iso_639_2b.rs
```diff
@@ -0,0 +1,1952 @@
+// This file is automatically generated.  Please do not edit it directly,
+// instead run `bundle exec rake languages:iso639_2b:refresh_list`
```
christinach, line 2: 👍🏻

PR #2820, opened 2025-08-07
[#2639] Add page_count in ephemera_folder

christinach:
part of #2639 

Add page_count in ephemera_folder
Combine page_count and origin_place_labels to description_display


regineheberlein (approved):
thanks @christinach !

PR #2821, opened 2025-08-07
Remove duplicate impl EphemerFolder namespace

christinach:


sandbergja (approved):
Thanks, @christinach !  🐬   It's nice to have these two `impl` blocks combined into one.

PR #2825, opened 2025-08-07
[#2639] Ephemera should map the value: Online in the access facet.

christinach:
Ephemera should map the value: Online in the access facet.

part of #2639


sandbergja (approved):
Thanks, @christinach ! 🐬

rladdusaw (approved):
Thanks @christinach!

PR #2826, opened 2025-08-07
[#2639] Map sort_title from ephemera into title_sort Solr field

christinach:
part of #2639


rladdusaw (approved):
Thanks @christinach @sandbergja!

PR #2832, opened 2025-08-11
Remove notes field from SolrDocument

christinach:
It does not exist as a field in solr


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2834, opened 2025-08-11
Correct the Deserialization of solr::ElectronicAccess

sandbergja:
In the current ephemera indexing routine, we deserialize a JSON solr document from a string to a SolrDocument struct, then serialize it back to a string for POSTing to solr.

Eventually, we should not need to have this round-trip at all.  In the meantime, the electronic_access1display field does not survive the round-trip.  This is because this field is presented in the JSON in an odd format that combines a map and an array.  Prior to this commit, there was a custom Serialize implementation that handled this on the serialize side, but we were relying on the derived Deserialize implementation which had no idea about this unusual structure.

This commit adds a custom Deserialize implementation, which allows the ephemera full reindex rake task to run without error.


christinach (approved):
Thanks @sandbergja !!

PR #2836, opened 2025-08-12
Give more details on parsing failures when using reqwest's json() method

sandbergja:
By default, when you call `.json()` on a reqwest response that contains invalid JSON, you get a `reqwest::Error` that displays a very vague error message 'error decoding response body'

We don't have anything sensitive in any of our existing calls, and getting more information about the parsing failures would be extremely helpful.  Therefore, this commit maps these errors to `anyhow` errors that contain all of the useful debugging information.


christinach (approved):
Thanks @sandbergja ! This is very helpful!!

PR #2837, opened 2025-08-12
Normalize figgy ephemera id. Use only the uuid.

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2838, opened 2025-08-12
Rename Figgy env variable.

christinach:
We have FIGGY_URL in the vms
The ephemera task is expecting a FIGGY_URL


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2839, opened 2025-08-12
Include title_display in the ephemera_solr_mapping

christinach:


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #2842, opened 2025-08-13
Remove unreachable code from BibTypes

sandbergja:
These code paths were only called for certain values of bib_format that could never exist based on the logic in the BibFormat class.

Also, refactor two `.each` loops to `.any?` because rubocop started complaining about them. 🤖


christinach (approved):
Thanks @sandbergja !

Issue #2524, opened 2024-10-15
Develop Process to Index Figgy Ephemera Collection for Library Miscellaneous Items. 


kevinreiss:
## User Story
As a library researcher I want to have access to the Library's miscellaneous objects that have been cataloged using a Figgy ephemera collection after the dataspace system is sunset. 

## Acceptance Criteria
- [ ] Confirm that existing dataspace Library miscellaneous items can be ingested in Figgy as scanned resources linked to bibliographic descriptions in Alma. 
- [ ] Select sample items to index from the existing Dspace Corpus for our test ephemera items that will stand in for "newly" cataloged items once dataspace is shutdown. 
- [ ] Create sample figgy ephemera collection for miscellaneous items 
- [ ] Create a task that pull the CSV of this new collection on a daily basis
- [ ] Create a task that indexes the CSV into Orangelight

## Notes
Will need to consult with data creators and product owners on how to map emphemera data into the OL index.

kevinreiss:
Closing this ticket as we have tickets with more specificity for two processes, one that is a bulk index of all items in the collection, a second that is for event based indexing like we have for individual numismatics items.

PR #2843, opened 2025-08-13
In FormatFacet map the value 'Pamphlets' to a 'Book' format facet

christinach:
part of [#2524]


sandbergja (approved):
Nice, thanks @christinach ! 🐬

PR #2847, opened 2025-08-14
pub_created_display mapped solr field

christinach:
pub_created_display mapped solr field
should include date_created and origin_place
pub_created_display displays on the record page

related to #2524


rladdusaw (approved):
Thanks @christinach!

PR #2854, opened 2025-08-15
Include worker role in whenever_roles config/deploy.rb

christinach:
closes #https://github.com/pulibrary/orangelight/issues/5060

confirmed that it adds the figgy_mms_ids:build_translation_map task in the crontab


rladdusaw (approved):
Thanks!

PR #2863, opened 2025-08-19
Add the rest of the genre list in the format_facet.rs mapping

christinach:
Update from_str to compare lower case strings

related to #2639


regineheberlein (approved):
Thanks @christinach

PR #2870, opened 2025-08-19
Refactor items_by_852 method

christinach:
Extract code and write non_private_items in Rust


sandbergja (approved):
Thanks, @christinach !  🐬 🍄

PR #2874, opened 2025-08-20
Migrate oclc_number? method to Rust

sandbergja:
The rust implementation is faster, for a file of ~50,000 records, the ruby implementation takes .619 seconds while the rust implementation takes .162 seconds.


christinach (approved):
thanks @sandbergja !

PR #2880, opened 2025-08-21
Re-implement pub_created_display in rust

sandbergja:

PR #2881, opened 2025-08-21
Add parallel non-Latin publication statements back

sandbergja:
PR #2880 introduced a regression by neglecting to include non-Latin publication statements from the 880 fields.

This commit upgrades marctk to 0.6.0, which includes a convenience function for retrieving data from 880 fields, and fixes the regression by adding the non-Latin statements back.


christinach (approved):
Thanks @sandbergja !

PR #2886, opened 2025-08-21
Write build_call_number in rust

christinach:
Replace build_call_number in process_holdings.rb


sandbergja (approved):
Thanks, @christinach !  🐬 🍄

PR #2894, opened 2025-08-22
Migrate valid_language_code? to Rust

sandbergja:
Rewrite the ISO 639-5 rake task to create Rust structs rather than simply downloading the ISO 639-5 data file.

The Rust implementation is 3 orders of magnitude faster than the Ruby implementation!  The Ruby takes 77.262 seconds on a file of ~50,000 records, the Rust takes 0.054 seconds.


christinach (approved):
Thank you @sandbergja !

Issue #2888, opened 2025-08-21
[Ephemera] Index the manifest url from ephemera.


christinach:
## Current behavior
To construct the viewer we need the manifest url.
In orangelight we use graphql to send record ids to figgy and we use the figgy manifest manager to construct the container using the response from figgy that has the manifest url. 

## Desired behavior
Index the manifest url from ephemera into the electronic_access1_display with label 'Digital content'
Also update the electronic_access_1display to have the following structure:
```
"electronic_access_1display":
"{
"https://figgy.princeton.edu/catalog/fa30780e-dfd8-4545-b1b0-b3eec9fca96b":["Online Content", "Born Digital Monographic Reports and Papers"]}"
"https://catalog-staging.princeton.edu/catalog/ephemeraFolderID#view":["Digital Content"]
"iiif_manifest_paths":
{"ephemera_ark":"https://figgy.princeton.edu/concern/ephemera_folders/ephemeraFolderId/manifest"}
}"
```

PR #2895, opened 2025-08-22
Update ElectronicAccess Serialize to consider the rest of the properties. digital_content and iiif_manifets

christinach:
Set an index so that the resulting hash has a specific order. 
related to #2888


sandbergja (approved):
Thanks, @christinach ! 🐬 🍄

kevinreiss (approved):

Issue #2898, opened 2025-08-26
Update ephemera mapping


christinach:
- [x] Index the thumbnail. We're indexing the manifest url and use it to render the viewer. We want to index the thumbnail url and use it to dender the thumbnail
- [ ] If there are more than 1 manifest urls for the same ephemera folder then update the manifest_url_paths to be `ephemera_ark<incremental_number>`. The ephemera resources dont have an ark. Not yet. So we're hardcoding 'ephemera_ark' as a key. In catalog we're parsing the JSOn of the solr field. For that reason ephemera_ark needs to be different for each different manifest url otherwise the last one wins while parsing and we're loosing a manifest url.
- [x] Check why homosaurus subjects is displayed in the ephemera and update codebase. resolved by https://github.com/pulibrary/bibdata/pull/2902

related to #2639

christinach:
> [@christinach](https://github.com/christinach) [@kelea99](https://github.com/kelea99) is it realistic that there will be more than a single manifest for ephemera? I know this happens for MMS IDs but it seems unlikely with this material?

I don't know. I thought it would be good to cover it as a case.

christinach:
I haven't added support for more than one manifest urls. If this is the case for the ephemera resources reopen this ticket and resolve checkbox (2). 
@kelea99 tagging you to be aware.

PR #2902, opened 2025-08-26
Update code so that we map homoit controlled vocabulary to homoit subject and facet display

christinach:
Update the code base so that we map homoit controlled vocabulary to homoit subject and facet display
and loc controlled vocabulary to lc subject and facet display

related to #2898


sandbergja (approved):
Thanks, @christinach ! 🐬

kevinreiss (approved):

PR #2904, opened 2025-08-27
Bump change_the_subject version

sandbergja:
This gives us the performance improvement from
https://github.com/pulibrary/change_the_subject/pull/54


christinach (approved):
Thanks @sandbergja !

PR #2927, opened 2025-09-12
Index theses from dspace 7+

rladdusaw:


lib/bibdata_rs/src/theses/dataspace/document.rs
```diff
@@ -91,86 +90,84 @@ impl DataspaceDocument {
...
-            .with_mudd_walkin("yes")
+            .with_embargo_lift(metadatum_vec_from_string("2010-07-01"))
+            .with_mudd_walkin(metadatum_vec_from_string("yes"))
             .build();
 
-        assert_eq!(metadata.id.unwrap(), "123456");
-        assert_eq!(metadata.embargo_lift.unwrap(), vec!["2010-07-01"]);
-        assert_eq!(metadata.mudd_walkin.unwrap(), vec!["yes"]);
+        assert_eq!(metadata.id, Some("123456".to_string()));
+        assert_eq!(metadata.embargo_lift, Some(vec![Some("2010-07-01".to_string())]));        assert_eq!(metadata.embargo_lift, Some(vec![Some("2010-07-01".to_string())]));
```
sandbergja, line 110: We noticed that there are two assertions on the same line here

lib/bibdata_rs/src/theses/holdings.rs
```diff
@@ -81,19 +83,19 @@ pub fn on_site_only(
     location: bool,
     access_rights: bool,
     mudd_walkin: bool,
-    class_year: &[String],
+    class_year: Vec<String>,
```
sandbergja, line 86: Is there a way we can keep the old signature?  It would be more flexible to not require an owned `Vec`

sandbergja (approved):
Thanks, @rladdusaw ! 🐈‍⬛

PR #2930, opened 2025-09-16
Increase the solr_writer.thread_pool from the default 1

sandbergja:


marc_to_solr/lib/traject_config.rb
```diff
@@ -35,6 +35,10 @@
       raise exception
     end
   }
+  # These threads asynchronously POST to solr and wait for its response to confirm
+  # that everything is okay.  Since they spend most of their time waiting, it's
+  # fine to have more of these threads than we have CPU cores.
+  provide 'solr_writer.thread_pool', 5
```
christinach, line 41: 👍🏻

christinach (approved):
🍄 1

Issue #2924, opened 2025-09-10
Rescue a failed request to an ephemera resource so that the rake task does not break


christinach:
### What maintenance needs to be done?
When an ephemera resource does not have the expected structure then the async request breaks and the rake task stops.
We want to rescue this error, log the ephemera resource id and skip it so that the rake task does not fail.

### Level of urgency
- [x] High
- [ ] Moderate
- [ ] Low

### Why is this maintenance needed?
When an ephemera resource does not have the expected structure then the async request breaks and the rake task stops.
We want to rescue this error, log the ephemera resource id and skip it so that the rake task does not fail.

christinach:
Update the code to rescue for every struct that causes an error. 
The following [ephemera resource](https://figgy.princeton.edu/catalog/e35a19e5-cd0a-494d-a3bc-098f50926e52.jsonld) has  
```
origin_place: [
""
],
```
which breaks the rake task with the following error:
```
Error during indexing: Could not parse ephemera folder JSON at https://figgy.princeton.edu/catalog/e35a19e5-cd0a-494d-a3bc-098f50926e52.jsonld: reqwest::Error { kind: Decode, source: Error("invalid type: string \"\", expected struct OriginPlace", line: 1, column: 2324) }
```

christinach:
cover most of the case where we have bad structure from figgy . there are some more checks to add but it can happen at a later time.

PR #2932, opened 2025-09-17
Rescue cases where the json we receive from figgy is not the expected object

christinach:
handle invalid ephemera object for origin_place, subjects

related to [#2924]

### DIFF
--- a/lib/bibdata_rs/src/ephemera/ephemera_folder.rs
+++ b/lib/bibdata_rs/src/ephemera/ephemera_folder.rs
@@ -72,6 +72,8 @@ pub struct AuthorRoles {
     pub primary_author: String,
 }
 
+// VecSafe is a wrapper around Vec<T> that provides safe deserialization from JSON arrays
+// containing objects. It ignores any elements that fail to deserialize into T.
 #[derive(PartialEq, Debug, Clone)]
 pub struct VecSafe<T>(pub Vec<T>);
 
@@ -167,7 +169,6 @@ impl EphemeraFolder {
     pub fn origin_place_labels(&self) -> Vec<String> {
         match &self.origin {
             Some(origin_vector) => origin_vector
-                .0
                 .iter()
                 .filter(|origin| origin.exact_match.accepted_vocabulary())
                 .map(|origin| origin.label.clone())
@@ -590,7 +591,7 @@ mod tests {
             "Deserialization should not error on invalid origin_place type"
         );
         let folder = result.unwrap();
-        assert!(folder.origin.is_none() || folder.origin.as_ref().unwrap().0.is_empty());
+        assert!(folder.origin.is_none());
     }
 
     #[test]
@@ -610,6 +611,6 @@ mod tests {
             "Deserialization should not error on invalid subject type"
         );
         let folder = result.unwrap();
-        assert!(folder.subject.is_none() || folder.subject.as_ref().unwrap().0.is_empty());
+        assert!(folder.subject.is_none());
     }
 }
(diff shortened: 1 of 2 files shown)


christinach:
Deployed and indexed on catalog-qa
[ephemera reports](https://catalog-qa.princeton.edu/?f%5Bformat%5D%5B%5D=Report&page=1&search_field=all_fields)

lib/bibdata_rs/src/ephemera/ephemera_folder.rs
```diff
@@ -502,4 +572,44 @@ mod tests {
...
+        });
+
+        let result: Result<EphemeraFolder, _> = serde_json::from_value(invalid_json);
+
+        assert!(
+            result.is_ok(),
+            "Deserialization should not error on invalid origin_place type"
+        );
+        let folder = result.unwrap();
+        assert!(folder.origin.is_none() || folder.origin.as_ref().unwrap().0.is_empty());
```
sandbergja, line 593: This always is `None` for me -- would it be okay to simplify the assertion (and the one in the next test as well)?
```suggestion
        assert!(folder.origin.is_none());
```

lib/bibdata_rs/src/ephemera/ephemera_folder.rs
```diff
@@ -70,6 +72,73 @@ pub struct AuthorRoles {
     pub primary_author: String,
 }
 
+#[derive(PartialEq, Debug, Clone)]
+pub struct VecSafe<T>(pub Vec<T>);
```
sandbergja, line 76: It would be nice to have a short comment about what `VecSafe` is.

lib/bibdata_rs/src/ephemera/ephemera_folder_builder.rs
```diff
@@ -134,12 +136,12 @@ impl EphemeraFolderBuilder {
             format: self.format,
             id,
             language: self.language,
-            origin: self.origin_place,
+            origin: self.origin_place.map(|v| VecSafe::from(VecSafe(v))),
```
sandbergja, line 139: Clippy recommended for me:
```suggestion
            origin: self.origin_place.map(|v| VecSafe(v)),
```

lib/bibdata_rs/src/ephemera/ephemera_folder.rs
```diff
@@ -98,6 +167,7 @@ impl EphemeraFolder {
     pub fn origin_place_labels(&self) -> Vec<String> {
         match &self.origin {
             Some(origin_vector) => origin_vector
+                .0
```
sandbergja, line 170: I think we can remove `.0` here.  Since you implemented `Deref`, `origin_vector` has all the methods of `Vec` already available to it!  At least, that is what I understand from [the third bullet point here](https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion):

> If `T` implements `Deref<Target = U>`, [...] `T` implicitly implements all the methods of the type `U` which take the `&self` receiver.

lib/bibdata_rs/src/ephemera/ephemera_folder.rs
```diff
@@ -70,6 +72,73 @@ pub struct AuthorRoles {
     pub primary_author: String,
 }
 
+#[derive(PartialEq, Debug, Clone)]
+pub struct VecSafe<T>(pub Vec<T>);
+
+use std::ops::Deref;
+
+impl<T> Deref for VecSafe<T> {
```
sandbergja, line 80: Very cool!  The `Deref` trait seems very useful, thanks for introducing it!

lib/tasks/ephemera.rake
```diff
@@ -21,7 +21,8 @@ namespace :ephemera do
     puts "Successfully indexed #{documents.length} documents to #{solr_url}"
 
   rescue StandardError => e
-    puts "Error during indexing: #{e.message}"
+    Rails.logger.error("Error processing ephemera folder: #{e.message}")
```
sandbergja, line 24: Nice, I'm glad that we'll log these.

sandbergja (approved):
Nice, thanks @christinach ! 🐬

PR #2942, opened 2025-09-25
Rename any instances of catalog-alma-production to catalog-production

christinach:
Rename any instances of catalog-alma-production to catalog-production

related to https://github.com/pulibrary/pul_solr/issues/506


regineheberlein (approved):
thanks @christinach !

PR #2944, opened 2025-09-26
Update indexing documentation:

christinach:
- Numismatics
- Add section for ephemera


regineheberlein (approved):
thanks @christinach !

PR #2945, opened 2025-09-26
Update docs with solr8 references to solr9

christinach:
In a previous Pr we've updated the catalog collection with the new solr9 catalog collections.


regineheberlein (approved):
thanks @christinach !

PR #2960, opened 2025-10-06
Plasma Physics - delivery location Engineering

christinach:
Plasma Physics should have delivery location Engineering 
until we apply the branches pickup update


mzelesky (approved):
Looks like it follows the pattern for other changes.

PR #2968, opened 2025-10-21
Remove IL pickup from eastasian$ql.

christinach:
related to https://github.com/pulibrary/orangelight/issues/4410

### DIFF
--- a/config/locations/holding_locations.json
+++ b/config/locations/holding_locations.json
@@ -576,7 +576,7 @@
     },
     {
         "label": "",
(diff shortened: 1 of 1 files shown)


config/locations/holding_locations.json
```diff
@@ -895,7 +895,6 @@
             "PL",
```
kevinreiss: The five pickups for eastasian$ql should be

PF, QL, QP, QC, QT

kevinreiss (changes requested):

PR #2969, opened 2025-10-23
Remove unnecessary call from bibliographic holding availability endpoint

sandbergja:
This affects the endpoint /bibliographic/:bib_id/holdings/:holding_id/availability, which is used by the Requests form.

Prior to this commit, this endpoint made 2 or more requests to Alma:
* 1 to retrieve the MarcXML record along with requests and general availability data
* 1 or more to retrieve the item availability data

The first call was only used to retrieve the bib id -- something that we already have because the user provided it in the request. Therefore, we can eliminate this call and save an extra 500-1000ms each time we load the request form (and be gentler on our API limits).

Found while investigating performance issues related to https://github.com/pulibrary/orangelight/issues/3228

### DIFF
--- a/app/adapters/alma_adapter.rb
+++ b/app/adapters/alma_adapter.rb
@@ -83,18 +83,22 @@ class AlmaAdapter
     handle_alma_error(client_error: e)
   end
 
+  # Fetch the items for the holding and create
+  # the availability response for each item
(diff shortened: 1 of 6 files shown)


christinach (approved):
Thanks @sandbergja ! 🍄 🪈 🐬

PR #2991, opened 2025-12-24
Update health monitor rails gem

christinach:
The latest version uses lease_connection for Rails 8.1 For a Rails version before 8 it falls back to connection

helps with #2990

Please review and merge this one first. Then https://github.com/pulibrary/bibdata/pull/2992


sandbergja (approved):
Thanks, @christinach !

PR #2995, opened 2026-01-05
Invalid 008 should not cause indexing to crash

sandbergja:
See https://app.honeybadger.io/projects/54497/faults/117289412#notice-summary


christinach (approved):
Thank you @sandbergja @kevinreiss @VickieKarasic @christinach !

PR #3002, opened 2026-01-08
Rubocop updates and fixes

christinach:


regineheberlein (approved):
thanks @christinach !

PR #3004, opened 2026-01-12
Commit the db/schema.rb changes following the rails 8.1 upgrade

christinach:


regineheberlein (approved):
thanks @christinach !

PR #3007, opened 2026-01-13
Refactor: move Rust ruby method bindings to their own function and module

sandbergja:
This declutters our magnus init function a bit, and provides a module where we can manage Rust code related to the BibdataRs::Marc Ruby module.


christinach (approved):
Thanks @sandbergja ! 🍄

Issue #2815, opened 2025-07-24
[Results Page] Thesis Records say 2 Online Options when there is only 1.


kevinreiss:
### Expected behavior
On the search results page we display a count of the number of available ONLY when there is more than a single option for online attached to a record. 

### Actual behavior
For senior thesis imported from dataspace that have no access retrictions we currently say there are 2 online options for every record. [See this URL](https://catalog-qa.princeton.edu/catalog?per_page=20&search_field=all_fields&q=id%3Adsp*) for examples. 

### Steps to replicate
Visit 

### Impact of this bug
Inaccurate information on the search results page. 

### Implementation notes, if any
We want to move all the data into the electronic_portfolio_s field and no longer use the electronic_access_1display field for theses. 

The cause of this is the fact that every thesis record has both an electronic_access_1display and electronic_portfolio_s key. See this [record](https://catalog-qa.princeton.edu/catalog/dsp01br86b694j/raw) for an example. The electronic_portfolio_s keys for these records does not contain URL. We should add something to ignore this key data when making the online options count just for records of this type imported from dataspace.

kevinreiss:
We should consider an upstream fix in the indexing code for this data that doesn't introduce both a portfolio and an electronicaccess_1display key for these items.

kevinreiss:
I think if we just remove this line https://github.com/pulibrary/bibdata/blob/6b199db31e37c5682c1fc1d42488779b489c05a0/lib/bibdata_rs/src/solr/dataspace_solr_mapping.rs#L35 it will stop adding what seems extraneous data for the records we are loading from DSpace. Right all these have keys look like this added to each record we extract from either dataspace.
```
electronic_access_1display	'{"http://arks.princeton.edu/ark:/88435/dsp01ft848s955":["DataSpace","Full text"]}'
electronic_portfolio_s	
	'{"thesis":{"call_number":"AC102","call_number_browse":"AC102","dspace":true}}'
```

I cannot see how the second data point is useful for anything. 

Example record: https://catalog.princeton.edu/catalog/dsp01ft848s955/ 

Rather than just trying to add a workaround the code that produces the online count value on search results page, it seems cleaner to just take this out of the indexing process.

PR #3016, opened 2026-01-16
Stopping adding electoric_portfolio_s to solr documents for dataspace records

kevinreiss:
Will address https://github.com/pulibrary/bibdata/issues/2815 after a full re-index of the datspace records.


christinach (approved):
@kevinreiss I indexed locally one of the thesis. As you said it fixes the display of the two online options. 

### Search results
Thesis search results

### Record page

Record page

@kevinreiss I'm approving it,  can you remove the comment and commit again?

PR #3019, opened 2026-01-22
Update openssl gem 

kevinreiss:
Gets around ssl verification erros when working …with omniauth and localhost applications.


christinach (approved):
Thanks @kevinreiss 🚴🏻‍♂️

PR #3050, opened 2026-02-19
Include item "requested" attribute in the holdings API call

christinach:
Include item "requested" attribute in the holdings API call

bibliographic/bib_id/holdings/holding_id/availability.json API call

This API call is used in the catalog request form.

related to https://github.com/pulibrary/orangelight/issues/5632


regineheberlein (approved):
thanks @christinach !

PR #3101, opened 2026-03-10
Exclude rust and js-related directories from rubocop check

sandbergja:
On main, running `bundle exec rubocop` takes 33 seconds for me. With this branch, it takes 3 seconds.


christinach (approved):
Thanks @sandbergja !

PR #3102, opened 2026-03-10
Optimize genres function by removing regular expressions

sandbergja:
Rather than use regular expressions, we can divide each term into windows, and see if the windows of one term contain the windows of the second term.

From `cargo bench`:

```
genres                  time:   [19.782 µs 19.858 µs 19.946 µs]
                        change: [−11.756% −11.240% −10.738%] (p = 0.00 < 0.05)
                        Performance has improved.
```


christinach (approved):
Thanks @sandbergja 🍄 ! inspiring 🪟

PR #3202, opened 2026-04-15
Upgrade marctk to 0.7.0

sandbergja:
It has not yet been released, but has some performance improvements. It also has a new repo: https://github.com/kcls/marctk

So this helps to test that the new repo is behaving as expected as well (see https://github.com/kcls/evergreen-universe-rs/issues/32)


christinach (approved):
Thanks @sandbergja 🍄 !

PR #3203, opened 2026-04-15
Remove unneeded use<> lifetime captures in impl Trait bounds

sandbergja:
These are no longer needed in Rust 2024

See https://doc.rust-lang.org/nightly/edition-guide/rust-2024/rpit-lifetime-capture.html


rladdusaw (approved):
Thanks @sandbergja!

christinach (approved):
nice Thanks @sandbergja 🍄

PR #3204, opened 2026-04-15
Add jsonld_genere_field with all 655 fields

rladdusaw:
Supports pulibrary/orangelight#5768

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -979,7 +979,7 @@ to_field 'rbgenr_genre_facet' do |record, accumulator|
 end
 
 # used for the JSONLD output. Uses all the 655 fields to get all genre data
-to_field 'jsonld_genre_facet', extract_marc('655|*7|avxyz')
+to_field 'jsonld_genre_display', extract_marc('655|*7|avxyz')
 
 to_field 'cjk_subject' do |_record, accumulator, context|
   accumulator.replace(context.clipboard[:solr_fields]['cjk_subject'])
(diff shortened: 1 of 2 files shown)


sandbergja (commented):
Thanks, @rladdusaw !  🐈‍⬛   It might be nice to include a test as well (especially so we can be sure to keep the same behavior as we migrate fields to Rust).  Also, just to check: would the existing `genre_facet` field meet your needs?  If so, it would be nice to not have to double-index those values.

rladdusaw:
> Thanks, @rladdusaw ! 🐈‍⬛ It might be nice to include a test as well (especially so we can be sure to keep the same behavior as we migrate fields to Rust). Also, just to check: would the existing `genre_facet` field meet your needs? If so, it would be nice to not have to double-index those values.

The `genre_facet` field doesn't appear in the solr document for the example records Trey mentioned in the issue.

sandbergja:
> The `genre_facet` field doesn't appear in the solr document for the example records Trey mentioned in the issue.

I think that might just be because it isn't stored?

rladdusaw:
> I think that might just be because it isn't stored?

It doesn't even appear to be in the solr config and I don't really understand what it's doing to create the field

```
# 600/610/650/651 $v, $x filtered
# 655 $a, $v, $x filtered
to_field 'genre_facet' do |_record, accumulator, context|
  accumulator.replace(context.clipboard[:solr_fields]['genre_facet'])
end
```

but the comment seems to imply it's bringing in more than the 655s

sandbergja (approved):
Thanks, @rladdusaw !

PR #3206, opened 2026-04-15
Stop using MarcBreaker for passing Marc records from Ruby to Rust

sandbergja:
Instead, use magnus to get the values directly from Ruby Record and DataField objects.

This is slightly faster, and certainly simpler.

Before:
```
2026-04-15T10:43:18-07:00  INFO finished Traject::Indexer#process: 45288 records in 99.090 seconds; 457.0 records/second overall.
```

After:
```
2026-04-15T11:19:16-07:00  INFO finished Traject::Indexer#process: 45288 records in 95.555 seconds; 473.9 records/second overall.
```


lib/bibdata_rs/src/marc.rs
```diff
@@ -131,22 +114,6 @@ pub fn build_call_number(
     Ok(call_number.filter(|s| !s.is_empty()))
 }
 
-pub fn private_items(
```
sandbergja, line 134: This function is not actually used

christinach (approved):
@sandbergja 🍄 This is great and inspiring ! Thank you for taking the time to read this PR with me.

PR #3209, opened 2026-04-15
Use bundle install, rather than circleci ruby/install-deps

sandbergja:
The ruby/install-deps keeps failing with the message 'Killed'


christinach (approved):
🍄

sandbergja:
Never mind, the same thing happened to this PR 🫨

PR #3210, opened 2026-04-15
Be more selective about which gems we install for the rust tests in CI

sandbergja:
Hopefully this keeps us from getting the 'Killed' signal on bundle install...


sandbergja:
This also didn't work

PR #3212, opened 2026-04-15
Use official ruby docker image instead of circleci's

sandbergja:
This seems to resolve the 'killed' signal that we were getting when running `bundle install`


christinach (approved):
Thanks @sandbergja !

Issue #3190, opened 2026-04-09
Temporary Location Should be reflected in the Library and Holdings Location Facet Filters


kevinreiss:
### User story
As a user I would like to see the temporary location of item used when the Library facet filter is applied. 

### Acceptance criteria

- [ ] Items that are currently in a temporary location should have that location included in the Library Facet. 
- [ ] Items that are currently in a temporary location should also be filtered by holding location relfecting their temporary location. 

### Concrete example
See [MMS ID](https://catalog.princeton.edu/catalog?search_field=all_fields&q=9929638013506421) for a permanent Firestone item temporarily placed 

This item should be under the "Commons Library" library facet rather than it's current facet of "Firestone Library". This items should be under the "[commons$stacks](https://catalog.princeton.edu/catalog?advanced_type=advanced&clause%5B0%5D%5Bfield%5D=all_fields&clause%5B0%5D%5Bquery%5D=&boolean_operator1=AND&clause%5B1%5D%5Bfield%5D=author&clause%5B1%5D%5Bquery%5D=&boolean_operator2=AND&clause%5B2%5D%5Bfield%5D=title&clause%5B2%5D%5Bquery%5D=&f_inclusive%5Badvanced_location_s%5D%5B%5D=commons%24stacks&range%5Bpub_date_start_sort%5D%5Bbegin%5D=&range%5Bpub_date_start_sort%5D%5Bend%5D=&commit=Search)" holding location, currently they are still under "firestone$stacks". There are 1568 current items in this state for the commons library. 

## Notes
There is some urgency to get this done soon because the Commons Library will soon start circulating items.

christinach:
resolved by https://github.com/pulibrary/bibdata/pull/3218

PR #3218, opened 2026-04-21
When a temporary location exists index it in the location field

christinach:
related to #3190

The library facet will now display the temporary location if a holding is in a temp location

example with holding in temporary location Commons:

Screenshot 2026-04-21 at 5 56 06 PM


sandbergja (approved):
Thanks, @christinach !  🐬 🍄

PR #3221, opened 2026-04-22
Replace unwrap() with expect to provide better error messages.

christinach:


sandbergja (approved):
Thanks, @christinach !  🐬 🍄

PR #3222, opened 2026-04-22
[Indexing doc] Update ephemera

christinach:
Update ephemera section to SET the solr url in the rake task


kevinreiss (approved):
Thanks @christinach.

PR #3224, opened 2026-04-22
Fix author_roles serialization in Solr documents

sandbergja:
The author_roles_1display field should be a string with escaped JSON inside, but instead was incorrectly just regular JSON with no escaping.

As a result, we could not index ephemera because solr rejected it with an unintuitive error:

```
Exception
org.apache.solr.common.SolrException: Can't use SignatureUpdateProcessor with partial update request containing signature field: id
```

This commit fixes the serialization to be an escaped string again.


christinach (approved):
Thanks @sandbergja @christinach 🍄 🐬

PR #3225, opened 2026-04-22
Remove CompositeCacheMap class

sandbergja:
When it was first introduced, it provided a compatibility layer for searching the Figgy and PLUM caches at the same time.  See https://github.com/pulibrary/bibdata/pull/336

We no longer have PLUM, so this is essentially doing the same thing as the regular ark cache, and we can simplify things by using the basic Figgy ark cache everywhere.


kevinreiss (approved):
Thanks @sandbergja.

Issue #3230, opened 2026-04-23
Some "Completed" Figgy Ephemera Objects Do Not Show up in the Catalog


kevinreiss:
### Expected behavior
All ["Completed"](https://figgy.princeton.edu/catalog?f%5Bephemera_project_ssim%5D%5B%5D=Born+Digital+Monographic+Reports+and+Papers&f%5Bstate_ssim%5D%5B%5D=complete&q=) Figgy Born Digital Ephemera Objects should appear in the catalog. 

### Actual behavior
Some objects marked completed appear in the Catalog some do not. 

### Steps to replicate
Visit any of these completed item IDs in the Catalog and they DO NOT current display:

1. fe256c85-da9d-43c7-8c13-f29668f187fd
2. fb49c0c8-8eeb-4e32-afbe-04d8d14edacf
3. fa30780e-dfd8-4545-b1b0-b3eec9fca96b
4. 0379f1af-4b26-4373-a6bd-ea53f754bfa3

Visit any of these completed item IDs in the Catalog and they DO display:

1. [fef49497-386b-49ec-a6ad-3be4732c8674](https://catalog.princeton.edu/catalog/fef49497-386b-49ec-a6ad-3be4732c8674)
2. [015fdd2f-a55f-4025-8fe3-d3bfbc4bf2fe](https://catalog.princeton.edu/catalog/015fdd2f-a55f-4025-8fe3-d3bfbc4bf2fe)

### Impact of this bug
Items that should be accessible aren't accessible. 

### Acceptance criteria
- [x] Understand why certain objects aren't indexing
- [ ] Index all the available/completed ephemera documents.

christinach:
The main problem in our code is that we parse data from a hardcoded url of per_page=100. We never check for more pages. So we are indexing 100 documents. This is why you can't see the rest of the documents.  
Currently working on parsing the total pages of the response and provide a fix as soon as possible.

christinach:
resolved by https://github.com/pulibrary/bibdata/pull/3239

PR #3239, opened 2026-04-27
Support paging of figgy born ephemera - εφήμερα objects

christinach:
related to [#3230]

### DIFF
--- a/lib/bibdata_rs/src/ephemera/born_digital_collection.rs
+++ b/lib/bibdata_rs/src/ephemera/born_digital_collection.rs
@@ -1,11 +1,27 @@
-use crate::solr::SolrDocument;
-
 use super::CatalogClient;
+use crate::solr::SolrDocument;
 use serde::Deserialize;
 
 #[derive(Deserialize, Debug)]
 pub struct FoldersResponse {
     pub data: Vec<BornDigitalCollection>,
+    pub meta: Option<Meta>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Meta {
+    pub pages: Pages,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Pages {
+    // pub current_page: Option<u32>,
+    // pub next_page: Option<u32>,
+    // pub prev_page: Option<u32>,
+    pub total_pages: u32,
+    // pub limit_value: Option<u32>,
+    // pub offset_value: Option<u32>,
+    // pub total_count: Option<u32>,
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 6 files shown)


christinach:
Indexed ephemera on  [catalog-staging figgy ephemera](https://catalog-staging.princeton.edu/catalog?search_field=all_fields&q=) count: 390
in figgy I'm seeing figgy [ephemera project complete state](https://figgy.princeton.edu/catalog?f%5Bephemera_project_ssim%5D%5B%5D=Born+Digital+Monographic+Reports+and+Papers&f%5Bstate_ssim%5D%5B%5D=complete&q=) count: 391. Currenlty I'm not sure which object is not being indexed. I can review this in another Pr. 
I would like with this Pr to support the paging and index ephemera in the catalog.

lib/bibdata_rs/src/ephemera/born_digital_collection.rs
```diff
@@ -1,11 +1,27 @@
...
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Meta {
+    pub pages: Pages,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Pages {
+    // pub current_page: Option<u32>,
```
sandbergja: Do you think we could remove these commented-out fields?

lib/bibdata_rs/src/ephemera/born_digital_collection.rs
```diff
@@ -25,10 +41,26 @@ pub async fn read_ephemera_folders(
...
+    let mut page_number = 1;
+    // call the first page page=1 to get the responce meta info, total_pages etc.
+    let response = client.get_folder_data(page_number).await?;
+    let total_pages = response
+        .meta
+        .as_ref()
+        .map(|m| m.pages.total_pages)
+        .unwrap_or(1);
+
+    all_ids.extend(response.data.into_iter().map(|item| item.id));
```
sandbergja: Since `response.data.into_iter().map(|item| item.id))` is repeated multiple times, do you think it would be worth it to pull it out as a method in an `impl` block for the `FoldersResponse` struct?

lib/bibdata_rs/src/ephemera/born_digital_collection.rs
```diff
@@ -25,10 +41,26 @@ pub async fn read_ephemera_folders(
...
+
+    while page_number < total_pages {
+        page_number += 1;
+
+        let response = client.get_folder_data(page_number).await?;
+        all_ids.extend(response.data.into_iter().map(|item| item.id));
```
sandbergja: 👍🏻

sandbergja (commented):
Nice, thanks @christinach !  Two questions/suggestions -- neither one is a blocker

lib/bibdata_rs/src/ephemera/born_digital_collection.rs
christinach: oh yes I'm sorry !

christinach:
@sandbergja thanks for reviewing can you please check again?

lib/bibdata_rs/src/ephemera/born_digital_collection.rs
```diff
@@ -1,11 +1,32 @@
...
+
+impl FoldersResponse {
+    pub fn response_ids(
+        response: FoldersResponse,
+    ) -> std::iter::Map<
+        std::vec::IntoIter<BornDigitalCollection>,
+        impl FnMut(BornDigitalCollection) -> String,
+    > {
+        response.data.into_iter().map(|item| item.id)
+    }
```
sandbergja, line 19: Two ideas to simplify the signature a bit:

- If we use `&self`, we can replace `FoldersResponse::response_ids(response)` with `response.response_ids` below
- We don't necessarily have to be as specific about the return type

```suggestion
    pub fn response_ids(
        &self
    ) -> impl Iterator<Item=String>
    {
        self.data.into_iter().map(|item| item.id.clone())
    }
```

sandbergja (approved):
Thanks, @christinach !  Looks great!  🐬 

I had some (non-blocking) suggestions about the signature of the new method, otherwise looks great!

lib/bibdata_rs/src/ephemera/born_digital_collection.rs
christinach: @sandbergja thanks I will open a new Pr with your suggestion

PR #3243, opened 2026-04-29
Test both location and advanced_location_s when we have a holding in a temporary location

christinach:
Test both location and advanced_location_s when we have a holding in a temporary location

related to [#3190]


rladdusaw (approved):
Thanks @christinach!

PR #3246, opened 2026-04-30
Remove code that is never called

sandbergja:
When it was first checked in (in 83109f1da2be2c892886693669b209307360da5d), the holding_library translation map included values like "Marquand Library" and "Special Collections".

However, today, it only includes a single value, "ReCAP". Therefore, this condition is never true and the code is never called.


christinach (approved):
Thanks @sandbergja 🍄 🐬

PR #3259, opened 2026-05-05
Add arm64-darwin to Gemfile.lock

sandbergja:
This allows me to download the precompiled nokogiri with `bundle install` so I don't get a compilation error locally


christinach (approved):
Thanks @sandbergja !

PR #3262, opened 2026-05-06
Remove all index-time responsibilities from old AugmentTheSubject class

### LINKED ISSUE #2871: Rewrite AugmentTheSubject behavior in Rust

Rewrite AugmentTheSubject behavior in Rust

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

This is one of the slower processes in the indexing process.  It is not as complex as some other parts of the indexing code, making it an impactful piece of the code to work with.

Overall, we'd like to move more of our indexing to Rust so that we can keep our indexing process performant even as we add more partner records and other functionality.

### Acceptance criteria

 - [ ] AugmentTheSubject is implemented in Rust
 - [ ] We can demonstrate that the new implementation is significantly faster through a micro-benchmark or a profile of indexing production-like data

sandbergja:
Rust is now responsible for all index-time work.  AugmentTheSubject is now only responsible for parsing the CSV files to prepare data for the rake task.

Closes #2871

### DIFF
--- a/marc_to_solr/lib/augment_the_subject.rb
+++ b/marc_to_solr/lib/augment_the_subject.rb
@@ -1,18 +1,10 @@
 # frozen_string_literal: true
 
 ##
-# The creation and management of metadata are not neutral activities.
+# This class is responsible for preparing data from a CSV file into hashes that can be
+# used to determine whether or not subjects are relevant to the field of Indigenous Studies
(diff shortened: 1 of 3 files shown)


christinach (approved):
Thanks @sandbergja 🍄 !

PR #3268, opened 2026-05-08
Use Rust to calculate location names  for advanced_location_s field

christinach:
Use Rust to calculate location names to assign to advanced_location_s field
and remove translation maps for this case

related to #3176

The current advanced_location_s would return `expect(record['advanced_location_s']).to eq ['lewis$stacks', 'Lewis Library', 'firestone$stacks', 'Firestone Library']` 

I tested it in orangelight by reindexing the record and there is no impact.

### DIFF
--- a/spec/marc_to_solr/lib/config_spec.rb
+++ b/spec/marc_to_solr/lib/config_spec.rb
@@ -156,6 +156,7 @@ describe 'From traject_config.rb', :indexing do
       it 'indexes the location_code_s' do
         record = @indexer.map_record(fixture_record('9992320213506421'))
         expect(record['location_code_s']).to eq ['lewis$stacks', 'firestone$stacks']
+        expect(record['advanced_location_s']).to eq ['lewis$stacks', 'firestone$stacks', 'Lewis Library', 'Firestone Library']
       end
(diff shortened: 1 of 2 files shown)


spec/marc_to_solr/lib/config_spec.rb
```diff
@@ -156,6 +156,7 @@ def fixture_record(fixture_name, indexer: @indexer)
       it 'indexes the location_code_s' do
         record = @indexer.map_record(fixture_record('9992320213506421'))
         expect(record['location_code_s']).to eq ['lewis$stacks', 'firestone$stacks']
+        expect(record['advanced_location_s']).to eq ['lewis$stacks', 'firestone$stacks', 'Lewis Library', 'Firestone Library']
```
christinach, line 159: The current advanced_location_s would return `expect(record['advanced_location_s']).to eq ['lewis$stacks', 'Lewis Library', 'firestone$stacks', 'Firestone Library']` 

I tested it locally in orangelight reindexing the record and there is no impact.

sandbergja (approved):
Nice, thanks @christinach 🐬

PR #3284, opened 2026-05-13
Signoz two electric

kayiwa:
add observability


christinach (approved):
Thank you @kayiwa and @sandbergja ! 🦒 🍄

Issue #3227, opened 2026-04-22
Replace ark cache with the figgy mms report data in redis


sandbergja:
### What maintenance needs to be done?

Replace ark cache with the figgy mms report data in redis

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?

* The ark cache is currently very out of date, so a lot of data is missing from it.  It uses Voyager IDs instead of Alma ids, so our code causes some incorrect urls (see #1720). We cannot update our current caches since Figgy has changed since the last time the ark cache worked.
* The ark cache is one of the bottlenecks of our current indexing code, and we'd like to improve performance
* The ark cache relies on a very old version of the lightly gem

### Acceptance criteria

 - [x] We no longer use the lightly gem
 - [x] We remove the rake tasks in the `liberate:arks` namespace
 - [x] The logic for the OrangelightUrlBuilder is re-implemented to use the cached figgy data from redis
 - [x] The logic for the IIIFManifestUrlBuilder is re-implemented to use the cached figgy data from redis
 - [x] This does not introduce a performance regression when indexing alma records with arks in 856$u

PR #3286, opened 2026-05-13
Replace IIIFManifestUrlBuilder with the Cached MMS IDs report

sandbergja:
Helps with #3227


christinach (approved):
Thanks @sandbergja ! 🍄

PR #3288, opened 2026-05-13
Remove ark cache

### LINKED ISSUE #3227: Replace ark cache with the figgy mms report data in redis

sandbergja:
Closes #3227

We do get a small speedup (12 records/second faster):

* Before: `45288 records in 94.395 seconds; 479.8 records/second overall.`
* After: `45288 records in 92.015 seconds; 492.2 records/second overall.`


.circleci/config.yml
```diff
@@ -84,7 +84,6 @@ jobs:
           command: bundle exec rspec spec
           environment:
             RAILS_ENV: test
-            FIGGY_ARK_CACHE_PATH: spec/fixtures/marc_to_solr/figgy_ark_cache
```
christinach, line 87: 🎉

lib/tasks/orangeindex.rake
```diff
@@ -1,14 +1,9 @@
 require 'faraday'
 require 'json'
-require 'lightly'
```
christinach, line 3: 🎉

christinach (approved):
Thanks @sandbergja !

christinach:
not a blocker for this pr but we want also to remove the FIGGY_ARK_CACHE env in ansible

PR #3289, opened 2026-05-13
Remove dependency on ActiveSupport #try extension from AccessFacetBuilder

sandbergja:
AccessFacetBuilder uses the Rails `#try` method, which is required by the ark cache classes.

This use of `#try` can be replaced with Ruby's built-in safe navigation, so that we no longer depend on the ark cache classes to provide the Rails `#try` method.


christinach (approved):
Thanks @sandbergja !

PR #3290, opened 2026-05-13
Add vernier gem

sandbergja:
This makes it easier to profile locally in one's dev environment and allows us to use Sidekiq's new profiling feature (see https://github.com/sidekiq/sidekiq/wiki/Profiling)


christinach (approved):
Thanks @sandbergja !

PR #3294, opened 2026-05-14
Remove delivery locations _form

christinach:
We don't edit the delivery locations.
We used to do this years ago and when the
locations were still in a gem


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #3295, opened 2026-05-14
Remove _form to build holding_locations

christinach:
This was a feature back in the days when locations were a gem. We don't use a form to edit the locations


sandbergja (approved):
Thanks, @christinach ! 🐬

Issue #3287, opened 2026-05-13
Use json file holding_locations.json and delivery_locations.json instead of the database


christinach:
## Current state
Currently we're using the json files to update the database and then we're using the controllers to serve the json from the database

## Desired State
Simplify the process and use json file holding_locations.json and delivery_locations.json instead of the database. Update the controllers to serve the json with no need to go through the database.

PR #3297, opened 2026-05-14
Organize locations - Create new module locations.rs

christinach:
Move code related to locations from holding_locations.rs

related to #3287


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #3339, opened 2026-06-09
Update profiling docs

sandbergja:
* Use vernier
* Update guidance on when to use profiling
* Add info about the figgy redis cache


kevinreiss (approved):
Thanks @sandbergja

PR #3344, opened 2026-06-10
Rewrite title_no_h_index in Rust

sandbergja:
Also, DRY up non-filing character calculation.

Partially written by pi + local qwen

### DIFF
--- a/lib/bibdata_rs/src/marc/ruby_bindings.rs
+++ b/lib/bibdata_rs/src/marc/ruby_bindings.rs
@@ -104,7 +104,7 @@ fn solr_fields(ruby: &Ruby, record: magnus::RObject) -> Result<RHash, magnus::Er
         .ok()
         .and_then(|date| date.maybe_to_string());
 
-    let hash = ruby.hash_new_capa(40);
+    let hash = ruby.hash_new_capa(41);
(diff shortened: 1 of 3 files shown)


kevinreiss (approved):
Just curious why the one-digit increment?

regineheberlein (approved):
Thanks @sandbergja !

sandbergja:
> Just curious why the one-digit increment?

`ruby.hash_new_capa()` creates a new Ruby hash with enough memory allocated to accommodate the specified number of entries.  Since we are adding a new entry to the hash later in the function (with `hash.aset`), we increase the initial capacity so it is big enough to accommodate all 41 entries without needing to wait for Ruby to allocate more memory partway through.  This is the same as calling `Hash.new(capacity: 41)` [in Ruby proper](https://docs.ruby-lang.org/en/4.0/Hash.html#method-c-new).

I have not actually benchmarked this.  Maybe it would be more pleasant to not need to remember to increment this (we could do `ruby.hash_new_capa(300)` or something so we don't have to adjust it each time, or not specify a capacity and just let Ruby automatically allocate more memory when it is needed).  I'd need to benchmark it to actually know if specifying the capacity in advance actually helps with performance in any meaningful way...

PR #3346, opened 2026-06-11
Increase traject solr writer threads

sandbergja:
This increases performance a lot for me locally:

Before:

```
INFO finished Traject::Indexer#process: 100000 records in 171.518 seconds; 583.0 records/second overall.
```

After:
```
INFO finished Traject::Indexer#process: 100000 records in 133.928 seconds; 746.7 records/second overall.
```


christinach (approved):
🍄 thanks

PR #3347, opened 2026-06-12
Migrate other_id field to Rust

sandbergja:
Pi and qwen 3.6 helped with this commit


christinach (approved):
Thanks @sandbergja ! 🍄

PR #3348, opened 2026-06-13
Fix regression in calculating holding_location

sandbergja:
Prior to 6c3c88df7ec41bfe1c01e55f0bd14dc4f4d52b3c, we did not index partial location codes when 852 was not present.  This commit restores this behavior so that records don't get assigned location codes like $pa
the advanced search form location dropdown.  It has All Princeton Holdings, $pa, Architecture Library, and Commons Library


kevinreiss (approved):

kevinreiss:
Thanks @sandbergja

PR #3357, opened 2026-06-15
Migrate oclc_s field to Rust

sandbergja:
Also, remove OCLC-related ruby bindings that are no longer needed


kevinreiss (approved):
Thanks @sandbergja

PR #3358, opened 2026-06-15
Handle location codes Harvard records that have 852$8

sandbergja:
Harvard puts the Alma ID into the 852$8 of some records (which we should ignore).  Before this commit, we consulted the 852$8 and confirmed that it was a Princeton Alma holding, and then fell back to not including any location code.  We should instead disregard 852$8s in partner records and focus on the 852$0.


christinach (approved):
Thanks 🍄 🪈 🐬 !

PR #3359, opened 2026-06-16
Replace library-standard-numbers gem with calling our Rust crate directly

sandbergja:
We already use the crate that this gem wraps in the project, so we can call the crate directly and slightly simplify installation (one less gem with native extensions that needs to be installed).


marc_to_solr/lib/princeton_marc.rb
```diff
@@ -1,5 +1,4 @@
 require 'active_support'
-require 'library_standard_numbers'
```
christinach, line 2: 🎉

Gemfile
```diff
@@ -27,7 +27,6 @@ gem 'jbuilder'
 gem 'jquery-rails'
 gem 'jquery-tablesorter', '~> 1.28'
 gem 'lcsort'
-gem 'library_standard_numbers'
```
christinach, line 30: 👍🏻

christinach (approved):
Thanks @sandbergja ! 🍄

PR #3363, opened 2026-06-17
Migrate scale_display field to Rust

sandbergja:
This commit introduces a macro extract_marc! that works similarly to the extract_marc method in traject.

In traject, you might write:

```
extract_marc('255b:342a')
```

With this macro, you can achieve a similar effect with:

```
extract_marc!("255b", "342a")(record)
```


christinach (approved):
Thanks @sandbergja !

PR #3364, opened 2026-06-17
Move several simple fields to Rust using extract_marc macro

sandbergja:
Before:

```
INFO finished Traject::Indexer#process: 100000 records in 131.488 seconds; 760.5 records/second overall.
```

After:

```
INFO finished Traject::Indexer#process: 100000 records in 125.937 seconds; 794.0 records/second overall.
```


christinach (approved):
nice @sandbergja ! thanks 🍄

PR #3365, opened 2026-06-17
Migrate more simple solr fields to use the extract_marc! macro

sandbergja:
The change in performance is negligible.  My theory is that most of these fields are quite uncommon.


christinach (approved):
Go Rust! Thanks @sandbergja 🍄 !

PR #3367, opened 2026-06-18
Make AGENTS.md more specific

sandbergja:
These changes have caused my local pi + qwen 3.5 setup use less context when doing small Ruby -> Rust rewrites.


christinach (approved):
Thanks @sandbergja !

PR #3368, opened 2026-06-18
Simplify: return boolean values instead of closures

sandbergja:


christinach (approved):
thanks @sandbergja 🍄 !

PR #3369, opened 2026-06-18
Make ExtractSpec more flexible regarding scripts to index

sandbergja:
This is an incremental step towards allowing the extract_marc! macro to be able to index only Latin or only non-Latin scripts


regineheberlein (approved):
thanks @sandbergja !

PR #3370, opened 2026-06-18
Re-implement title_display in Rust

sandbergja:
As part of this work, make the extract_marc! macro more flexible so that it can index Latin and Non-Latin scripts (default), only Latin, or only Non-Latin.


regineheberlein (approved):
thanks @sandbergja !

PR #3371, opened 2026-06-18
Disregard locations from Alma 852 fields with $0

sandbergja:
Rather than just consulting the 852 field to determine the location, we also check the 001 to check if it is an Alma or Partner record.

This excludes a few unusual values that we had in Alma 852 fields, for example (as pictured below), locations named `thx`, `ues`, and `هگ`

the advanced search location dropdown, showing some unusual location names


christinach (approved):
Thanks @sandbergja 🍄 !

PR #3393, opened 2026-06-30
Correct the name of solr field

sandbergja:
This led the `cataloged_tdt` field to not get indexed at all in the last full re-index, causing the Recently Added facet not to display


christinach (approved):
Thanks @sandbergja !

PR #3394, opened 2026-06-30
Increase dependabot cooldown to 7 days

sandbergja:
To comply with this [new rule from semgrep](https://github.com/semgrep/semgrep-rules/pull/3805)


christinach (approved):
thanks @sandbergja !

PR #3415, opened 2026-07-14
Add another test case to confirm cataloged_tdt date extraction/sorting

sandbergja:


christinach (approved):
thanks @sandbergja 🍄 !

PR #3416, opened 2026-07-14
Address new clippy warnings

sandbergja:


christinach (approved):
Thanks @sandbergja 🍄 
📎

PR #3427, opened 2026-07-20
Fix original language display regression

sandbergja:
The solr field `original_language_display` is only intended to display 880 fields that do not match any non-880 fields.  However, I introduced a regression in #3365, and it started to display all 880 fields, regardless of whether or not they have a pair.


christinach (commented):
Thank you @sandbergja 🍄 🐬 !

christinach (approved):
Thank you @sandbergja 🍄 🐬 !

PR #3442, opened 2026-07-29
Remove PF from annex$stacks

christinach:
Annex should not have delivery location PF

related to https://github.com/pulibrary/orangelight/issues/6056


rladdusaw (approved):
Thanks @christinach!

PR #3456, opened 2026-08-04
Remove step from setup instructions

sandbergja:
I was able to setup locally without this step -- perhaps we no longer need it?


christinach (approved):
thanks @sandbergja !

Issue #3404, opened 2026-07-07
Investigate using docValues for sort fields


sandbergja:
### What maintenance needs to be done?

Investigate whether we can reduce how much field cache the catalog index requires by switching our `*_sort` fields to docValues.

To check the field cache stats, from @tpendragon :
1. From the solr admin UI, select the core
2. Go to plugins and metrics
3. Check the field cache stats -- search for "MB".  Many are sort fields.

To use docValues for these fields, also from @tpendragon !

> [Use] a solr.TextField for _sort you used a string with a docValues=true  , but you'd have to send that sort field in at document writing time rather than using the analyzers in alphaNumSort.
> Basically change https://github.com/pulibrary/orangelight/blob/main/solr/conf/schema.xml#L386 to be `<fieldType name="alphaNumSort" class="solr.StrField" sortMissingLast="true" docValues="true" />` then get your tests passing by making bibdata strip the whitespace, icu fold, and replace non-alphanumeric characters.

### Level of urgency
- [ ] High
- [x] Moderate
- [ ] Low

### Why is this maintenance needed?
It could improve the performance and reliability of our solr!

### Acceptance criteria

 - [ ] We try using docValues for these fields
 - [ ] We check to see if it requires less field cache
 - [ ] We run a realistic load test on staging
 - [ ] If everything looks good, we reindex and put this into prod.

PR #3458, opened 2026-08-05
Remove all punctuation from title_sort

christinach:
Use new method remove_all_punctuation in string_normalize.rs on title_sort

In our current schema.xml the title_sort field removes punctuation as part of Analyses this work saves time for solr and then when we migrate to use title_sort_key it will ensure that sorting does not take punctuation into account

related to #3404


sandbergja (approved):
Thanks, @christinach ! 🐬 🍄 🐄

PR #3461, opened 2026-08-05
Index title_sort_key directly, rather than title_sort

sandbergja:
See https://github.com/pulibrary/pul_solr/pull/564

Currently, there is a copyField that copies title_sort into title_sort_key.  However, once https://github.com/pulibrary/orangelight/pull/6079 is merged and deployed, we will use title_sort_key directly, and can safely stop indexing into title_sort.


christinach (approved):
thanks @sandbergja !

PR #3462, opened 2026-08-06
Normalize author_sort field before sending it to solr

sandbergja:
We are migrating from an author_sort TextField in solr to an author_sort_key ICUCollationField.  The author_sort analysis removes all numbers and puncutation from the author name.  To acheive a similar effect, we normalize the value before sending it to solr, so that it works with both the author_sort and the new author_sort_key (which is being populated via a copyField).


christinach (approved):
Thanks @sandbergja 🍄

PR #3463, opened 2026-08-06
User author_sort_key instead of author_sort field

sandbergja:
We need to merge and deploy https://github.com/pulibrary/orangelight/pull/6086 first


christinach (approved):
thanks 🍄

PR #3481, opened 2026-08-18
Add necessary packages to devbox

sandbergja:
With this, I was about to run `cargo test` successfully!

Also, remove node and yarn, since we don't use them in this project.


rladdusaw (approved):
Thanks @sandbergja!

PR #3483, opened 2026-08-20
WIP add display_format to holdings fields

### LINKED ISSUE #3480: Adding "display_format" attribute to Physical and Electronic Holding Objects
### User story
When multiple matching records are combined in a cluster the holdings data needs to retain the original record ID for use when supporting request and availability features. 

### Acceptance criteria

- [ ] Each physical holding object in the "holdings_1display" key has a new attribute "display_format" that stores the display format calculated for the original record. 
- [ ] Each electronic object in the "electronic_portfolio_s" key has a new attribute "display_format" that stores the display format calculated for the original record.  
- [ ] Each electronic object in the "figgy_1display" key has a new attribute "display_format" that stores the display format calculated for the original record. 
- [ ] The display format should "EBook" when the display_format is being calculated for an "electronic_portfolio_s" holding that would otherwise have the format of "Book". 
- [ ] The display format should "EBook" when the display_format is being calculated for an "figgy_1display" holding that would otherwise have the format of "Book". 

### Concrete example
"[This is Happiness](https://catalog.princeton.edu/catalog?search_field=title&q=This+is+happiness)" is a record that has two electronic and two physical items. The display formats for the four records should be two "Book" records, one "EBook" record and one "Audio" format record. 

The algorithm to translate the current "Format" attribute to "display_format" is documented in this [spreadsheet](https://docs.google.com/spreadsheets/d/1eTF2E0Ay8kLlzUBw-Y1p6Sa6OYIcveyWa8aeBMfn404/edit?gid=0#gid=0). We will need to add a common method that can be used by all three attributes we want to add this to. Additional logic will need to "pick" a preferred format in instances when two formats are present for a record. The rules for preference when multiple formats are available are as follows:

Prefer the more specific format in roughly this order for resources that wouldn't be determined to be an "EBook" for figgy_1display or electronic_portfolio_s objects:

1. Microform, Manuscript, Data File, Score, Thesis, Archival Item, Video, Image, Map, Report, Coin, Databases, Audio, Book, Journal.

rladdusaw:
Resolves #3480

### DIFF
--- a/lib/bibdata_rs/src/marc/figgy.rs
+++ b/lib/bibdata_rs/src/marc/figgy.rs
@@ -1,16 +1,31 @@
-use crate::marc::control_field::control_number::ControlNumber;
+use crate::marc::{control_field::control_number::ControlNumber, utils::display_format::display_format};
 use figgy_marc::{FiggyMmsIdCache, ark_eq, iiif_manifest_url};
 use marctk::Record;
+use serde_json::{Value, json};
 use std::sync::LazyLock;
+use crate::marc::record_facet_mapping::formats;
 
 static FIGGY_MMS_REPORT_CACHE: LazyLock<FiggyMmsIdCache> =
     LazyLock::new(figgy_marc::redis_cache::read);
 
-pub fn figgy_1display(record: &Record) -> Option<String> {
+pub fn figgy_1display(record: &Record, cache: Option<&LazyLock<FiggyMmsIdCache>>) -> Option<String> {
+    let figgy_cache = cache.unwrap_or_else(|| &FIGGY_MMS_REPORT_CACHE);
+    
     match ControlNumber::from(record) {
-        ControlNumber::Alma(mms_id) => FIGGY_MMS_REPORT_CACHE
+        ControlNumber::Alma(mms_id) => figgy_cache
             .get(mms_id)
-            .and_then(|figgy_items| serde_json::to_string(figgy_items).ok()),
+            .map(|figgy_items| {
+                figgy_items
+                    .iter()
+                    .map(|figgy_item| {
+                        let mut item = figgy_item.as_object().unwrap().clone();
+                        let format = display_format(formats(&record));
+                        item.insert("display_format".to_owned(), json!(format));
+                        serde_json::to_value(item).unwrap()
+                    })
+                    .collect::<Vec<Value>>()
+            })
+            .and_then(|figgy_items| serde_json::to_string(&figgy_items).ok()),
         _ => None,
     }
 }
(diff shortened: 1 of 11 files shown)


lib/bibdata_rs/src/marc/figgy.rs
```diff
@@ -1,16 +1,31 @@
...
+    
     match ControlNumber::from(record) {
-        ControlNumber::Alma(mms_id) => FIGGY_MMS_REPORT_CACHE
+        ControlNumber::Alma(mms_id) => figgy_cache
             .get(mms_id)
-            .and_then(|figgy_items| serde_json::to_string(figgy_items).ok()),
+            .map(|figgy_items| {
+                figgy_items
+                    .iter()
+                    .map(|figgy_item| {
```
sandbergja, line 26: For better separation of concerns (i.e. this package is focused on getting data from Figgy, not on calculating formats), could we instead inject this behavior?

So the signature would be something like (very rough idea, I haven't tried to compile this and probably have a syntax error 😆):

```
pub fn figgy_1display<M>(
    record: &Record,
    cache: Option<&LazyLock<FiggyMmsIdCache>>,
    modify_items: M
) -> Option<String>
where M: Fn(Map<String, Value>) -> Map<String, Value>
```

And you could use it like:

```
figgy_1display(&record, None, |item| {
  item.insert("display_format".to_owned(), json!(format));
  item
})
```

lib/bibdata_rs/src/marc/ruby_bindings.rs
```diff
@@ -109,7 +112,7 @@ fn solr_fields(ruby: &Ruby, record: magnus::RObject) -> Result<RHash, magnus::Er
         .ok()
         .and_then(|date| date.maybe_to_string());
 
-    let hash = ruby.hash_new_capa(130);
+    let hash = ruby.hash_new();
```
sandbergja: Could we keep the capacity for the hash?

lib/bibdata_rs/src/marc/utils/display_format.rs
```diff
@@ -0,0 +1,21 @@
+// Choses a display format given a vector of formats an if it's electronic
```
sandbergja: Could we also describe how a "display format" is different from a format?  Alternative thought: maybe the `FormatFacet` enum could have an `abbreviation` method or something of that nature to help clarify how the two concepts are related?

sandbergja (commented):
Thanks, @rladdusaw ! 🐈‍⬛  I had two more architectural ideas, let me know what you think when you are back.

lib/bibdata_rs/src/marc/ruby_bindings.rs
rladdusaw: Yep, I was in a rush and forgot to change this back. It fails for me locally unless I change it for testing.

lib/bibdata_rs/src/marc/utils/display_format.rs
```diff
@@ -0,0 +1,25 @@
+use crate::solr::FormatFacet::*;
+use crate::solr::format_facet::Abbreviation;
+
+// Choses an abbreviated display format given a vector of formats
+// this is intended for displaying compact badges inline with search results
+pub fn display_format(formats: Vec<String>) -> String {
+    match formats {
+        _ if formats.contains(&"Microform".to_string()) => Microform.abbreviation(),
```
sandbergja, line 8: This seems really duplicative now.  Is there a way we could reduce the duplication?  Perhaps we could add these to the existing FromStr impelmentation?

lib/bibdata_rs/src/marc/figgy.rs
sandbergja: I guess my goal is that there should be no `display_format` implementation details in this package, for two reasons:

1. compilation time -- right now, this package can compile on its own, if we rely on `display_format` here, it needs to wait to see if there are any `bibdata_rs` changes it needs to compile first
2. simplicity -- it just seems simpler to not need to think about display formats when you are working on the redis figgy output, and likewise to not think about redis figgy output when you are working on display formats...

rladdusaw:
Closed in favor of #3495

PR #3489, opened 2026-08-25
Allow tests to run in devbox

sandbergja:
This involved some changes to tests to remove assumptions that they are being run in CI or lando.

All tests pass for me in `devbox` and `lando`

### DIFF
--- a/flakes/README.md
+++ b/flakes/README.md
@@ -1,0 +1,17 @@
+### Nix flakes
+
+This directory contains nix flakes that are used for a Bibdata environment.
+
+#### Developing flakes
+
+To test a flake to make sure it builds correctly:
+
+1. `cd` into its directory
+1. `nix build --extra-experimental-features flakes`
+
+To see a flake on disk:
+
+1. `ls /nix/store` and find the output that corresponds to your flake
+1. If you have been iterating on the flake a lot and have many different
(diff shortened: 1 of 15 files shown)


flakes/README.md
```diff
@@ -0,0 +1,17 @@
+### Nix flakes
```
sandbergja, line 1: Copied from Orangelight

spec/jobs/index/dump_file_job_spec.rb
```diff
@@ -6,7 +6,7 @@
 
   describe '#perform' do
     it 'raises an error when traject errors' do
-      expect { described_class.new.perform(dump_file_id, 'http://localhost:8983/solr/badcollection') }.to raise_error(RuntimeError)
+      expect { described_class.new.perform(dump_file_id, 'http://localhost:123456/solr/badcollection') }.to raise_error(RuntimeError)
```
sandbergja, line 9: This test was really only passing by chance -- and indeed it would fail if you happened to have solr running locally on port 8983.  `123456` is an invalid port number (they only go up to 65535), so we should consistently get the expected error, regardless of what is running on our test environments.

kevinreiss (approved):

christinach (approved):
Thanks @sandbergja 🍄

PR #3492, opened 2026-08-25
Remove old publication_place_hierarchical_facet field

### LINKED ISSUE #3126: Safely remove publication_place_hierarchical_facet

Remove publication_place_hierarchical_facet from the indexing process after migrating all uses of it to publication_place_hierarchical_pipe_facet.

### Level of urgency
- [ ] High
- [ ] Moderate
- [x] Low

- There is a [comment in traject saying that we should remove this](https://github.com/pulibrary/bibdata/blob/fb33b3be5ab9207522dd3895516bf07c522da138/marc_to_solr/lib/traject_config.rb#L954)
- It seems that there may be some uses of this field that have since snuck into orangelight
- Removing it would simplify and speed up the indexing process

### Acceptance criteria

 - [ ] All uses of publication_place_hierarchical_facet are removed or migrated to publication_place_hierarchical_pipe_facet
 - [ ] We remove publication_place_hierarchical_facet from the indexing process

sandbergja:
Closes #3126

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -813,14 +813,6 @@ to_field 'publication_place_hierarchical_pipe_facet', extract_marc('008[15-17]')
   accumulator.replace(places.compact.flatten)
 end
 
-# TODO: Remove in favor of publication_place_hierarchical_pipe_facet once reindex complete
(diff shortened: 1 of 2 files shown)


christinach (approved):
Thanks @sandbergja 🍄

PR #3493, opened 2026-08-26
Remove legacy lc_facet field in favor of lc_pipe_facet

sandbergja:
The new field was introduced in #2684 and is working well.  We can remove it as soon as https://github.com/pulibrary/orangelight/pull/6178 is merged (keeping this as a draft until then).


christinach (approved):
Thanks @sandbergja !

PR #3506, opened 2026-09-03
Index subject unstem terms into a single field

sandbergja:
Prior to this commit, we indexed subject unstem terms into a variety of different fields, for example `siku_subject_unstem_search`, `homoit_subject_unstem_search`, etc.  All of these fields were used in the same qf and pf, and were weighted the same.  Having these as separate fields meant that the queries that solr generated for keyword searches had many more clauses than they needed

Getting this into production will take 3 separate PRS:
1. This one
1. https://github.com/pulibrary/pul_solr/pull/577, which can be merged after this goes in and we reindex.
1. A PR to remove the existing siku_subject_unstem_search, homoit_subject_unstem_search, local_subject_display, and fast_subject_display from the indexing, which can be merged after the pul_solr pr goes in.

Helps with https://github.com/pulibrary/orangelight/issues/5691

### DIFF
--- a/marc_to_solr/lib/traject_config.rb
+++ b/marc_to_solr/lib/traject_config.rb
@@ -768,14 +768,25 @@ end
 
 rust_multi_value_field 'icpsr_subject_unstem_search'
 
-# Adds lc, siku, local, and homoit subject unstem_search fields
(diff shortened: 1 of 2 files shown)


rladdusaw (approved):
Thanks @sandbergja!

PR #3515, opened 2026-09-05
Stop indexing into specific unstem subject fields

sandbergja:
As of https://github.com/pulibrary/pul_solr/pull/577, we no longer use these fields, and instead use a single consolidated field to reduce the number of queries that solr generates.

A cleanup as part of #5691


kevinreiss (approved):
Thanks @sandbergja

PR #3519, opened 2026-09-08
Remove display_format from electronic_access_1display

sandbergja:
This is related to the bug described in #6111 and #6200


christinach (approved):
Thanks 🍄 🐬

PR #3523, opened 2026-09-09
Remove default redis crate features

sandbergja:
We do not use any of them (acl, geospatial, scripts, or streams).  This removes a few small dependencies.  If it saves any compile time at all, it's very slight, but can't hurt!


christinach (approved):
Thanks @sandbergja 🍄

PR #3524, opened 2026-09-09
Add a way to pass --timings to cargo

sandbergja:
This produces a nice HTML report that lists where cargo spent its time while compiling.


christinach (approved):
nice @sandbergja Thanks 🍄!

PR #3528, opened 2026-09-10
Refactor: extract functions for getting data from MARC 008 field

sandbergja:
Just DRYing up some code


lib/bibdata_rs/src/marc/utils/display_format.rs
```diff
@@ -29,7 +29,7 @@ pub fn display_format(formats: Vec<String>) -> String {
 #[cfg(test)]
 mod tests {
     use super::*;
-#[test]
+    #[test]
```
sandbergja, line 32: a `cargo fmt` fix

christinach (approved):
Thanks @sandbergja 🍄

PR #3529, opened 2026-09-10
Rewrite `mult_languages_iana_s` solr field in Rust

sandbergja:
Also do 2 small cleanups:
* Separate `is_valid_language_code` into a version that deals only with owned strings (for use in Ruby) and a version that can accept a string slice (for use in Rust, avoiding the allocation needed to create a new String).
* Simplify calling the `slice_from_008` method: it calls `.into()` so that the caller does not have to.


christinach (approved):
Thanks @sandbergja !!

PR #3530, opened 2026-09-11
Reduce allocations in solr mappings

sandbergja:
Output from `cargo bench --bench dataspace_bench`

```
Benchmarking solr_document_from_dataspace_document/from: Collecting 100 samples in estimated 5.0027 s (3.2M iteratiosolr_document_from_dataspace_document/from
                        time:   [1.5535 µs 1.5556 µs 1.5581 µs]
                        change: [−11.386% −11.021% −10.664%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Also, fix an issue with the dataspace benchmark. It previously spent most of its time reading the fixture file rather than in the code we actually wanted to measure


lib/bibdata_rs/benches/dataspace_bench.rs
```diff
@@ -22,19 +22,18 @@ fn title_normalize_benchmark(c: &mut Criterion) {
 
 fn dataspace_to_solr_benchmark(c: &mut Criterion) {
     let mut group = c.benchmark_group("solr_document_from_dataspace_document");
+    let fixture = File::open("../../spec/fixtures/files/theses/dsp01b2773v788.json").unwrap();
```
sandbergja, line 25: Previously, we included reading the fixture file in our measurements, even though this benchmark is meant to test converting Dataspace data into Solr data.  Reading the file took much more time than converting the data, so the benchmark was not very meaningful.

lib/bibdata_rs/src/solr/builder.rs
```diff
@@ -3,37 +3,37 @@
 use super::{AccessFacet, AuthorRoles, ElectronicAccess, FormatFacet, LibraryFacet, SolrDocument};
 
 #[derive(Debug, Default)]
-pub struct SolrDocumentBuilder {
-    author_s: Option<Vec<String>>,
-    author_sort_key: Option<String>,
+pub struct SolrDocumentBuilder<'a> {
+    author_s: Option<&'a [&'a str]>,
```
sandbergja, line 7: Almost all of the changes are to this struct -- rather than using owned types like `String` and `Vec` (which we often need to `.clone()`, we use borrowed types for several of the fields.

lib/bibdata_rs/src/solr/builder.rs
```diff
@@ -318,25 +312,32 @@ impl SolrDocumentBuilder {
...
     }
     pub fn build(&self) -> SolrDocument {
         SolrDocument {
             access_facet: self.access_facet,
-            advanced_location_s: self.advanced_location_s.clone(),
+            advanced_location_s: self.advanced_location_s.map(|locations| {
+                locations
+                    .iter()
+                    .map(|location| location.to_string())
+                    .collect()
```
sandbergja, line 326: We still end up allocating here (with `to_string()` and `collect()`), but it is later in the process, so we are able to avoid some `clone()`s

lib/bibdata_rs/src/theses/dataspace/document/normalize.rs
```diff
@@ -71,7 +66,7 @@ impl DataspaceDocument {
         )
     }
 
-    pub fn authorized_ceritificates(&self) -> Option<Vec<String>> {
```
sandbergja, line 74: fix typo in method name

lib/bibdata_rs/src/theses/department.rs
```diff
@@ -1,150 +1,121 @@
 // This module is responsible for mapping department names in Dataspace to the Library of Congress authorized names
 
-pub fn map_department(original: &str) -> Option<String> {
+pub fn map_department(original: &str) -> Option<&str> {
     match original {
         "African American Studies" => {
-            Some("Princeton University. Department of African American Studies".to_owned())
```
sandbergja, line 6: The diff for these files looks unnecessarily confusing (due to `cargo fmt` changes I think).  The whole difference is removing `to_owned()` from each entry.

christinach (approved):
Thanks @sandbergja 🍄 !

PR #3531, opened 2026-09-11
Pin json rubocop rails

christinach:
pin json to 2.21
v3.0 causes errors
there is a fix that was backported to `8-1-stable`

bundle update rubocop-rails
rubocop fixes


sandbergja (approved):
Thanks, @christinach ! 🐬

PR #3532, opened 2026-09-11
Store Metadatum contents as a CoW rather than an owned string

sandbergja:
Previously, the `Metadatum` struct stored an owned string inside.  This PR changes it to a `CoW`

This does not directly cause any change in benchmark performance, but it opens up possibilities for us to avoid more allocations in future work.

Also, there is one change in behavior.  When a value is missing, we now represent it as an empty vec rather than a vec that contains an empty string.

### DIFF
--- a/lib/bibdata_rs/src/theses/dataspace/document.rs
+++ b/lib/bibdata_rs/src/theses/dataspace/document.rs
@@ -1,5 +1,6 @@
 // This module is responsible for representing Dataspace Document's metadata
 
+use std::borrow::Cow;
 use std::collections::HashMap;
 
 use serde::de::Deserializer;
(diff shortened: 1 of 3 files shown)


christinach (approved):
@sandbergja 🍄 question but not a blocker. It is much better to have `[]` rather than `[""]` . Will this still bring the field in the solr document?

sandbergja:
> @sandbergja 🍄 question but not a blocker. It is much better to have `[]` rather than `[""]` . Will this still bring the field in the solr document?

Good question!  We will still post the field to solr as an empty JSON array, but solr appears to ignore empty arrays that we send it.  So ultimately, the document in solr will not have the field.

FWIW, as far as I can tell, none of these fields are missing from the production data in the new https://theses-dissertations.princeton.edu.

PR #3544, opened 2026-09-14
Don't index into title_vern_sort field

sandbergja:
Prior to this commit, we indexed a field called `title_vern_sort`, which we never used directly.  The field was copied into `title_245_la` with a solr `copyField` (and we do use `title_245_la`).

Rather than indexing into both `title_vern_sort` and `title_245_la` (and needing to do analysis for both fields when we index), we can index directly into `title_245_la` and skip the unused field.


christinach (approved):
Thanks @sandbergja !

PR #3545, opened 2026-09-14
Commit the redis config file

sandbergja:
Without it, we get the error:

```
Fatal error, can't open config file '[...]bibdata/devbox.d/redis/redis.conf': No such file or directory
```


christinach (approved):
Thanks @sandbergja ! 🍄 
This resolves the issue I'm seeing in my local with redis.

PR #3547, opened 2026-09-14
electronic_access_1display index the catalog production url

christinach:
related to [orangelight#6225](https://github.com/pulibrary/orangelight/issues/6225)

this commit hardcodes the production environment so that we make the Digital content to point to the production viewer. It was hardcoded to point to the staging one. 

Next steps are to update bibdata to use an env ( CATALOG_URL) for the catalog url.

I've indexed ephemera on catalog-staging and confiremd that the Digital content is now hardcoded to look [catalog production](https://catalog-staging.princeton.edu/catalog/759b025c-e3c8-4afd-a268-96456fdf968a).


kevinreiss (approved):
Thanks @christinach

